PR #109, opened 2014-08-29
[WIP] Set default viewport size to layout content

redapple:
Experimenting with http://qt-project.org/doc/qt-4.8/qwebpage.html#preferredContentsSize-prop

Also using a wider default viewport (1366x768): this can be reverted of course


kmike:
Closing it in favor of https://github.com/scrapinghub/splash/pull/159.

@immerrr: what do you think about a larger default window size, like @redapple suggested?

immerrr:
@kmike, it makes sense to make default viewport wider and go for some standard 16:9 resolution. 1366x768 is standard for low end laptop screens, so webmasters have take it into account when planning their adaptive layout, let's use it.

Issue #121, opened 2014-10-22
rasterizing of very large pages


kmike:
Issue described here: https://code.google.com/p/phantomjs/issues/detail?id=54 is valid for Splash. This fails: http://127.0.0.1:8050/render.png?url=http://www.w3.org/TR/2010/REC-xhtml-basic-20101123/&viewport=full&wait=0.5

Workaround: https://github.com/ariya/phantomjs/blob/eddb0db1d253fd0c546060a4555554c8ee08c13c/src/webpage.cpp#L1027

PR #131, opened 2014-12-10
[WIP] Web view geometry inheritance revisited.

sardok:
I tried to bring the notion of different device screen option which allows you to render the page as though the webpage is visited from a particular screen sized device.

The motivation:
I've come across with a website which dynamically changes its content and hides some of information if the device is smaller than a desktop. QWebPage inherits screen geometry information from QWebView (which does the same from QWidget most probably). So far, the geometry run at default value (640x480 if i'm not wrong). This PR, make QWebPage to set geometry immediate after it is created, i didnt test the effect changing the geometry at runtime and what happens to QWebPage.

This PR brings a simple unit test which expects a screen width to be a default value. I added custom yield_fixtures, having thinking to run TestServer() with different settings object which should be carried down to the WebpageRender. Now, there is no way to write test which uses different settings/configuration unless defaults.py is physically changed on disk.

So, do you think, is there a need for small refactoring about about running server at different configuration or is it better to look for an update which allows you to change the geometry (device screen size) at runtime, based on a url parameter for example?

### DIFF
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -14,6 +14,7 @@ from PyQt4.QtNetwork import QNetworkRequest
 from twisted.internet import defer
 from twisted.python import log
 from splash import defaults
+from splash.utils import parse_viewport
 from splash.qtutils import qurl2ascii, OPERATION_QT_CONSTANTS, qt2py, WrappedSignal
 from splash.har.utils import without_private
 
@@ -67,6 +68,8 @@ class BrowserTab(object):
         self._setup_webpage_events()
 
         self.web_view = QWebView()
+        geometry = render_options.get_geometry()
+        self.web_view.setGeometry(0, 0, *parse_viewport(geometry))
         self.web_view.setPage(self.web_page)
         self.web_view.setAttribute(Qt.WA_DeleteOnClose, True)
 
--- a/splash/render_options.py
+++ b/splash/render_options.py
@@ -170,6 +170,9 @@ class RenderOptions(object):
         except ValueError:
             raise BadOption("Invalid viewport format: %s" % viewport)
 
+    def get_geometry(self):
+        return self.get("geometry", self.get_viewport())
+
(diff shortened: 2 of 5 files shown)


immerrr:
Now that we've got #121 fixed, we're proceeding with this one.

The idea is as follows:
- replace geometry -> window_size for the purpose of this issue: technically,
  geometry also includes window position and we are not interested in it anywhere

- make window size the primary size: changing window size should change
  viewport size, but leave a way to override viewport size for backward
  compatibility and PNG rendering (viewport=full)

- default/fallback sizes should be provided for window, not viewport

- `/render.xxx` endpoints should be backward compatible, except for the default
  window size which is expected to fix things, not break things

- add `window_size=` parameter to `/render.xxx` query parameters. If specified alone it should
  modify window & viewport size simultaneously, but if `viewport=` is also used it must override viewport size

- provide `set_window_size` function to Lua runtime

- ensure that `set_viewport('full')` raises an error in Lua where the user can
  choose report or handle the error, but provide a silent fallback for
  "non-interactive" scenarios

@kmike, have I forgotten anything? :)

@sardok, are you OK with me taking it from here or do you intend to push this
yourself?

kmike:
@immerrr there are other possible changes in this direction (e.g. allowing clip rects), but what you've outlined looks like a good start.

> default/fallback sizes should be provided for window, not viewport

Do we need a "fallback" size, what is it?

> provide set_window_size function to Lua runtime

It'd great to check if a webpage adapts to a new size immediately or not. If not, we should decide whether splash:set_window_size should be async or should we document that webpages could need some time to adapt to the new size.

sardok:
@immerrr feel free to proceed.

kmike:
I'm closing this PR in favor of https://github.com/scrapinghub/splash/pull/159.

@sardok thanks for working on it, it was very helpful to discuss this feature with you and check the real code!

PR #147, opened 2015-01-06
Fix PNG rendering of web pages when viewport is huge

### LINKED ISSUE #121: rasterizing of very large pages

immerrr:
This should close #121.

### DIFF
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -1,20 +1,22 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+from cStringIO import StringIO
 import os
 import base64
 import copy
 import pprint
 import weakref
 import functools
+from PIL import Image
 from PyQt4.QtWebKit import QWebPage, QWebSettings, QWebView
-from PyQt4.QtCore import (Qt, QUrl, QBuffer, QSize, QTimer, QObject,
-                          pyqtSlot)
-from PyQt4.QtGui import QPainter, QImage, QMouseEvent, QKeyEvent
+from PyQt4.QtCore import Qt, QUrl, QSize, QTimer, QObject, pyqtSlot
+from PyQt4.QtGui import QMouseEvent, QKeyEvent
 from PyQt4.QtNetwork import QNetworkRequest
 from twisted.internet import defer
 from twisted.python import log
 from splash import defaults
-from splash.qtutils import qurl2ascii, OPERATION_QT_CONSTANTS, qt2py, WrappedSignal
+from splash.qtutils import (qurl2ascii, OPERATION_QT_CONSTANTS, qt2py,
+                            WrappedSignal, render_qwebpage)
 from splash.har.qt import cookies2har
 from splash.har.utils import without_private
 
@@ -489,20 +491,25 @@ class BrowserTab(QObject):
     def png(self, width=None, height=None, b64=False):
         """ Return screenshot in PNG format """
         self.logger.log("getting PNG", min_level=2)
-
-        image = QImage(self.web_page.viewportSize(), QImage.Format_ARGB32)
-        painter = QPainter(image)
-        self.web_page.mainFrame().render(painter)
-        painter.end()
+        image = render_qwebpage(self.web_page, self.logger)
         self.store_har_timing("_onScreenshotPrepared")
 
         if width:
-            image = image.scaledToWidth(width, Qt.SmoothTransformation)
+            assert width > 0
+            old_width, old_height = image.size
+            if old_width > 0:
+                new_height = int(old_height * width / float(old_width))
+                image = image.resize((width, new_height),
+                                     resample=Image.BILINEAR)
+            else:
+                image = image.crop((0, 0, width, old_height))
+        else:
+            width = image.size[0]
         if height:
-            image = image.copy(0, 0, width, height)
-        b = QBuffer()
+            image = image.crop((0, 0, width, height))
+        b = StringIO()
         image.save(b, "png")
-        result = bytes(b.data())
+        result = bytes(b.getvalue())
         if b64:
             result = base64.b64encode(result)
         self.store_har_timing("_onPngRendered")
--- a/splash/defaults.py
+++ b/splash/defaults.py
@@ -18,6 +18,18 @@ MAX_HEIGTH = 1080
 
 AUTOLOAD_IMAGES = 1
 
+# There's a bug in Qt that manifests itself when width or height of rendering
+# surface (aka the png image) is more than 32768.  Usually, this is solved by
+# rendering the image in tiled manner and obviously, TILE_MAXSIZE must not
+# exceed that value.
+#
+# Other than that, the setting is a tradeoff between performance and memory
+# usage, because QImage that acts as a rendering surface is quite a resource
+# hog.  So, if you increase tile size you may end up using a lot more memory,
+# but there is less image pasting and the rendering is faster.  As of now, 2048
+# size is chosen to fit commonly used 1080p resolution in one tile.
+TILE_MAXSIZE = 2048
+
 # defaults for render.json endpoint
 DO_HTML = 0
 DO_IFRAMES = 0
--- a/splash/qtutils.py
+++ b/splash/qtutils.py
@@ -6,16 +6,20 @@ import sys
 import time
 import itertools
 import functools
+import struct
+import array
 
 from twisted.python import log
-from PyQt4.QtGui import QApplication
+from PIL import Image
+from PyQt4.QtGui import QApplication, QImage, QPainter
 from PyQt4.QtCore import (
     QAbstractEventDispatcher, QVariant, QString, QObject,
-    QDateTime, QRegExp,
+    QDateTime, QRegExp, QSize, QRect
 )
 from PyQt4.QtCore import QUrl
 from PyQt4.QtNetwork import QNetworkAccessManager, QNetworkReply
 
+from splash import defaults
 from splash.utils import truncated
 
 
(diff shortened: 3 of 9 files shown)


kmike:
It doesn't fix an example from #121 for me - returned image is still invalid. Does it work for you?

Also, for me the new test passes even without introduced BrowserTab.png changes.

I'm using OS X 10.10.1,  Qt 4.8.6, PyQt 4.11.1, WebKit 534.34, sip 4.16.3, Twisted 14.0.2, Lua 5.2

kmike:
.. but it could - http://127.0.0.1:8050/render.png?url=http://www.w3.org/TR/2010/REC-xhtml-basic-20101123/&viewport=full&wait=0.5 uses 4.7GB peak on my local machine.

splash/browser_tab.py
```diff
@@ -731,3 +730,64 @@ def log(self, message, min_level=None):
...
+                       min_level=2)
+            web_page.mainFrame().render(tile_painter)
+            pil_tile_image = qimage_to_pil_image(tile_image)
+
+            logger.log("Pasting rendered tile to coords: %s" %
+                       ((left, top),),
+                       min_level=2)
+            pil_image.paste(pil_tile_image, (left, top))
+    tile_painter.end()
+    return pil_image_to_qimage(pil_image)
```
kmike: Why do we need to create QImage from PIL image? `pil_image_to_qimage` takes extra 6 seconds for w3.org example, and PIL can resize images and convert them to PNG just fine (I even expect it to be faster).

kmike: .. and it seems it is QImage creation which uses 4GB of memory; building of PIL image tops at 360MB if a smaller tile size is used (1024px). For 512px tiles top memory usage is 280MB. I suggest to use smaller tile size like this (or maybe larger, to fit common screen widths) because it seems Qt is a memory hog here - with 16384 tile it uses more than 2GB when rendering. With 512px tiles rendering is about 2x slower than with 16K tiles; we need to find a balance.

splash/browser_tab.py
```diff
@@ -731,3 +729,95 @@ def log(self, message, min_level=None):
...
+
+def _qimage_to_pil_image(qimage):
+    """Convert QImage (in ARGB32 format) to PIL.Image (in RGBA mode)."""
+    # In our case QImage uses 0xAARRGGBB format stored in host endian order,
+    # we must convert it to [0xRR, 0xGG, 0xBB, 0xAA] sequences used by pillow.
+    buf = qimage.bits().asstring(qimage.numBytes())
+    if not _is_little_endian:
+        # If host is big-endian, convert to little-endian first.
+        count = len(buf) // 4
+        buf = struct.pack('<%di' % count, *struct.unpack('=%di' % count, buf))
```
immerrr: It is documented as follows [here](http://qt-project.org/doc/qt-4.8/qimage.html#image-formats):

*... the QRgb typedef is equivalent to an unsigned int containing an ARGB quadruplet on the format 0xAARRGGBB.*
*32-bit images have no color table; instead, each pixel contains an QRgb value. There are three different types of 32-bit images storing RGB (i.e. 0xffRRGGBB), ARGB and premultiplied ARGB values respectively....*

Yes, we can and it is indeed faster. We'll lose exotic middle-endian support, but it should be ok. Changed.

splash/browser_tab.py
```diff
@@ -731,3 +729,95 @@ def log(self, message, min_level=None):
...
+                       ((left, top),),
+                       min_level=2)
+            pil_image.paste(pil_tile_image, (left, top))
+    tile_painter.end()
+    return _pil_image_to_qimage(pil_image)
+
+
+# XXX: brain-dead simple endianness detection, is there a better way?
+_is_little_endian = \
+    struct.pack('<i', 0x11223344) != struct.pack('=i', 0x11223344)
```
immerrr: Done.

splash/browser_tab.py
immerrr: The only reason is making the function a drop-in replacement. Sure, we can ditch QImage completely.

kmike: Tile size looks like an important number to tweak - could you please move it to defaults.py?
I also wonder if square tiles of fixed size are always the best. For example, if we have a page which is 1200px wide and use 1000px tiles then we'll render 2 tiles per row: one 1000px wide and another 200px wide, both 1000px high. Won't it be better to use a single 1200x(1000/1.2)px tile? It has the same area (and thus uses the same amount of memory), but we'll need less such tiles, so rendering could be faster (which I'm unsure of though, it is better to check that).

splash/browser_tab.py
```diff
@@ -731,3 +730,64 @@ def log(self, message, min_level=None):
...
+    tile_painter = QPainter(tile_image)
+    viewport = QRect(0, 0, size.width(), size.height())
+    tile_painter.setViewport(viewport)
+
+    # The other bug manifests itself when you do painter.drawImage trying to
+    # concatenate tiles onto a single image and once you reach 32'768 along
+    # either dimension all of a sudden drawImage simply stops drawing anything.
+    # The simplest workaround that comes to mind is to use pillow for pasting
+    # images.
+    pil_image = Image.new(mode='RGBA',
```
kmike: A minor style tweak :) If you create a "size" variable and then use it in `Image.new(.., size=size)` then it'll be the same amount of code lines, but less indentation.

immerrr: There already *is* a `size` variable, but it's QSize and is not iterable to be converted to tuple inline. Do you insist on making another size variable? :)

If not, I'd leave it as is for now but look out for such situations in future.

kmike: it was only a suggestion, don't worry :)

splash/qtutils.py
```diff
@@ -217,3 +221,125 @@ def connect(self, func, **kwargs):
...
+    if size.width() < tile_maxsize and size.height() < tile_maxsize:
+        image = QImage(size, QImage.Format_ARGB32)
+        painter = QPainter(image)
+        web_page.mainFrame().render(painter)
+        # It is important to end painter explicitly in python code, because
+        # Python finalizer invocation order, unlike C++ destructors, is not
+        # deterministic and there is a possibility of image's finalizer running
+        # before painter's which may break tests and kill your cat.
+        painter.end()
+        return qimage_to_pil_image(image)
```
kmike: Could you please check if Splash becomes slower these basic use cases or not?

splash/qtutils.py
```diff
@@ -217,3 +221,125 @@ def connect(self, func, **kwargs):
...
+        # If host is big-endian, convert endianness first.
+        buf = swap_byte_order_i32(buf)
+    # QImage's 0xARGB in little-endian becomes [0xB, 0xG, 0xR, 0xA] for pillow,
+    # hence the 'BGRA' decoder argument.
+    return Image.frombytes(
+        'RGBA', (qimage.size().width(), qimage.size().height()), buf,
+        'raw', 'BGRA')
+
+
+def pil_image_to_qimage(pil_image):
```
kmike: This function is unused; even though it may be useful in some cases, I think it is better to delete it from Splash.

splash/qtutils.py
```diff
@@ -217,3 +221,125 @@ def connect(self, func, **kwargs):
...
+_is_little_endian = struct.pack('<i', 0x1122) == struct.pack('=i', 0x1122)
+
+
+def qimage_to_pil_image(qimage):
+    """Convert QImage (in ARGB32 format) to PIL.Image (in RGBA mode)."""
+    # In our case QImage uses 0xAARRGGBB format stored in host endian order,
+    # we must convert it to [0xRR, 0xGG, 0xBB, 0xAA] sequences used by pillow.
+    buf = qimage.bits().asstring(qimage.numBytes())
+    if not _is_little_endian:
+        # If host is big-endian, convert endianness first.
```
kmike: a nitpick: this comment is unneeded, variable and function names are already good

splash/qtutils.py
```diff
@@ -217,3 +221,125 @@ def connect(self, func, **kwargs):
...
+    """
+    size = web_page.viewportSize()
+
+    # A bug is triggered when width or height is >= 32768 and tile maxsize must
+    # not exceed that value.
+    #
+    # Other than that, the setting is a tradeoff between performance and memory
+    # usage, because QImage is quite a resource hog.  So, if you increase tile
+    # size you may end up using a lot more memory, but there is less image
+    # pasting and the rendering is faster.
```
kmike: This comment could be more visible if put in defaults.py.

splash/qtutils.py
kmike: If Splash becomes slower we need to keep the old fast-path.

splash/defaults.py
```diff
@@ -18,6 +18,8 @@
 
 AUTOLOAD_IMAGES = 1
 
+TILE_MAXSIZE = 2048
```
kmike: :+1: looks like a good default value

splash/qtutils.py
immerrr: done

splash/qtutils.py
immerrr: done

splash/qtutils.py
immerrr: done

kmike:
:cake:

PR #154, opened 2015-01-12
(WIP) Optimize png rendering

immerrr:
This is a work-in-progress PR to avert performance impact of the latest tiled rendering patch.

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -934,8 +934,9 @@ Return performance-related statistics.
 
 **Signature:** ``splash:get_perf_stats()``
 
-This function returns a table that can be useful for performance analysis.  As
-of now, it provides:
+**Returns:** a table that can be useful for performance analysis.
+
+As of now, this table contains:
 
 * ``walltime`` - (float) number of seconds since epoch, analog of ``os.clock``
 * ``cputime`` - (float) number of cpu seconds consumed by splash process
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -925,6 +925,23 @@ Example:
 
 Named arguments are not supported for this function.
 
+.. _splash-get-perf-stats:
+
+splash:get_perf_stats
+---------------------
+
+Return performance-related statistics.
+
+**Signature:** ``splash:get_perf_stats()``
+
+This function returns a table that can be useful for performance analysis.  As
+of now, it provides:
+
+* ``walltime`` - (float) number of seconds since epoch, analog of ``os.clock``
+* ``cputime`` - (float) number of cpu seconds consumed by splash process
+* ``maxrss`` - (int) high water mark number of bytes of RAM consumed by splash
+  process
+
 .. _splash-args:
 
 splash.args
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -498,7 +498,7 @@ class BrowserTab(QObject):
         self.store_har_timing("_onScreenshotPrepared")
 
         b = StringIO()
-        image.save(b, "png")
+        image.save(b, "png", compress_level=defaults.PNG_COMPRESSION_LEVEL)
         result = bytes(b.getvalue())
         if b64:
             result = base64.b64encode(result)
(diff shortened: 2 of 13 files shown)


immerrr:
I've implemented resizing the images during painting, here are the visual differences: http://imgur.com/a/0qTp1. In short, letters are crispier, but "jumpier".

As for performance, I've made up a simple benchmark script that employs `benchmark_png` function implemented in https://github.com/immerrr/splash/commit/7cd72feb802b1024fbcbe34dfd50d7a13caed2ed:
```python
from splash.tests.stress import benchmark_png

for url in ["http://wikipedia.org",
            "http://google.com",
            "http://www.w3.org/TR/2010/REC-xhtml-basic-20101123/"]:
    for width in [None, 500, 2000]:
        try:
            print "url: ", url, "width: ", width, "result: ", benchmark_png(
                url=url, width=width)
        except Exception:
            import traceback
            print traceback.format_exc()
```

The results are as follows:

this-pr: ec149a5
after-fix: 7cd72fe
before-fix: 9adf4e4

```
| url                                             | width | this-pr | after-fix | before-fix |
|-------------------------------------------------+-------+---------+-----------+------------|
| http://wikipedia.org                            |  None |   0.128 |     0.127 |      0.113 |
| http://wikipedia.org                            |   500 |   0.090 |     0.163 |      0.127 |
| http://wikipedia.org                            |  2000 |   0.963 |     1.617 |      1.450 |
| http://google.com                               |  None |   0.030 |     0.030 |      0.030 |
| http://google.com                               |   500 |   0.020 |     0.033 |      0.030 |
| http://google.com                               |  2000 |   0.190 |     0.273 |      0.243 |
| http://w3.org                                   |  None |   0.213 |     0.213 |      0.193 |
| http://w3.org                                   |   500 |   0.163 |     0.240 |      0.193 |
| http://w3.org                                   |  2000 |   1.840 |     2.380 |      2.023 |
| http://w3.org/TR/2010/REC-xhtml-basic-20101123/ |  None |   3.490 |     3.477 |         xx |
| http://w3.org/TR/2010/REC-xhtml-basic-20101123/ |   500 |   0.763 |     2.340 |         xx |
| http://w3.org/TR/2010/REC-xhtml-basic-20101123/ |  2000 |   9.943 |    16.213 |         xx |

```

kmike:
Looks good! New resizing provides a better image quality IMHO, and the speed boost looks great.

A few questions:

1. Is splash/benchmark.py still needed?
2. Could you please check the speed with width > TILE_MAXSIZE, e.g 2050px?
3. Could you please also check it on an images-heavy website like http://blog.pinterest.com/?
4. How did the memory consumption change? 
5. Tests are failing - why? They shold be fixed before merging.

immerrr:
re:splash/benchmark.py
That file is useful to run detailed benchmarks with proper utilities.  I agree it doesn't make much sense to update it with every change to rendering workflow, but it can be useful for developers to do some hard-figure benchmarks. Can we have it stashed somewhere with a big disclaimer saying that it may get out of date?

The reason why I'm asking is because one of such benchmarks has shown that:
- most of rendering time is spent serializing the image in PNG format, 70 - 90%, on all of the sites I've tested.
- `QImage->QBuffer` serialization takes same time as `QImage->PIL.Image->cStringIO.StringIO`, and it is no more than 5% of total render time, decreasing when increasing image size
- rendering starts to take same amount of time as png serialization at tile size of 128x128

On one hand, this means that we cannot win much cpu time fine-tuning the rendering algo, e.g. increasing tile size. On the other we can easily trade 15-50% of cpu time for 10-15% of png size by using minimum compression_level. I'd say we do that and probably add `compression_level` argument to query params for users to play with.

re: maxrss, re:images-heavy, re: width>2048
Here's the updated benchmark result table: https://gist.github.com/immerrr/bd0a474d538e46b24415. Note the "4.complevel" entry that denotes saving png with undocumented `compression_level=1` parameter.

re: failing tests
looks like binary mismatch of resulting pngs, will look closer tomorrow

kmike:
Looks great, thanks!

PR #155, opened 2015-01-15
max-timeout command line option.

sibiryakov:


splash/render_options.py
```diff
@@ -20,14 +20,15 @@ def __init__(self, data):
         self.data = data
 
     @classmethod
-    def fromrequest(cls, request):
+    def fromrequest(cls, request, max_timeout=defaults.MAX_TIMEOUT):
```
kmike: I think it is cleaner to provide a `get_max_timeout` method - other options are not passed to the fromrequest method.

splash/render_options.py
```diff
@@ -20,14 +20,15 @@ def __init__(self, data):
...
         """
         Initialize options from a Twisted Request.
         """
 
-        # 1. GET / POST data
         data = {key: values[0] for key, values in request.args.items()}
+        data.setdefault('max-timeout', max_timeout)
```
kmike: this breaks tests, see https://travis-ci.org/scrapinghub/splash/jobs/47154223

splash/render_options.py
```diff
@@ -20,14 +20,15 @@ def __init__(self, data):
...
 
+        # 1. GET / POST data
```
kmike: "GET/POST data" comment is for `request.args.items`, so it makes more sense to put it above `data = {key: values[0] for key, values in request.args.items()}`

splash/render_options.py
```diff
@@ -97,7 +98,7 @@ def get_wait(self):
         return self.get("wait", defaults.WAIT_TIME, type=float, range=(0, defaults.MAX_WAIT_TIME))
 
     def get_timeout(self):
-        return self.get("timeout", defaults.TIMEOUT, type=float, range=(0, defaults.MAX_TIMEOUT))
+        return self.get("timeout", defaults.TIMEOUT, type=float, range=(0, self.get("max-timeout", defaults.MAX_TIMEOUT, type=float)))
```
kmike: I think it shouldn't be possible to override max_timeout via GET or POST request. It should be an option set via command-line. It is there to limit resource usage, and allowing to set it using GET/POST doesn't make sense.

splash/render_options.py
kmike: ah, disregard this comment

splash/server.py
```diff
@@ -29,6 +29,8 @@ def parse_opts():
         help="port to listen to (default: %default)")
     op.add_option("-s", "--slots", type="int", default=defaults.SLOTS,
         help="number of render slots (default: %default)")
+    op.add_option("--max-timeout", type="float", default=defaults.MAX_TIMEOUT,
+        help="maximum request processing time (default: %default)")
```
kmike: It is not maximum request processing time - maximum request processing time is set in [timeout](http://splash.readthedocs.org/en/latest/api.html#arg-timeout) GET or POST argument. `max-timeout` is a maximum allowed value for timeout.

splash/render_options.py
kmike: It seems that setting this option to default.MAX_TIMEOUT here doesn't make sense, because the default value is already set in server.py. The advantage of not setting it here is that if you forget to pass this variable you'll get an exception instead of a wrong value.

sibiryakov: I think it would be better to implement max-timeout as a usual member of RenderOptions. What do you think?

sibiryakov:
May be it makes more sense to generalize the way non-request options are passed to the RenderOptions? I.e. to create a special GlobalOptions object?

kmike:
The code looks good now :+1: 

Could you please add a test for --max-timeout option and rebase the code?

>  I.e. to create a special GlobalOptions object?

Yes, it makes sense. It is not necessary to get this PR merged, but GlobalOptions is a welcome change.

sibiryakov:
Just for myself: fix the issue with default timeout.

splash/tests/test_render.py
```diff
@@ -627,6 +626,38 @@ def _iframes_request(self, params):
         return self.request(query).json()
 
 
+class CommandLineOptionsTest(BaseRenderTest):
+
+    def test_max_timeout(self):
+        with SplashServer(extra_args=['--max-timeout=5.0']) as splash:
```
kmike: what about using a smaller max-timeout, to make tests run faster?

splash/tests/test_render.py
```diff
@@ -627,6 +626,38 @@ def _iframes_request(self, params):
...
+                    'timeout': '30.0',
+                },
+            )
+            self.assertStatusCode(r1, 400)
+
+            r2 = requests.get(
+                url=splash.url("render.html"),
+                params={
+                    'url': self.mockurl("delay?n=10"),
+                    'timeout': '5.0',
```
kmike: we should also check a case when 'timeout' argument is not passed: loading should fail if it takes longer than --max-timeout

splash/tests/test_render.py
```diff
@@ -627,6 +626,45 @@ def _iframes_request(self, params):
         return self.request(query).json()
 
 
+class CommandLineOptionsTest(BaseRenderTest):
+
+    def test_max_timeout(self):
+        with SplashServer(extra_args=['--max-timeout=0.1']) as splash:
```
kmike: It seems r4 test can become unstable with this timeout - what about 0.5 or so?

kmike: let's see how it goes with 0.1 first.

kmike:
:cake:

PR #157, opened 2015-01-16
lua_runonce: add kwargs to be passed to Lua script

immerrr:
This PR also addresses several points about `benchmark_png`.

* benchmark_png: add timeout kwarg

* benchmark_png: fix Lua script to use get_perf_stats

* benchmark_png: use splash.args to pass in arguments


kmike:
thanks!

Issue #150, opened 2015-01-08
Rendering errors on long pages


mehaase:
I have found that very long pages have some PNG rendering bugs. For example, [here is a forum thread with infinite scroll](http://what.thedailywtf.com/t/papers-please/6851). I have written a Lua directive that scrolls the page to the bottom several times (to trigger infinite scroll) and then scrolls back to the top and takes a snapshot.

Here's what the QWebView looks like in a real window:

![discourse1](https://cloud.githubusercontent.com/assets/320904/5670674/02176e50-9750-11e4-8c21-0e193ca82d5f.png)

And here's what it looks like when rendered as a PNG (warning, this is _very_ tall):

![discourse2](https://cloud.githubusercontent.com/assets/320904/5670684/11a4b224-9750-11e4-8849-654b09a84ea9.png)

Notice that there's a huge white gap in the PNG that does not exist in the QWebView.

Here's an excerpt from my Lua script:

```lua
  function main(splash)
    local url = splash.args.url
    local page_count = splash.args.page_count or 3

    splash:autoload(splash.args.js_source)
    assert(splash:go(url))

    for i=1,page_count do
      scroll_to_bottom_and_wait(splash)
    end

    splash:runjs([[
      window.__scrollTop();
    ]])
    splash:stop()
    splash:set_viewport("full")

    return {
      html = splash:html(),
      png = splash:png{width=640},
      har = splash:har(),
    }
  end
```

Obviously I'm omitting quite a bit but I think this shows the core of the problem. This may not be a bug in splash -- I'm not sure. With even larger/long PNGs, the rendering is even further degraded, until it looks almost nothing like the original web page at all.

kmike:
There are two known issues with these large pages.

The first issue is a QT bug which prevents a page from being rendered if one of the dimensions is greater than 32K. It is worked on here: https://github.com/scrapinghub/splash/pull/147.

The second issue is that viewport doesn't match window geometry, and it may confuse a webpage. It is worked on here: https://github.com/scrapinghub/splash/pull/131.

I'm not sure if these PRs will fix the problem described in this ticket or not, but they could. Similar issues:  https://github.com/scrapinghub/splash/issues/145, https://github.com/scrapinghub/splash/issues/80.

kmike:
[...]

Issue #145, opened 2015-01-01
viewport=full does not resize to correct width


bettertest-org:
Png is too narrow when setting viewport to full.

Example pages to grab:
http://allegro.pl/ShowItem2.php?item=4892071886
http://allegro.pl/ShowItem2.php?item=4879256257 

viewport cannot be set to fixed parameters because height is different

kmike:
Hi,

This PR can be related: https://github.com/scrapinghub/splash/pull/131

kmike:
The described issue is fixed in master.

By default, page loading still fails though, because some requests to ftp resources take too long to finish. After filtering out those requests page renders fine. So I'm closing this issue.

There is a ticket for allowing to stop rendering without waiting for the full page load: https://github.com/scrapinghub/splash/issues/127

PR #159, opened 2015-01-19
Split viewport size and full-page rendering parameters

immerrr:
This is yet another attempt at #131.

The problem was that when "viewport=full" was selected,
viewport was not resized till the very rendering and inherited the
default window size which is small (640x480).  In presence of so called
fluid layouts this could significantly affect the appearance of the site (see #150 and #145)
and there was no way to tweak that initial size, because "viewport="
argument was already occupied with "viewport=full".

This PR does the following:

* ~~add "window_size=" query argument to "render.xyz" endpoints so that one can simultaneously select full viewport and tweak the window size~~

* ~~add "splash:set_window_size" function to Lua environment~~

* add `render_all=1` query argument to "render.xyz", so that one can simultaneously select full viewport and tweak the window size

* add `render_all` flag to `splash:png` function available in Lua scripts

* increase default window size from 640x480 to ~1024x768~ 1366x768

* if "splash:set_viewport('full')" fails in Lua environment, raise an
  error instead of silently falling back to default size

TODO:
- [x] bump Lua documentation
- [x] fix tests

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -90,20 +90,29 @@ allowed_domains : string : optional
 .. _arg-viewport:
 
 viewport : string : optional
-  View width and height (in pixels) of the browser viewport
-  to render the web page. Format is "<width>x<heigth>", e.g. 800x600.
-  It also accepts 'full' as value; viewport=full means that the whole
-  page (possibly very tall) will be rendered. Default value is 1024x768.
 
-  'viewport' parameter is more important for PNG rendering;
-  it is supported for all rendering endpoints because javascript
-  code execution can depend on viewport size.
+  View width and height (in pixels) of the browser viewport to render the web
+  page. Format is "<width>x<height>", e.g. 800x600.  Default value is 1366x768.
+
+  'viewport' parameter is more important for PNG rendering; it is supported for
+  all rendering endpoints because javascript code execution can depend on
+  viewport size.
+
+  For backward compatibility reasons, it also accepts 'full' as value;
+  ``viewport=full`` is semantically equivalent to ``render_all=1`` (see
+  :ref:`arg-render-all`).
+
+.. _arg-render-all:
+
+render_all : int : optional
+  Render the whole webpage (possibly very tall) when ``1``, possible values are
+  ``1`` and ``0``.  Default is ``0``.
 
 .. note::
 
-    viewport=full requires non-zero 'wait' parameter. This is
-    an unfortunate restriction, but it seems that this is the only
-    way to make rendering work reliably with viewport=full.
+    ``render_all=1`` requires non-zero :ref:`arg-wait` parameter. This is an
+    unfortunate restriction, but it seems that this is the only way to make
+    rendering work reliably with ``render_all=1``.
 
 .. _arg-images:
 
@@ -735,6 +744,9 @@ X-Splash-width : string
 
 X-Splash-height : string
   Same as :ref:`'height' <arg-height>` argument for `render.png`_.
+
+X-Splash-render-all : string
+  Same as :ref:`'render_all' <arg-render-all>` argument for `render.png`_.
 
 X-Splash-html : string
   Same as :ref:`'html' <arg-html>` argument for `render.json`_.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -105,8 +105,9 @@ viewport : string : optional
 .. _arg-render-all:
 
 render_all : int : optional
-  Extend the viewport to include the whole webpage (possibly very tall) when
-  ``1`` before rendering.  Possible values are ``1`` and ``0``.  Default is ``0``.
+  Possible values are ``1`` and ``0``.  When ``render_all=1``, extend the
+  viewport to include the whole webpage (possibly very tall) before rendering.
+  Default is ``render_all=0``.
 
 .. note::
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -56,8 +56,8 @@ wait : float : optional
   (defaults to 0). Increase this value if you expect pages to contain
   setInterval/setTimeout javascript calls, because with wait=0
   callbacks of setInterval/setTimeout won't be executed. Non-zero
-  'wait' is also required for PNG rendering when viewport=full
-  (see later).
+  :ref:`wait <arg-wait>` is also required for PNG rendering when doing
+  full-page rendering (see :ref:`render_all <arg-render-all>`).
 
 .. _arg-proxy:
 
@@ -100,17 +100,17 @@ viewport : string : optional
 
   For backward compatibility reasons, it also accepts 'full' as value;
   ``viewport=full`` is semantically equivalent to ``render_all=1`` (see
-  :ref:`arg-render-all`).
+  :ref:`render_all <arg-render-all>`).
 
 .. _arg-render-all:
 
 render_all : int : optional
-  Render the whole webpage (possibly very tall) when ``1``, possible values are
-  ``1`` and ``0``.  Default is ``0``.
+  Extend the viewport to include the whole webpage (possibly very tall) when
+  ``1`` before rendering.  Possible values are ``1`` and ``0``.  Default is ``0``.
 
 .. note::
 
-    ``render_all=1`` requires non-zero :ref:`arg-wait` parameter. This is an
+    ``render_all=1`` requires non-zero :ref:`wait <arg-wait>` parameter. This is an
     unfortunate restriction, but it seems that this is the only way to make
     rendering work reliably with ``render_all=1``.
 
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -959,7 +959,7 @@ Set the browser viewport size.
 * height - integer, requested viewport height in pixels.
 
 This will change the size of the visible area and subsequent rendering
-commands, e.g. :ref:`splash-png`, will produce an image with the specified
+commands, e.g., :ref:`splash-png`, will produce an image with the specified
 size.
 
 :ref:`splash-png` uses the viewport size.
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -610,10 +610,10 @@ not a size of an area screenshot is taken of. For example, if the viewport
 is 1024px wide then ``splash:png{width=100}`` will return a screenshot
 of the whole viewport, but an image will be downscaled to 100px width.
 
-To set the viewport size use :ref:`splash-set-viewport` method or use
-*render_all* argument.  *render_all=true* is equivalent to running
-``splash:set_viewport('full')`` just before the rendering and restoring the
-viewport size afterwards.
+To set the viewport size use :ref:`splash-set-viewport-size`,
+:ref:`splash-set-viewport-full` or *render_all* argument.  *render_all=true* is
+equivalent to running ``splash:set_viewport_full()`` just before the rendering
+and restoring the viewport size afterwards.
 
 If the result of ``splash:png()`` is returned directly as a result of
 "main" function, the screenshot is returned as binary data:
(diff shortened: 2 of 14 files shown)


splash/browser_tab.py
```diff
@@ -85,6 +87,10 @@ def _init_webpage(self, verbosity, network_manager, splash_proxy_factory, render
         self.web_view.setPage(self.web_page)
         self.web_view.setAttribute(Qt.WA_DeleteOnClose, True)
 
+        self.set_window_size(render_options.get_window_size())
```
kmike: The fact `render_options` is available to BrowserTab is a bug :) BrowserTab shouldn't be tied to GET/POST parameters extracted by a web server. 

Setting window size (and viewport size) here based on `render_options` means that the window size is set in scripts if `window_size` argument is passed to the /execute endpoint regardless of a script itself. I think we should require an explicit command for that. The fact `allowed_domains`, `proxy` and `filters` work in a similar way (see http://splash.readthedocs.org/en/latest/api.html#execute endpoint) is a bug/unfinished refactoring.

splash/tests/test_execute.py
```diff
@@ -1809,3 +1811,160 @@ def test_get_perf_stats(self):
...
+                splash:set_window_size(%s)
+            end
+            """ % (repr(size_str), ))
+
+        for size_str, errmsg in cases:
+            self.assertRaisesRegexp(RuntimeError, errmsg,
+                                    run_test, size_str)
+
+    def test_viewport_validation(self):
+        cases = [
```
kmike: pytest.mark.parametrize?

splash/qtrender_lua.py
```diff
@@ -452,7 +452,21 @@ def set_custom_headers(self, headers):
...
+        return sz.width(), sz.height()
+
+    @command()
+    def get_window_size(self):
+        sz = self.tab.web_view.size()
+        return sz.width(), sz.height()
+
+    @command()
+    def set_window_size(self, sz):
+        self.tab.set_window_size(sz)
```
kmike: splash:set_window_size(splash:get_window_size()) won't work

splash/tests/test_execute.py
```diff
@@ -1809,3 +1811,160 @@ def test_get_perf_stats(self):
...
+    alter_state(splash)
+    return get_dims(splash)
+end
+
+"""
+
+    def run_dims_test(self, alter_fn, **kwargs):
+        script = dedent(self.BASIC_SCRIPT) + "\n" + dedent(alter_fn)
+        resp = self.request_lua(script, kwargs)
+        if resp.ok:
```
kmike: This could be more consistent with the rest of the testing suite:
```
self.assertStatusCode(resp, 200)
return resp.json()
```

splash/tests/test_execute.py
```diff
@@ -1809,3 +1811,160 @@ def test_get_perf_stats(self):
...
+    def test_window_size(self):
+        out = self.run_dims_test("""
+        function alter_state(splash)
+            splash:set_window_size("2000x2000")
+        end
+        """)
+        self.assertEqual(out,
+                         {'inner': '2000x2000',
+                          'outer': '2000x2000',
+                          'client': '2000x2000'})
```
kmike: What about something like assertSizeAfter?
```
self.assertSizeAfter(
    'splash:set_window_size("2000x2000")', 
    {
        'inner': '2000x2000',
        'outer': '2000x2000',
        'client': '2000x2000'
    }
)
```

splash/browser_tab.py
```diff
@@ -144,7 +142,32 @@ def set_custom_headers(self, headers):
...
+        self.web_view.resize(size)
+        self.logger.log("window size is set to %sx%s" %
+                        (size.width(), size.height()), min_level=2)
+        # As self.web_page is attached to self.web_view, normally one would
+        # expect page's viewport to resize automatically.  This indeed happens
+        # as a reaction to QResizeEvent generated when doing resize().
+        #
+        # However, Qt only guarantees delivery of that event before the widget
+        # is next shown on the screen which means it won't happen as long as
+        # splash is running headless.  So, fall back to manual mode.
```
kmike: nice catch

kmike:
Please also check the test failures - they are real.

splash/tests/test_execute.py
immerrr: does not work: see the last paragraph of http://pytest.org/latest/unittest.html.

kmike: aha, i see

splash/tests/test_execute.py
immerrr: This raises a catchable RuntimeError which is checked against in some tests. Although, I suppose I could use `resp.json` in successful code path.

splash/tests/test_execute.py
immerrr: yup, that looks and reads nicer

splash/qtrender_lua.py
immerrr: I'm not a fan of forwarding QSize values into Lua runtime (C++ value wrapped as Python value wrapped as Lua value, eh?). I guess it should return `WxH` strings then to roundtrip successfully, it feels odd but could work.

kmike: I see 3 options:

1. "WxH" strings. Strings make it harder to get or change individual size components, I don't like them either.
2. Multiple return values: `local w, h = splash:get_window_size()`, then `splash:set_window_size(100, 200)` and `splash:set_window_size{width=w, height=h}`. It seems that `splash:set_window_size(splash:get_window_size())` still works here.
3. A table: `local sz = splash:get_window_size()`, then `splash:set_window_size{width=sz.width, height=sz.height}` or `splash:set_window_size(sz)`. The return value is a Lua table here, not a QSize object. 

(3) doesn't prevent `splash:set_window_size(100, 200)` syntax from being supported. So the difference between (2) and (3) boils down to the return value of `splash:get_window_size()`. I like (3) a bit more because it makes it harder to accidentally discard a size component, and because it allows for easier/more consistent error reporting if we need it (return nil, error). But (2) is also ok because it doesn't look like we need error reporting here (do we?).

kmike: The only Python object accessible from Lua is Splash object instance now, and it is possible to access only its whitelisted attributes. So forwarding QSize and accessing its fields to get width or height shouldn't even work.

splash/tests/test_execute.py
immerrr: which reminds me: why do we have this strange mix of unittest & pytest machinery? is this a result of an incomplete migration to pytest?

kmike: Yes. When implementing a migration from pytest to nose I did just enough changes to parallelize the tests using pytest-xdist (nose parallelization didn't work for some reason). 

I don't think it worths it to port all the tests from xUnit style to test_functions style now.

splash/tests/test_execute_emulation.py
```diff
@@ -60,21 +60,15 @@ class EmulatedMetaRedirectTest(Base.EmulationMixin, test_redirects.MetaRedirectT
 class EmulatedRenderPngTest(Base.EmulationMixin, test_render.RenderPngTest):
     script = 'main = require("emulation").render_png'
 
-    # TODO: default width and height are not applied
-    @pytest.mark.xfail
     def test_ok(self):
```
kmike: These tests just call super() methods; they were introduced in order to have something to mark with xfail. I think it is safe to delete these methods - parent's methods will be discovered and executed directly.

splash/tests/test_execute.py
```diff
@@ -1809,3 +1811,172 @@ def test_get_perf_stats(self):
...
+
+        def run_test(size_str):
+            self.get_dims_after('splash:set_viewport(%s)' % repr(size_str))
+
+        for size_str, errmsg in cases:
+            self.assertRaisesRegexp(RuntimeError, errmsg,
+                                    run_test, size_str)
+
+    def test_viewport_full(self):
+        w = int('1366x768'.split('x')[0])
```
kmike: isn't it a constant? And if you want to check for the default value explicitly then it is better to use defaults.WINDOW_SIZE

splash/qtrender_lua.py
```diff
@@ -449,10 +449,37 @@ def set_custom_headers(self, headers):
         self.tab.set_custom_headers(self.lua2python(headers, max_depth=3))
 
     @command()
+    def get_viewport_size(self):
+        sz = self.tab.web_page.viewportSize()
+        return {'width': sz.width(), 'height': sz.height()}
+
+    @command(table_argument=True)
```
kmike: Does the following work?
```
@command()
def set_viewport(self, width, height): 
    # ...
```
If it works, then it is better because it would allow both positional and named arguments. "Named arguments" in Lua is just passing a single table as an argument, and this is what we do anyways.

kmike: See http://splash.readthedocs.org/en/latest/scripting-tutorial.html#calling-splash-methods and https://github.com/scoder/lupa#python-callables

splash/browser_tab.py
```diff
@@ -719,12 +745,14 @@ def log(self, message):
...
     """ This class logs various events that happen with QWebPage """
-    def __init__(self, uid, web_page, verbosity):
+    def __init__(self, uid, verbosity):
         self.uid = uid
-        self.web_page = web_page
         self.verbosity = verbosity
+        self.web_page = None
 
-    def enable(self):
+    def add_web_page(self, web_page):
```
kmike: "add" implies it can be called several times, and all pages will be stored in some list. What about `set_web_page` or `init_web_page`?

immerrr: technically, nothing prevents one to add more pages, the only thing it does is connecting signals to their respective handlers.

kmike: right, but that's not what the code does

splash/tests/lua_modules/emulation.lua
```diff
@@ -28,9 +28,20 @@ function Splash:go_and_wait(args)
 
   self:set_images_enabled(self.args.images)
 
+  if args.window_size ~= nil then
+    w, h = string.match(args.window_size, '^(%d+)x(%d+)')
```
kmike: local w, h

splash/browser_tab.py
immerrr: not anymore :)

splash/browser_tab.py
immerrr: removed that, it now initialized browser window & viewport size with `defaults.VIEWPORT_SIZE`

splash/tests/test_execute_emulation.py
immerrr: removed

splash/tests/test_execute.py
immerrr: replaced

splash/qtrender_lua.py
immerrr: fixed the signature (and docs)

splash/tests/lua_modules/emulation.lua
immerrr: done

docs/api.rst
```diff
@@ -90,20 +90,29 @@ allowed_domains : string : optional
...
+
+  For backward compatibility reasons, it also accepts 'full' as value;
+  ``viewport=full`` is semantically equivalent to ``render_all=1`` (see
+  :ref:`render_all <arg-render-all>`).
+
+.. _arg-render-all:
+
+render_all : int : optional
+  Extend the viewport to include the whole webpage (possibly very tall) when
+  ``1`` before rendering.  Possible values are ``1`` and ``0``.  Default is ``0``.
```
kmike: 'when 1' doesn't read well.

docs/scripting-ref.rst
```diff
@@ -928,29 +932,71 @@ Example:
...
 
-* size - string, width and height of the viewport.
-  Format is ``"<width>x<heigth>"``, e.g. ``"800x600"``.
-  It also accepts ``"full"`` as a value; ``"full"`` means that the viewport size
-  will be auto-detected to fit the whole page (possibly very tall).
+* width - integer, requested viewport width in pixels;
+* height - integer, requested viewport height in pixels.
+
+This will change the size of the visible area and subsequent rendering
+commands, e.g. :ref:`splash-png`, will produce an image with the specified
```
kmike: `:ref:`splash-png`, will` - comma looks unnecessary here

docs/scripting-ref.rst
```diff
@@ -928,29 +932,71 @@ Example:
...
+         return {png=splash:png()}
+     end
+
+.. note::
+
+   This will affect ``window.innerWidth`` and ``window.innerHeight`` JS
+   variables and invoke ``window.onresize`` event callback.  However this will
+   only happen during the next asynchronous operation and :ref:`splash-png` is
+   notably synchronous, so if you have resized a page and want it to react
+   accordingly before taking the screenshot, use :ref:`splash-wait`.
```
kmike: :+1:

splash/qtrender_lua.py
```diff
@@ -448,11 +448,18 @@ def set_user_agent(self, value):
...
-    def set_viewport(self, size):
-        if size is None:
-            return
-        return self.tab.set_viewport(size)
+    def set_viewport_size(self, width, height):
+        self.tab.set_viewport('%dx%d' % (width, height))
+
+    @command()
+    def set_viewport_full(self):
+        self.tab.set_viewport('full')
```
kmike: It is documented to return the size.

splash/lua_modules/splash.lua
```diff
@@ -59,7 +59,7 @@ local function unpacks_multiple_return_values(func)
     -- Max allowed list size is 10; it is more than enough for
     -- functions which return multiple values. This trick is needed
     -- to handle `nil` as a first value correctly.
-    return table.unpack(func(...), 1, 10)
+    return table.unpack(func(...), 1, 2)
```
kmike: the comment above should be changed

splash/tests/test_execute.py
```diff
@@ -1809,3 +1810,141 @@ def test_get_perf_stats(self):
...
+            self.GET_DIMS_AFTER_SCRIPT % lua_script, **kwargs)
+
+    def assertSizeAfter(self, lua_script, etalon, **kwargs):
+        out = self.get_dims_after(lua_script, **kwargs)
+        self.assertEqual(out, etalon)
+
+    def test_get_viewport_size(self):
+        script = """
+        function main(splash)
+        w, h = splash:get_viewport_size()
```
kmike: local w, h - let's show a good example to users :)

docs/scripting-ref.rst
```diff
@@ -609,7 +610,10 @@ not a size of an area screenshot is taken of. For example, if the viewport
 is 1024px wide then ``splash:png{width=100}`` will return a screenshot
 of the whole viewport, but an image will be downscaled to 100px width.
 
-To set the viewport size use :ref:`splash-set-viewport` method.
+To set the viewport size use :ref:`splash-set-viewport-size`,
+:ref:`splash-set-viewport-full` or *render_all* argument.  *render_all=true* is
+equivalent to running ``splash:set_viewport_full()`` just before the rendering
+and restoring the viewport size afterwards.
```
kmike: I see that `render_all` argument is (implicitly) tested in render_png emulation script, but it seems the viewport restoring behaviour is not tested.

kmike:
Looks good!
It can be merged once a few minor issues are resolved.

docs/scripting-ref.rst
immerrr: Hmm, is it not a parenthetical expression? I did miss a comma though, according to s&w e.g. must be surrounded with commas from both sides.

kmike: it is splash:png who produces the image, right? but i see what you mean

docs/scripting-ref.rst
immerrr: added the test

splash/tests/test_execute.py
immerrr: fixed

splash/lua_modules/splash.lua
immerrr: done

splash/qtrender_lua.py
immerrr: yup, an omission. fixed.

immerrr:
Yaay :)

Issue #122, opened 2014-10-31
allow to yield to Lua from a JS callback


kmike:
We should support something like this; I think it's doable:

```lua
local res = splash:runjs([[
function main(splash){
    window.setTimeout(function(){
        splash.return("I'm a result!");
    }, 1000);
}
]])
-- res is now "I'm a result!"
```

``runjs`` function will probably need another name, it is already taken.

mehaase:
Hey Mikhail, I have a working implementation of this here: https://github.com/mehaase/splash/commit/ef04d75a2fc67f361e6e063432dc391f95767bd7. 

I chose the name `splash:runjs_async()`. Please look it over and let me know what you think. I can send a PR if you wish.

mehaase:
Oh, here is a sample lua test script to demonstrate:

    function main(splash)
      local url = splash.args.url

      splash:autoload(splash.args.js_source)
      assert(splash:go(url))

      local async_result, err = splash:runjs_async([[
        setTimeout(function () {
          lua_resume("Good things come to those who wait.");
        }, 2000);
      ]])
      print(async_result, err)

      local async_result, err = splash:runjs_async([[
        setTimeout(function () {
          lua_error("But sometimes bad things come to those who wait.");
        }, 2000);
      ]])
      print(async_result, err)

      local async_result, err = splash:runjs_async([[
        setTimeout(function () {
          lua_resume(["Return","multiple","arguments","in","a","list!"]);
        }, 2000);
      ]])
      print(async_result, err)

      local async_result, err = splash:runjs_async([[
        setTimeout(function () {
          lua_resume("Don't wait too long or you might get interr-");
        }, 2000);
      ]], 1)
      print(async_result, err)

      splash:stop()
      splash:set_viewport("full")

      return {
        png = splash:png{width=640},
      }
    end

When executed, this prints out (with the appropriate delays in between):

    Good things come to those who wait.     nil
    nil     error: But sometimes bad things come to those who wait.
    Return,multiple,arguments,in,a,list!    nil
    nil     error: Timed out while waiting for runjs_async()

Cheers!

kmike:
@mehaase looks great!
Could you please open a PR so that we can discuss it more easily? 
I have a lot of comments :)

kmike:
Fixed by https://github.com/scrapinghub/splash/issues/162.

Issue #161, opened 2015-01-20
Change the way errors are handled in Lua scripts


kmike:
In http://splash.readthedocs.org/en/1.3.1/scripting-tutorial.html#error-handling it is described how are errors handled in Splash Lua scripts:

> 1. for developer errors (e.g. incorrect function arguments) exception is raised;
> 2. for errors outside developer control (e.g. a non-responding remote website) status flag is returned: functions that can fail return ok, reason pairs which developer can either handle or ignore.

This means that by default script won't stop on error; to make code robust almost every statement needs to be wrapped in `assert()`:

```
assert(splash:go(splash.args.url))
assert(splash:wait(0.5))
```

Even in the documentation I was sometimes sloppy and haven't written these asserts. One example is  `wait_for` from http://splash.readthedocs.org/en/latest/scripting-tutorial.html#writing-modules:

```
function Splash:wait_for(condition)
    while not condition() do   -- condition function can't follow our conventions here
        self:wait(0.05)        -- assert is missing
    end
end
```

It is too easy to forget about error checking, and code that checks for errors looks uglier than code which doesn't check.

I propose to change the error handling convention and always raise an error in case of errors.

Pros:
* errors are not silently discarded;
* less code to write and read in common cases (?);
* it is possible to wrap several statements in pcall and handle errors once.

Cons:
* not ideomatic Lua;
* handling errors with pcall requires more work;

TODO: figure out how code using pcall will look like:
```
-- option 1, no-brainer - does it work?
local ok, res = pcall(splash:evaljs, "document.title")

-- option 2, ugly
local ok, res = pcall(splash.evaljs, self, "document.title")

-- option 3, verbose
local ok, res = pcall(function() 
    return splash.evaljs:("document.title")
end)
```

Any thoughts?

pawelmhm:
+1 

This is very important IMO. I just realized in some situations we take screenshots of 500 and 404 pages. Script like this

```lua
function main(splash)
            splash:go("http://httpbin.org/status/500")
            return splash:png()
end
```
returns perfectly nice 200 response from splash, and client gets screenshot of 500 without anyone noticing this. 

> I propose to change the error handling convention and raise an error both for developer errors and for error conditions outside developer control (e.g. 404 pages).

This sounds like a good idea for me, but just to clarify, which response codes are we talking about here? If splash makes 60 requests while rendering page, and 1-2 return 404 will error be raised? Or are we talking only about response codes for first initial request?

One thing about:

[...]

PR #162, opened 2015-01-21
Allow JavaScript callback to resume Lua coroutine

mehaase:
As we discussed here: https://github.com/scrapinghub/splash/issues/122

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -257,7 +257,8 @@ If a JavaScript function throws an error, it is re-throwed as a Lua error.
 To handle errors it is better to use JavaScript try/catch because some of the
 information about the error can be lost in JavaScript → Lua conversion.
 
-See also: :ref:`splash-runjs`, :ref:`splash-evaljs`, :ref:`splash-autoload`.
+See also: :ref:`splash-runjs`, :ref:`splash-evaljs`, :ref:`wait-for-resume`,
+:ref:`splash-autoload`.
 
 .. _splash-evaljs:
 
@@ -327,7 +328,8 @@ Compare:
         -- ...
     end
 
-See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`, :ref:`splash-autoload`.
+See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`, :ref:`wait-for-resume`,
+:ref:`splash-autoload`.
 
 .. _splash-runjs:
 
@@ -390,6 +392,124 @@ Compare:
 
         -- ...
     end
+
+See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`, :ref:`splash-autoload`,
+:ref:`wait-for-resume`.
+
+.. _wait-for-resume:
+
+splash:wait_for_resume
+----------------------
+
+Run asynchronous JavaScript code in page context. The Lua script will
+yield until the JavaScript code tells it to resume.
+
+**Signature:** ``msg, error = splash:wait_for_resume(snippet, timeout)``
+
+**Parameters:**
+
+* snippet - a string with a JavaScript source code to execute. This code
+  must include a function called ``main``. The first argument to ``main``
+  is an object that has the properties ``resume`` and ``error``. ``resume``
+  is a function which can be used to resume Lua execution. It takes an optional
+  string argument which will be returned to Lua in the ``ok`` return value.
+  ``error`` is a function which can be called with a required string value
+  that is returned in the ``error`` return value.
+* timeout - a float which determines (in seconds) how long to allow JavaScript
+  to execute before cancelling it and returning control to Lua. The timeout
+  must be strictly greater than zero and less than or equal to 60. The timeout
+  defaults to 5 seconds.
+
+**Returns:** ``msg, error`` pair. When the execution is successful
+``msg`` is either nil or contains a stringified JavaScript value. In case of
+timeout or JavaScript errors ``msg`` is ``nil`` and ``error`` contains
+an error message string.
+
+Example:
+
+.. code-block:: lua
+
+    local msg, error = splash:wait_for_resume([[
+        function main(splash) {
+            setTimeout(function () {
+                splash.resume("Hello, world!");
+            }, 3000);
+        }
+    ]])
+
+    print(msg, error)
+
+This trivial example shows how to transfer control of execution from Lua
+to JavaScript and then back to Lua. This command will tell JavaScript to
+sleep for 3 seconds and then return to Lua. Note that this is an async
+operation: the Lua event loop and the JavaScript event loop continue to run
+during this 3 second pause, but the Lua ``print()`` function will not be called
+until JavaScript calls ``splash.resume()``.
+
+.. note::
+
+    The ``main()`` function is required, and you will get an error if
+    you do not include it. The first argument to this function can have any
+    name you choose, of course. We will call it ``splash`` by convention in
+    this documentation.
+
+The example prints ``Hello, world!   nil`` on the console. The
+``Hello, world!`` is the string returned by ``splash.resume()`` and the ``nil``
+indicates that no error occurred.
+
+.. code-block:: lua
+
+    local msg, error = splash:wait_for_resume([[
+        function main(splash) {
+            setTimeout(function () {
+                splash.error("Goodbye, cruel world!");
+            }, 3000);
+        }
+    ]])
+
+    print(msg, error)
+
+This example shows a similar flow of execution as the previous example,
+but it uses ``splash.error()`` to return an error value instead of a
+success value. The example prints ``nil     error: Goodbye, cruel world!``.
+The ``nil`` indicates no ``msg`` was returned, and
+``error: Goodbye, cruel world!`` contains the error string returned from
+``splash.error()``.
+
+.. code-block:: lua
+
+    local msg, error = splash:wait_for_resume([[
+        function main(splash) {
+            setTimeout(function () {
+                splash.resume("Hello, world!");
+            }, 3000);
+        }
+    ]], 1)
+
+    print(msg, error)
+
+This example is similar to the first example, but here we have changed the
+``timeout`` argument from its default 5 seconds to just 1 second. Of course,
+our JavaScript code will not call ``splash.resume()`` for 3 seconds, which
+guarantees that it will time out. The example prints
+``nil     error: Timed out while waiting for resume() or error().``
+
+.. note::
+
+    Remember that the timeout t must satisfy ``0 < t <= 60`` seconds. You
+    will receive an error if you choose an invalid timeout value.
+
+.. code-block:: lua
+
+    local msg, error = splash:wait_for_resume([[
+        console.log('hello!');
+    ]])
+
+    print(msg, error)
+
+This final example shows an incorrect usage of ``splash:wait_for_resume()``:
+the JavaScript code does not contain a ``main()`` function. This example
+prints ``nil     error: wait_for_resume(): no main() function defined``,
 
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`, :ref:`splash-autoload`.
 
@@ -459,7 +579,8 @@ is not a constant.
 If :ref:`splash-autoload` is called multiple times then all its scripts
 are executed on page load, in order they were added.
 
-See also: :ref:`splash-evaljs`, :ref:`splash-runjs`, :ref:`splash-jsfunc`.
+See also: :ref:`splash-evaljs`, :ref:`splash-runjs`, :ref:`splash-jsfunc`,
+:ref:`wait-for-resume`.
 
 .. _splash-http-get:
 
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -576,10 +576,10 @@ have no effect, as shown in the next example.
 
     end
 
-The next example shows the effect of the ``timeout`` argument. We have changed
-``timeout`` argument from its default 5 seconds to just 1 second, but
-our JavaScript code will not call ``splash.resume()`` for 3 seconds, which
-guarantees that ``splash:wait_for_resume()`` will time out.
+The next example shows the effect of the ``timeout`` argument. We have set
+the ``timeout`` argument to 1 second, but our JavaScript code will not call
+``splash.resume()`` for 3 seconds, which guarantees that
+``splash:wait_for_resume()`` will time out.
 
 When it times out, ``result`` will be nil, ``error`` will contain a string
 explaining the timeout, and Lua will continue executing. Calling
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -497,8 +497,9 @@ in ``error``.
 
 .. note::
 
-    Remember that the timeout t must satisfy ``0 < t <= 60`` seconds. You
-    will receive an error if you choose an invalid timeout value.
+    The timeout must be >= 0. If the timeout is 0, then
+    ``splash:wait_for_resume()`` will never timeout (although Splash's
+    HTTP timeout still applies).
 
 .. code-block:: lua
 
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -412,40 +412,44 @@ yield until the JavaScript code tells it to resume.
   must include a function called ``main``. The first argument to ``main``
   is an object that has the properties ``resume`` and ``error``. ``resume``
   is a function which can be used to resume Lua execution. It takes an optional
-  string argument which will be returned to Lua in the ``value`` return value.
+  argument which will be returned to Lua in the ``value`` return value.
   ``error`` is a function which can be called with a required string value
   that is returned in the ``error`` return value.
-* timeout - a float which determines (in seconds) how long to allow JavaScript
-  to execute before cancelling it and returning control to Lua. The timeout
-  must be strictly greater than zero and less than or equal to 60. The timeout
-  defaults to 5 seconds.
+* timeout - a number which determines (in seconds) how long to allow JavaScript
+  to execute before forceably returning control to Lua. Defaults to
+  zero, which disables the timeout.
 
 **Returns:** ``value, error`` pair. When the execution is successful
-``value`` is either nil or contains a stringified JavaScript value. In case of
-timeout or JavaScript errors ``value`` is ``nil`` and ``error`` contains
-an error message string.
+``value`` contains the JavaScript value passed to
+``splash.resume(…)``. ``value`` will be nil if no argument is passed to
+``splash.resume()``. In case of timeout or JavaScript errors ``value``
+is ``nil`` and ``error`` contains an error message string.
 
-Example:
+Examples:
 
-.. code-block:: lua
-
-    local value, error = splash:wait_for_resume([[
-        function main(splash) {
-            setTimeout(function () {
-                splash.resume("Hello, world!");
-            }, 3000);
-        }
-    ]])
-
-    -- value is "Hello, world!"
-    -- error is nil
-
-This trivial example shows how to transfer control of execution from Lua
+The first, trivial example shows how to transfer control of execution from Lua
 to JavaScript and then back to Lua. This command will tell JavaScript to
 sleep for 3 seconds and then return to Lua. Note that this is an async
 operation: the Lua event loop and the JavaScript event loop continue to run
-during this 3 second pause, but the Lua ``print()`` function will not be called
-until JavaScript calls ``splash.resume()``.
+during this 3 second pause, but Lua will not continue executing the current
+function until JavaScript calls ``splash.resume()``.
+
+.. code-block:: lua
+
+    function main(splash)
+
+        local value, error = splash:wait_for_resume([[
+            function main(splash) {
+                setTimeout(function () {
+                    splash.resume();
+                }, 3000);
+            }
+        ]])
+
+        -- value is nil
+        -- error is nil
+
+    end
 
 .. note::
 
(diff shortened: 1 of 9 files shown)


splash/browser_tab.py
```diff
@@ -503,6 +505,29 @@ def runjs(self, js_source, handle_errors=True):
         js_source = "%s;undefined" % js_source
         self.evaljs(js_source, handle_errors=handle_errors)
 
+    def runjs_async(self, js_source, callback, errback, timeout=0):
+        """
+        An async variant of runjs.
+
+        It injects a callback function named <lua_resume> into the scope of
+        the user's script. When lua_resume() is called, the corresponding
+        Lua coroutine will resume.
```
kmike: Isn't is an async variant of evaljs? Some other ideas for the function name: 

* `evaljs_async`
* `wait_for_value` 
* `wait_for_return`

mehaase: I originally called it `evaljs_async`, but I changed it as I was working, because I think `evaljs` implies that we evaluate an expression (or group of expressions) and return the value of that expression (or value of the last expression). On the other hand, `runjs` does not care about any expressions. I agree it is perhaps confusing to overload the meaning of these terms.

`wait_for_return` makes sense to me. Also, what about `wait_for_callback`, `wait_for_resume`, or `yield_to_js`?

splash/browser_tab.py
```diff
@@ -768,3 +793,64 @@ def log(self, message, min_level=None):
 
         message = "[%s] %s" % (self.uid, message)
         log.msg(message, system='render')
+
+
+class _LuaResume(QObject):
```
kmike: +1 to creating such object.

I don't like that it has 'Lua' in its name, docstring, member names though: _LuaResume doesn't require Lua to work, and BrowserTab is not tied to Lua scripts in any way. If in some far future someone will add e.g. Python scripting support to Splash it should be possible to reuse this object.

This object provides a way to call one of the Python callbacks from the JS code, and ensures that

1. callback is called once, and only one callback is called;
2. allowed execution time is limited.

Any ideas for a better name? `RunOnceCallbackWrapper`?  <-- it is quite ugly, so better names are welcome :)

splash/browser_tab.py
```diff
@@ -768,3 +793,64 @@ def log(self, message, min_level=None):
...
+    """ Allows us to resume a Lua coroutine from a Javascript context. """
+
+    def __init__(self, parent, callback, errback, timeout=0):
+        self.name = _LuaResume.generate_name()
```
kmike: I think it should be self.generate_name(), otherwise @classmethod decortor is pointless

mehaase: Good catch. I will fix it.

splash/browser_tab.py
kmike: `wait_for_..` names depend on the API, check e.g.

```
local res = splash:wait_for_return([[
function main(splash){
    setTimeout(function(){
        splash.return("hello!");
    }, 100);
}
]])
```
(I'm using an old proposal with an explicit wrapper function)

splash/browser_tab.py
mehaase: Good point. It *is* tied to JavaScript, so what about `JsOneShotCallback`, `JsRunOnceCallback`, etc.?

splash/browser_tab.py
kmike: `wait_for_resume` or `yield_to_js` also look good

splash/browser_tab.py
kmike: `JsRunOnceCallback` reads like "JS callback that runs once" for me, but here we are wrapping Python callbacks

kmike:
The approach looks good!

Some general comments:

1. I think this function should handle errors like JS redirects. When an redirect happens window object gets cleared, and we can be certain the callback won't be called ever. It is better to fail fast in this case, with an error, without waiting for the full timeout.

2. Explicit wrapper function (as in original proposal) allows users to choose a name for the argument. Injecting `resume` variables is worse because these variables could shadow variables from the global scope, and users can't fix it.

3. Regarding error reporting with `(ok, res)` - it's tricky; maybe you can provide some feedback on https://github.com/scrapinghub/splash/issues/161?

4. Do you think it is possible to remove the _LuaResume object from `window` once it is passed to the function? Having these objects hanging around is not good - regular browsers don't do that.

5. When are _LuaResume objects destroyed, and are they destroyed at all? Interaction between Qt, PyQT and Python GC is often interesting :) Could you please add tracking of these objects here: https://github.com/scrapinghub/splash/blob/master/splash/utils.py#L40?

6. To merge it we need tests and docs.

7. How are JS exceptions handled? JS syntax errors? I think they shouldn't just timeout or hang forever. Is it possible to reuse some of the code from BrowserTab.evaljs? If a copy-paste is needed then shame on me - it'll be the third place we handle JS errors (BrowserTab.evaljs and qtrender_lua.Splash.jsfunc are two other places).

splash/browser_tab.py
kmike: .. but RunOnceCallbackWrapper is even worse :)

splash/browser_tab.py
kmike: re `evaljs_async` - a good point, the function doesn't evaluate an expression and return its value.

kmike:
Another question: are attributes of _LuaResume (other than methods marked with `@pyqtSlot()`) available to JS? Is there a way to prevent it? It'd be good to have a test case for this as well.

kmike:
A related question: what happens when JS code goes crazy and calls `lua_resume` with itself? Is the returned result empty, or is it a bound Python method, or is it some other object, or does it cause an early error?

mehaase:
I think I understand everything you're asking for, except for the "explicit wrapper function" (item 2 in your list). How do you think it should work? Your example is:

```
local res = splash:wait_for_return([[
function main(splash){
    setTimeout(function(){
        splash.return("hello!");
    }, 100);
}
]])
```

Should we require the user to wrap their code in `function main() {}`?

kmike:
@mehaase wraping their code in `function main(splash) {…}` makes "splash" a parameter passed to this function. This way it is clear that this parameter is not available to other JS code as a global variable, and users can use a different name for this parameter if there is an existing global "splash" JS variable they want to access.

docs/scripting-ref.rst
```diff
@@ -391,6 +393,124 @@ Compare:
...
+
+    local msg, error = splash:wait_for_resume([[
+        function main(splash) {
+            setTimeout(function () {
+                splash.resume("Hello, world!");
+            }, 3000);
+        }
+    ]])
+
+    print(msg, error)
```
kmike: Could you please replace it with a comment or a return value? 
I'm not 100% sure we'll keep 'print' in the sandboxed Splash version.

docs/scripting-ref.rst
```diff
@@ -391,6 +393,124 @@ Compare:
...
+
+This example is similar to the first example, but here we have changed the
+``timeout`` argument from its default 5 seconds to just 1 second. Of course,
+our JavaScript code will not call ``splash.resume()`` for 3 seconds, which
+guarantees that it will time out. The example prints
+``nil     error: Timed out while waiting for resume() or error().``
+
+.. note::
+
+    Remember that the timeout t must satisfy ``0 < t <= 60`` seconds. You
```
kmike: I think it makes sense to remove the upper timeout limit. Why should we preven users from waiting for more than 60s?

docs/scripting-ref.rst
```diff
@@ -391,6 +393,124 @@ Compare:
...
+
+**Signature:** ``msg, error = splash:wait_for_resume(snippet, timeout)``
+
+**Parameters:**
+
+* snippet - a string with a JavaScript source code to execute. This code
+  must include a function called ``main``. The first argument to ``main``
+  is an object that has the properties ``resume`` and ``error``. ``resume``
+  is a function which can be used to resume Lua execution. It takes an optional
+  string argument which will be returned to Lua in the ``ok`` return value.
```
kmike: `ok`?

docs/scripting-ref.rst
```diff
@@ -391,6 +393,124 @@ Compare:
...
+  ``error`` is a function which can be called with a required string value
+  that is returned in the ``error`` return value.
+* timeout - a float which determines (in seconds) how long to allow JavaScript
+  to execute before cancelling it and returning control to Lua. The timeout
+  must be strictly greater than zero and less than or equal to 60. The timeout
+  defaults to 5 seconds.
+
+**Returns:** ``msg, error`` pair. When the execution is successful
+``msg`` is either nil or contains a stringified JavaScript value. In case of
```
kmike: Why is the value stringified? Other functions try to return value converted to Lua objects when possible.

splash/browser_tab.py
```diff
@@ -503,6 +511,64 @@ def runjs(self, js_source, handle_errors=True):
...
+
+        The JavaScript must contain a method called `main()` that accepts
+        one argument. The first argument will be an object with `resume()`
+        and `error()` methods. The code _must_ call one of these functions
+        before the timeout or else it will be canceled.
+
+        Note: this cleans up the JavaScript global variable that it creates,
+        but QT seems to notice when a JS GV is deleted and it destroys the
+        underlying C++ object. Therefore, we can only delete the JS GV _after_
+        the user's code has called us back. This should change in QT5, since
```
kmike: What happens when splash.resume is called multiple times? Will it segfault (as splash object is deleted from window)?

splash/browser_tab.py
```diff
@@ -503,6 +511,64 @@ def runjs(self, js_source, handle_errors=True):
...
+        it will then be possible to specify a different object ownership
+        policy when calling addToJavaScriptWindowObject().
+        """
+
+        frame = self.web_page.mainFrame()
+        script_text = js_source.replace('\n', ' \\\n').replace('\"', '\\"')
```
kmike: `_WrappedJavascriptFunction.__call__` uses another method for string escaping: it encodes string to JSON and strips the quotes. I think it is better to use only one method of escaping. What are advantages of `.replace('\n', ' \\\n').replace('\"', '\\"')`?

docs/scripting-ref.rst
mehaase: I added that for Scraping Hub's benefit :) In a previous version, I used `timeout==0` to indicate no timeout at all. I was worried if the user's code never calls `resume()` or `error()`, then all the resources associated with that request will never be released, right? Or does Splash have a master request timeout? I don't want to rely on the client's timeout, since they can manipulate that to tie up server resources for a very long time.

Is it okay for the user to opt out of the timeout? Or do you still want the `>0` constraint?

kmike: There is a global timeout for the HTTP request (specified in 'timeout' GET/POST argument); if request takes too long it will be cancelled even if the script is waiting for something.

splash/browser_tab.py
```diff
@@ -503,6 +511,64 @@ def runjs(self, js_source, handle_errors=True):
...
+        callback_proxy = OneShotCallbackProxy(self, callback, errback, timeout)
+        frame.addToJavaScriptWindowObject(callback_proxy.name, callback_proxy)
+
+        wrapped = """
+        (function () {
+            try {
+                eval("%(script_text)s");
+            } catch (err) {
+                var main = function (splash) {
+                    splash.error(err);
```
kmike: hmm, I used err.toString() in `_WrappedJavascriptFunction.__call__` - does it produce a different result? If no, it makes sense to remove .toString() in other places.

splash/browser_tab.py
```diff
@@ -759,12 +825,72 @@ def on_initial_layout_completed(self):
...
         self.log("mainFrame().urlChanged %s" % qurl2ascii(url))
 
-    def log(self, message, min_level=None):
+    def log(self, message, min_level=None, encode=True):
         if min_level is not None and self.verbosity < min_level:
             return
 
-        if isinstance(message, unicode):
+        if encode and isinstance(message, unicode):
             message = message.encode('unicode-escape').decode('ascii')
```
kmike: Log messages are encoded in order to make then ascii-safe - why is thi option needed? I think we should either encode all messages or encode none.

splash/browser_tab.py
mehaase: This is related to your 'why stringified' question. QT slots like `resume()` and `error()` must declare the types of their arguments, and those two currently declare 1 `str` argument. Therefore, QT will stringify whatever is passed to `resume()` or `error()`. That's why `toString()` is necessary in other places, perhaps, but not here. I'll take another look at it just to confirm.

I'm going to see if QT will let me overload `resume()` so that it could accept different types of arguments. If I can do that, then I can convert those arguments to the proper Lua types.

kmike: Maybe declaring the slots with QObject will work?

mehaase: I did some research on it this morning. `QObject` does not work in the `pyqtSlot` decorator. QT will say:

> Error: incompatible type of argument(s) in call to resume(); candidates were
> resume(QObject*)

I get a similar message if I try Python's `object`.

> Error: incompatible type of argument(s) in call to resume(); candidates were
> resume(PyQt_PyObject)

I finally found that `QVariant` works: it allows QT to call the resume() method. It sometimes produces QT types (e.g. QString) that are not useful in Lua (b/c they are not converted to Lua types automatically) and break JSON serialization if you try to include one in your response. We can work around by adding logic to recursively convert QT types to built-in Python types, and that supports nil, boolean, number, and string types correctly.

The only hitch is that a JavaScript `Array` cannot be passed to a slot that expects `QVariant`. I get this error:

> Error: incompatible type of argument(s) in call to resume(); candidates were
> resume(QVariant)

I can add an overload of `resume()` that declares a `QList` argument:

    @pyqtSlot('QList<QVariant>', name='resume')
    def does_this_work(self):
        print('yes it works')

This works for int, bool, and array, but it breaks strings! If I pass a string:

> Error: ambiguous call of overloaded function resume(); candidates were
> resume(QVariant)
> resume(QList<QVariant>)

Arghhh... QT is frustrating.

I think the only way to make this work is to explicitly overload `resume()` for each C++ type that we support, e.g. `float`, `bool`, `void`, etc. `QList`, etc. This is a more verbose and a less flexible because we can't rely on QVariant's type conversion magic, but at least it works. Here's an example:

    @pyqtSlot
    @pyqtSlot('bool')
    @pyqtSlot('double')
    @pyqtSlot('QByteArray')
    ... etc ...
    def resume(self, value):
        ... convert value to built-in python type and then do resume stuff ...

The alternative is to modify the JS wrapper in `browser_tab.py:wait_for_resume()` to serialize arguments in JSON format and then decode that JSON in our Python handler. The code would be shorter and more general than using the overloads described above. I think it is hacky, but I found a post on [StackOverflow](http://stackoverflow.com/a/23550868/122763) that prefers using JSON. What do you think?

I implemented this today using overloads instead of JSON, but let me know if you prefer the JSON approach. I don't think it would be too hard to try it out, and the test coverage is good enough that I would quickly find out if it works well.

kmike: Ouch! 

The problem of using JSON from JS is that `JSON` variable may be overwritten by the webpage, so for me your existing solution looks better.

splash/browser_tab.py
```diff
@@ -768,3 +838,146 @@ def log(self, message, min_level=None):
...
+        """
+
+        list_ = list()
+
+        for i in range(0, len(dict_)):
+            list_.append(dict_[str(i)])
+
+        return list_
+
+    def _qt_object_to_py_object(self, qt_object):
```
kmike: There is [splash.qtutils.qt2py](https://github.com/scrapinghub/splash/blob/de93746525e1f92e1bfe3112ed4915fc6dd54497/splash/qtutils.py#L158) function. Do you think it can be reused (or improved and reused)?

splash/browser_tab.py
```diff
@@ -768,3 +838,146 @@ def log(self, message, min_level=None):
...
+        """
+        QT converts JS arrays and objects into QVariantMap, which it then
+        converts to Python dict.
+
+        The problem with this behavior is that we don't know if the dict
+        we received started life as a JS Array or a generic JS object. We
+        use a simple heuristic to determine if dict_ "looks like" a
+        list: the dict_ is a list if dict_.keys() contains all of the integers
+        in the range [0,n-1], where n is the number of keys in the dict_.
+        """
```
kmike: I think we shouldn't do this; if it is hard to support Array properly we may document that Array is unsupported for now.
The problem with auto-detection is that some Objects are converted to Arrays, so for the caller it is not clear what data format to expect.

mehaase: Agreed. It's also ambiguous for an empty array/object. If we explicitly disallow arrays, that actually simplifies a lot of the other code surrounding this. I will clean it up and update the docs.

splash/browser_tab.py
mehaase: Yes, that's very helpful! At a glance, I think it supports everything I need. I'll try it out today.

kmike:
@mehaase could you also check the test failures? `test_callback_proxy.py` tests are failing on Travis

mehaase:
OK, I removed the dict-to-list hack and my QT conversion code and I'm using qt2py instead. This also seems to have fixed the build.

splash/qtutils.py
```diff
@@ -166,7 +166,7 @@ def qt2py(obj, max_depth=100):
 
     # print(obj, obj.__class__)
 
-    if isinstance(obj, QString):
+    if isinstance(obj, (QString,QByteArray)):
         return unicode(obj)
```
kmike: Hmm, I'm not sure returning unicode(obj) is correct for QByteArray. It will likely try to decode the data using some encoding (latin1?), and this is incorrect if the data uses different encoding. Based only on QByteArray name, I think it should be bytes(obj), but I haven't tried it.

Where are these QByteArray objects from? Could you please add a test with non-latin text and a value which ends in a QByteArray?

kmike:
Re error handling: here the common pattern for error checking won't work well: `assert(splash:wait_for_resume([[...]]))` can raise an error even if `error` is nil,  when `splash:resume` returns undefined; to check for errors user must check the second argument, not the first. I had exactly the same issue with [splash:evaljs](http://splash.readthedocs.org/en/master/scripting-ref.html#splash-evaljs), and worked around it by raising a Lua error in case of JS errors, which is also not ideal because it differs from e.g. what splash:runjs does.

I also don't quite like that the errors are reported the same in case of syntax errors (i.e. user typos) and execution errors (i.e. errors which may happen because of e.g. webpage changes).

Any ideas on how to fix that? Add a note to the docs, make wait_for_resume raise and error when splash:error() is called, change splash:evaljs, implement #161?

kmike:
What do you think about the following?

1. If the initial script execution failed, raise an error. This includes: syntax errors, errors while setting up the callbacks. If setting up the callbacks fails then it is likely a developer error; silencing it by default will make scripts harder to debug.
2. When splash.error() is called, raise an error. Users who want to just indicate an error may return e.g. {"ok": false} objects.
3. Return only a single value.

But I'm not sure what to do about timeout errors.. On one hand, it is good for timeout error to raise an error because value no longer indicates an error. On other hand, if I'm not mistaken, catching it won't work because of https://github.com/scrapinghub/splash/issues/165. So this is actually a blocker for my proposal above :)

So a different proposal: fix (1), but don't do (2) and (3) - keep the existing (value, error) return value; in case of timeout errors return an error. 

To fix an issue with "undefined" being a nil value (and breaking the assert) we can return a table instead of just a value, `{result=...}`. On one hand, it is ugly. One the other hand, it allows to make the API more rich, e.g. by providing splash.set("foo", "bar") function which won't return, but will add `foo="bar"` to the result.

I think supporting undefined is nice because this looks like a common use case for `wait_for_resume`:

```
splash:wait_for_resume([[
    function main(splash){
        onSomeEvent(function(){
            splash:resume()
        }
    } 
]])
```

splash/qtutils.py
mehaase: Hmmm... if I decorate `resume()` with a `QString` argument, then `null` values in JavaScript will turn into empty strings `''` in Python. I decorated with `QByteArray` because it does not interfere with `null`. However I tried testing with non-ASCII characters this morning and QT converts from native JavaScript string encoding (UTF-16) to ASCII before it calls our slot method! So if JavaScript returns '你好', then the Python slot method sees '??' as its `value` argument. Argh...

One possible simplification (at the expense of making the API more complex) is to have multiple resume methods, each of which handles a specific type, e.g. `resume_number()`, `resume_array()`, etc. Or alternatively, we can force the caller to pass a specific data type, e.g. `resume()` must take a JavaScript object parameter. I know its not ideal, but QT's metatype system is so screwy that I think

In the meantime, I'll keep investigating how to correctly pass strings from JavaScript back to Python.

mehaase:
Re: error handling, I just read about Lua `assert` and now I understand the problem. There's a deeper issue, which is that we can really only detect JS syntax/execution errors that are directly inside `splash:wait_for_resume` callback. If the user's code runs a callback and that callback has an execution error, then our wrapper will never even notice.

Anyway, in the realm of things we can reasonably handle, what about requiring the user to return a value? If `splash.resume(undefined)` is not supported, then that might simplify string handling and it also means that Lua's `assert` would work correctly.

I'll keep thinking about it. At the moment I don't have any good answers. Lua is still new and strange to me.

splash/qtutils.py
kmike: I think that supporting unicode but returning null as '' is a reasonable tradeoff. And anyways, this is how it already works for splash:jsfunc, splash:runjs and splash:evaljs :) See http://splash.readthedocs.org/en/master/scripting-ref.html#js-lua-conversion-rules.

mehaase: I wasn't being precise when I said "null return values". I really meant "return undefined", which I agree is much more common than `return null`. I will try to make the behavior of this new API the same as the behaviors of existing APIs.

I just don't understand how it works in the existing APIs. [evaluateJavascript](http://qt-project.org/doc/qt-4.8/qwebframe.html#evaluateJavaScript) returns a QVariant, and according to [this](https://bugs.webkit.org/show_bug.cgi?id=105441), QVariant doesn't distinguish between `undefined` and `null`. I'll have to do some digging to figure out what's going on here.

mehaase:
I implemented the "return a table" idea and it works _great_. It solves several problems simultaneously: works with assert(), allows us to correctly return an array from JS to Lua, and correctly distinguish undefined from null. I added `splash.set` while I was at it. Thanks for both good ideas!

I'll revisit error handling next.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,223 @@ Compare:
...
+  argument which will be returned to Lua in the ``value`` return value.
+  ``error`` is a function which can be called with a required string value
+  that is returned in the ``error`` return value.
+* timeout - a number which determines (in seconds) how long to allow JavaScript
+  to execute before forceably returning control to Lua. Defaults to
+  zero, which disables the timeout.
+
+**Returns:** ``value, error`` pair. When the execution is successful
+``value`` is a table that contains the JavaScript value passed to
+``splash.resume(…)``. It also contains any additional key/value pairs
```
kmike: I think we should explain under which key the value is.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,223 @@ Compare:
...
+
+        local value, error = splash:wait_for_resume([[
+            function main(splash) {
+                setTimeout(function () {
+                    splash.resume([1, 2, 'red', 'blue']);
+                }, 3000);
+            }
+        ]])
+
+        -- value is {return=[1, 2, 'red', 'blue']}
```
kmike: 'return' is a reserved word in lua, and this causes inconveniences - e.g. you can't write `value.return` to get the value.

kmike:
> I implemented the "return a table" idea and it works great. It solves several problems simultaneously: works with assert(), allows us to correctly return an array from JS to Lua, and correctly distinguishes undefined from null. I added splash.set while I was at it. Thanks for both good ideas!

> I'll revisit error handling next.

That's great!

mehaase:
I'm looking at error handling right now, and I'm stumped. I can't find any examples of async methods in browser_tab.py that raise an error. They all call the errback instead. The methods that do raise error are all synchronous.

I tried raising a JsError in `BrowserTab#wait_for_resume()` and it causes Splash to return HTTP 500, not HTTP 400, because `BrowserTab#wait_for_resume()` runs in a different stack frame than `Splash#wait_for_resume()`.

It seems that I need to do something quite different from what's been done before, right? I'm not sure what that is, yet...

mehaase:
I removed the name 'return'.

I also tried implementing the type of error handling you're asking for. I'm not sure its exactly what you're looking for. I put it in a separate commit so that its easy to revert if needed. Let me know what you think.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+
+**Signature:** ``result, error = splash:wait_for_resume(snippet, timeout)``
+
+**Parameters:**
+
+* snippet - a string with a JavaScript source code to execute. This code
+  must include a function called ``main``. The first argument to ``main``
+  is an object that has the properties ``resume`` and ``error``. ``resume``
+  is a function which can be used to resume Lua execution. It takes an optional
+  argument which will be returned to Lua in the ``result`` return value.
```
kmike: I think it should be ``result.value`` or something like this - an optional argument is not returned as a value of ``result``.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+    end
+
+``result`` is set to an empty table to indicate that nothing was returned
+from ``splash.resume``. You can use ``assert(splash:wait_for_resume(…))``
+even when JavaScript does not return a value because the empty table signifies
+success to ``assert()``.
+
+.. note::
+
+    The ``main()`` function is required, and you will get an error if
```
kmike: There are two 'main' functions in the example above. The note applies to both, but it could make sense to stress that the note is about JS `main` function.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+
+        local result, error = splash:wait_for_resume([[
+            function main(splash) {
+                setTimeout(function () {
+                    splash.resume([1, 2, 'red', 'blue']);
+                }, 3000);
+            }
+        ]])
+
+        -- result is {return=[1, 2, 'red', 'blue']}
```
kmike: vaule=

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+        local result, error = splash:wait_for_resume([[
+            function main(splash) {
+                setTimeout(function () {
+                    splash.set("foo", "bar");
+                    splash.resume("ok");
+                }, 3000);
+            }
+        ]])
+
+        -- result is {foo="bar", return="ok"}
```
kmike: value=

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+            function main(splash) {
+                setTimeout(function () {
+                    splash.resume("ok");
+                    splash.resume("still ok");
+                    splash.error("not ok");
+                }, 3000);
+            }
+        ]])
+
+        -- result is "ok"
```
kmike: result is {value="ok}

splash/lua_modules/splash.lua
```diff
@@ -27,6 +27,23 @@ end
...
 
 --
+-- Allows an async function to raise a Lua error by returning ok, err, raise.
+--
+-- If raise is true and ok is nil, then an error will be raised using res
+-- as the reason.
+--
+local function raises_async(func)
+  return function(...)
+    ok, err, raise = func(...)
```
kmike: local ok, err, raise =

kmike:
Besides a few minor issues and a merge conflict it looks really good!

mehaase:
Great! I'll clean it up.

For the merge conflict, do you prefer that I merge master into my branch or rebase onto master?

kmike:
> For the merge conflict, do you prefer that I merge master into my branch or rebase onto master?

I'm fine with both options, no preference. Please do what you like more :)

kmike:
Ah, there is also http://splash.readthedocs.org/en/latest/scripting-tutorial.html#living-without-callbacks doc chapter which has a list of async functions; wait_for_resume should be mentioned there.

mehaase:
I fixed the docs this morning, added that `local` keyword, and merged in master.

docs/scripting-ref.rst
```diff
@@ -391,7 +393,225 @@ Compare:
...
+            }
+        ]])
+
+        -- result is {value="ok"}
+        -- error is nil
+
+    end
+
+The next example shows the effect of the ``timeout`` argument. We have changed
+``timeout`` argument from its default 5 seconds to just 1 second, but
```
kmike: Default value is zero now.

splash/lua_modules/splash.lua
```diff
@@ -96,6 +113,14 @@ function Splash.create(py_splash)
       command = yields_result(command)
     end
 
+    if opts.can_raise_async then
+      command = raises_async(command)
+    end
+
+    if opts.multiple_return_values then
```
kmike: if I'm not mistaken, opts.multiple_return_values and unpacks_multiple_return_values was deleted in master.

kmike:
Great work @mehaase!

mehaase:
Haha, phew... Time to open up a beer :beer:

PR #164, opened 2015-01-22
Refactor handling of multiple return values from Lua commands

immerrr:
This PR revisits returning multiple values from splash commands exported to Lua.

* It drops `unpacks_multiple_return_values` flag as it is unnecessary: tuples
  will be returned as Lua pseudo-tuples, lists will be returned as tables following `lupa` convention

* pseudo-tuples will contain the exact number of elements that were in
  their Python equivalent, regardless of None/nil values.

TODO:
- [x] ~~maybe there's a way to test arbitrary commands, like those that return `None`, lists or dicts as one of their return values~~ not worth it
- [x] I couldn't verify that tracebacks do not include code from `splash.lua`, because of issues with pcall (they deserve a separate issue)

### DIFF
--- a/splash/lua_modules/splash.lua
+++ b/splash/lua_modules/splash.lua
@@ -1,15 +1,35 @@
+-- This function works very much like standard Lua assert, but:
 --
--- Python Splash commands return `ok, result` pairs; this decorator
--- raises an error if "ok" is false and returns "result" otherwise.
+-- * the first argument is the stack level to report the error at (1 being
+--   current level, like for `error` function)
+-- * it strips the flag if it evaluates to true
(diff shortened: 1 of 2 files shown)


immerrr:
Also, something is wrong with error handling...

immerrr:
Error handling fixed, the question about testing arbitrary commands still stands.

kmike:
It seems that to test arbitrary commands we need to refactor Splash and LuaRender and extract something like _PyLuaObject and _Dispatcher from them. I think it doesn't worths it now; we don't have arbitrary commands ;)

immerrr:
Then there's nothing to add to it. Please review at your convenience :)

PR #168, opened 2015-01-23
Fix width scaling

### LINKED ISSUE #166: PNG: width parameter doesn't scale but crops the image?
Seems that I've broken it somewhere during recent fiddling with scaling/cropping/tiling code.

Reproduceable with http://127.0.0.1:8050/render.png?url=http://www.google.com&wait=0.5&render_all=1&width=600

immerrr:
This should fix #166.

The old code was broken, because painter.setClipRect operates in input
coordinates, but it was fed with clip region in outer coordinates.

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -969,8 +969,8 @@ Example:
 .. code-block:: lua
 
      function main(splash)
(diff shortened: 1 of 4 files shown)


kmike:
thanks!

PR #171, opened 2015-01-25
Make Lua viewport resize commands trigger a page relayout

### LINKED ISSUE #167: Lua: unexpected interaction between set_viewport_size and set_viewport_full
When there's a site that comfortably fits into the viewport horizontally, its `preferredContentsSize().width()` matches that of the viewport, so `set_viewport_full` usually only changes the height of the viewport.

However, if one does
```lua
splash:set_viewport_size(1920, 1080)
splash:wait(0.1)
splash:set_viewport_size(640, 480)
w,h = splash:set_viewport_full()
```
`w` and `h` will somewhat unexpectedly be `1920` and `1080` respectively. Most likely this happens because page layouting is asynchronous and hence the code that should relayout the page and change `mainFrame()->contentsSize` is not invoked between the last-but-one and the last lines.

immerrr:
This PR makes `splash:set_viewport_size` and `splash:set_viewport_full` operations synchronous w.r.t. Lua runtime. This should fix #167.

~~This is achieved by adding `qtutils.process_pending_events` function and using it in respective `Splash` commands. Internally visible `BrowserTab.set_viewport` is unchanged, because it seems desirable for the page not to react to temporary viewport resizes performed by `png` rendering methods, so we have to maintain some way to get the old behaviour. Also, python code unlike Lua scripts has access to `process_pending_events` which can be used whenever it is necessary.~~

This is achieved by using `QWebPage.setPreferredContentsSize` that triggers a synchronous relayout of the page's elements (but does not propagate values to JS). `process_pending_events`-based solution was technically synchronous, too, but it could for example carry out a redirect in response to an event  scheduled for the same time.

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -619,9 +619,9 @@ specified one without resizing the content.  The region created by such
 extension is transparent.
 
 To set the viewport size use :ref:`splash-set-viewport-size`,
-:ref:`splash-set-viewport-full` or *render_all* argument.  *render_all=true* is
-equivalent to running ``splash:set_viewport_full()`` just before the rendering
-and restoring the viewport size afterwards.
+:ref:`splash-set-viewport-full` or *render_all* argument.  ``render_all=true``
+is equivalent to running ``splash:set_viewport_full()`` just before the
+rendering and restoring the viewport size afterwards.
 
 *scale_method* parameter must be either ``'raster'`` or ``'vector'``.  When
 ``scale_method='raster'``, the image is resized per-pixel.  When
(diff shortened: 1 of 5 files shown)


splash/qtrender_lua.py
```diff
@@ -456,10 +456,13 @@ def get_viewport_size(self):
     @command()
     def set_viewport_size(self, width, height):
         self.tab.set_viewport('%dx%d' % (width, height))
+        process_pending_events()
 
     @command(multiple_return_values=True)
     def set_viewport_full(self):
-        return list(self.tab.set_viewport('full'))
+        size = self.tab.set_viewport('full')
+        process_pending_events()
```
kmike: Is it different from e.g. waiting for 0 seconds?
I'm asking because the question is what may happen during processEvents call. For example, can the page redirect to a different URL?

immerrr: technically, it could, if it has been scheduled before the control was lended over to Lua script. or if an event that would cause a redirect has been received before that.

immerrr:
I've found a way to trigger a synchronous relayout of the page without processing any pending events which should be perfectly safe w.r.t redirects and other asynchronous activities.

kmike:
Nice hack :)

Issue #169, opened 2015-01-23
Rendering doesn't work correctly when viewport size != image size


kmike:
Example: http://127.0.0.1:8050/render.png?url=http://google.com&viewport=800x600&width=700

![render-google](https://cloud.githubusercontent.com/assets/107893/5882630/fd09337a-a36c-11e4-9b7f-df39cbc91c1e.png)

When width=800 (matching the viewport size) the result looks fine.

immerrr:
It appears that some of 3d decorations cannot catch up with the resize of the viewport.

I cannot seem to reproduce this, neither on my local ubuntu nor on docker image of current master. Could you share your versions and/or logs for this situation?

kmike:
Yeah, the example in ticket description works fine in Docker. 
I'm using OS X 10.10.1, Qt 4.8.6, PyQt 4.11.3, WebKit 534.34, sip 4.16.5.

But if you add some wait time there will be a rendering issue in Docker as well:
http://192.168.59.103:8050/render.png?url=http://google.com&viewport=800x600&width=700&wait=0.5
![render-docker](https://cloud.githubusercontent.com/assets/107893/5888054/038affda-a412-11e4-83a1-9fe0484144b3.png)

Note the button in upper-right. When width=800 the screenshot becomes correct.

immerrr:
Yup, that I can reproduce.

There is indeed something wrong with rendering of that button: as the scale goes down, the button changes its size drastically while staying in bottom-right corner. Also its text doesn't change size: http://imgur.com/a/0uY5Z#0

kmike:
Fixed by #172 by switching to raster rendering by default.

PR #172, opened 2015-01-26
Make resizing raster-based again and offer "scale_method=vector" as a replacement

immerrr:
This PR enables back raster resizing that was replaced with "vector" one in 97f59adad2642d1e552206a3b0e95d9969b9673d, because of a seemingly unsolvable issue described [here](https://github.com/scrapinghub/splash/issues/169#issuecomment-71326285).

* new `scale_method` parameter added to HTTP & Lua API with values `raster` (default) and `vector`

* PNG rendering has undergone refactoring that among other things paves the path to "clip_rect" functionality.

Benchmark results are available [here](https://gist.github.com/immerrr/8f996b10f7a4d273566e). In short:
- this implementation of raster resize is faster than the pre-32k-fix because of decreased PNG compression level and some cropping of rendering area
- old vector-based resize, as expected, runs circles around raster-based, both in terms of CPU and RSS
but the two implementations are on par in no-resize and resized-image-heavy-site cases


splash/tests/test_render.py
```diff
@@ -589,6 +586,18 @@ def _do_same_requests(self, url, params, other_format):
         return r1, r2
 
 
+class RenderVectorPngTest(RenderPngTest):
+    def request(self, query, *args, **kwargs):
+        query.setdefault('scale_method', 'vector')
+        return super(RenderVectorPngTest, self).request(query, *args, **kwargs)
+
+
+class RenderVectorJsonTest(RenderJsonTest):
```
kmike: Why is it necessary to run render.json tests with scale_method=vector, what do they check which is not checked in render.png tests?

immerrr: I was trying to ensure that scale_method argument is actually passed through.

splash/qtutils.py
```diff
@@ -254,171 +256,380 @@ def swap_byte_order_i32(buf):
...
+    logger.log("png render: rendering %s of the web page" % web_rect,
+               min_level=2)
+    logger.log("png render: rendering into %s of the canvas" % render_rect,
+               min_level=2)
+    logger.log("png render: canvas size=%s" % canvas_size, min_level=2)
+
+    to_paint = render_rect.intersected(QRect(QPoint(0, 0), canvas_size))
+    # QPainter has several issues when its rendering area is more than 2**15 in
+    # any dimension.
+    need_tiling = max(to_paint.width(), to_paint.height()) >= (1 << 15)
```
kmike: `2**15` reads better :)

Also, the comment says "more than", while the code uses ">=". I don't care about it though, ">=" is fine to be on a safe side.

splash/qtutils.py
```diff
@@ -254,171 +256,380 @@ def swap_byte_order_i32(buf):
     return arr.tostring()
 
 
+def render_qwebpage(web_page, logger=None, width=None, height=None,
```
kmike: What do you think about moving all render-related functions and classes into a dedicated module? Something like qtutils_render.py.

kmike:
The code looks good, waiting for the benchmark results.

splash/qtutils.py
```diff
@@ -254,171 +256,380 @@ def swap_byte_order_i32(buf):
...
+        logger.log("png render: image is trimmed vertically")
+        hcut = image_size.height()
+        if in_viewport.size() == out_viewport.size():
+            # If no resizing then we need to render exactly the requested
+            # number of pixels vertically.
+            hrender = hcut
+        else:
+            # If there's a resize, there will be interpolation.  We must ensure
+            # that both pixels that contribute to the color of the last row of
+            # the resulting image are rendered.
```
kmike: What if more than 2 pixels contribute to the last row? is the formula above still valid?

immerrr: No more than 2 pixels in any dimension contribute to the value of an output pixel in bilinear interpolation algorithm. Bicubic or Lanczos filters have bigger windows, so we'll need to update the formula accordingly before using one of those.

immerrr: If the upscaled pixel lands on the original one perfectly, of course, there will be no second pixel, but this should be handled by this formula. I'm going to test this, just to be sure.

splash/qtutils.py
immerrr: Turns out, 32766 is the last safe value, updated the code.

immerrr:
I've updated the first message with [the results of the benchmark](https://gist.github.com/immerrr/8f996b10f7a4d273566e). In short:
- this implementation of raster resize is faster than the pre-32k-fix because of decreased PNG compression level and some cropping of rendering area
- old vector-based resize, as expected, runs circles around raster-based, both in terms of CPU and RSS
but the two implementations are on par in no-resize and resized-image-heavy-site cases

kmike:
Looks good!

PR #185, opened 2015-02-27
WIP: benchmark suite

### LINKED ISSUE #184: Basic benchmark suite
We need basic benchmark suite for Splash, something that renders a few typical web pages, hosted locally (to avoid network latencies affecting the results). Just a few (5-10) diverse enough websites should be enough to get started.

Goals for this benchmark in order of priority:
1. compare Splash performance reliable on different server configurations/deployments
2. find out typical render metrics, like average/peak memory consumption. This should help to figure out the optimal SLASH_SLOTS value for different server configurations
3. compare Splash performance with other tools (see #99)

immerrr:
This should close #184.

### DIFF
--- a/splash/benchmark/benchmark.py
+++ b/splash/benchmark/benchmark.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+
+"""
+Splash benchmark script.
+
+It takes a directory downloaded with splash & httrack, fires up a static file
+server and runs a series of requests via splash on those downloaded pages.
+
+"""
+
+import logging
+import random
+from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser
+from glob import glob
+from multiprocessing.pool import ThreadPool
+
+import requests
+from splash.file_server import serve_files
+from splash.tests.utils import SplashServer
+
+PORT = 8806
+#: URLs to benchmark against.
+PAGES = glob('localhost_8806/*.html')
+#: Combinations of width & height to test.
+WIDTH_HEIGHT = [(None, None), (500, None), (None, 500), (500, 500)]
+# XXX: add benchmark of different API endpoints.
+SPLASH_LOG = 'splash.log'
+
+parser = ArgumentParser(description=__doc__,
+                        formatter_class=ArgumentDefaultsHelpFormatter)
+parser.add_argument('--seed', type=int, default=1234, help='PRNG seed number')
+parser.add_argument('--thread-count', type=int, default=1,
+                    help='Request thread count')
+parser.add_argument('--request-count', type=int, default=10,
+                    help='Benchmark request count')
+
+
+def generate_requests(splash, args):
+    log = logging.getLogger('generate_requests')
+    log.info("Using pRNG seed: %s", args.seed)
+    rng = random.Random(args.seed)
+    for i in xrange(args.request_count):
+        page = rng.choice(PAGES)
+        width, height = rng.choice(WIDTH_HEIGHT)
+        url = 'http://localhost:%d/%s' % (PORT, page)
+        yield (i + 1, args.request_count,
+               {'url': splash.url('render.png'),
+                'params': {'url': url, 'width': width, 'height': height}})
+
+
+def parallel_map(func, iterable, thread_count):
+    if thread_count == 1:
+        return map(func, iterable)
+    else:
+        pool = ThreadPool(thread_count)
+        return pool.map(func, iterable)
+
+
+def invoke_request(invoke_args):
+    log = logging.getLogger('bench_worker')
+    req_no, total_reqs, kwargs = invoke_args
+    log.info("Initiating request %d/%d: %s", req_no, total_reqs, kwargs)
+    return requests.get(**kwargs)
+
+
+def main():
+    args = parser.parse_args()
+    logging.basicConfig(level=logging.DEBUG)
+
+    with SplashServer(logfile=SPLASH_LOG,
+                      extra_args=['--disable-lua-sandbox',
+                                  '--disable-xvfb',
+                                  '--max-timeout=600']) as splash, \
+         serve_files(PORT):
+        parallel_map(invoke_request, generate_requests(splash, args),
+                     args.thread_count)
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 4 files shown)


splash/benchmark/file_server.py
```diff
@@ -0,0 +1,31 @@
+import SimpleHTTPServer
```
kmike: It could be better to use Twisted (or some other async framework) because SimpleHTTPServer is single-threaded, and Splash can download multiple resources in parallel. I wonder if it is important for tests, maybe not. 

With Twisted we can also simulate conditions like non-responding servers, delayed responses, etc.

kmike:
I think to fix tests it could be enough to add benchmarks here: https://github.com/scrapinghub/splash/blob/master/conftest.py

splash/benchmark/benchmark.py
```diff
@@ -0,0 +1,169 @@
...
+            'width': kwargs['params']['width'],
+            'height': kwargs['params']['height']}
+
+
+def main():
+    log = logging.getLogger("benchmark")
+    args = parser.parse_args()
+    logging.basicConfig(level=logging.DEBUG)
+
+    splash = SplashServer(
```
kmike: It is better to allow to set Splash server address from the command line. This way it'll be easier to compare performance of different versions of Splash, or performance of dockerized Splash (maybe several dockerized Splash'es which use different base operating systems and/or qt/... versions).

We should really allow to run the same benchmark against different Splash versions. If the benchmark can only be executed against current Splash version we can't just checkout to older Splash version and run a benchmark again - checking out old Splash could also change the benchmark version, so we'll run a different benchmark.

splash/benchmark/download_sites.py
```diff
@@ -0,0 +1,107 @@
...
+            out['html'], ('script',))
+        root = html.fromstring(out['html'], parser=html.HTMLParser(),
+                               base_url=final_url)
+        try:
+            head = root.xpath('./head')[0]
+        except IndexError:
+            head = html.Element('head')
+            root.insert(0, head)
+        head.insert(0, html.Element('base', {'href': final_url}))
+        head.insert(0, html.Element('meta', {'charset': 'utf-8'}))
```
kmike: what if there was meta charset element already?

splash/benchmark/download_sites.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def preprocess_main_page(sites_dir, url):
+    out = json.loads(lua_runonce(SCRIPT_HTML, url=url,
+                                 splash_args=['--disable-lua-sandbox',
+                                              '--disable-xvfb',
+                                              '--max-timeout=600'],
+                                 timeout=600.,))
+    final_url = urlsplit(out['url'])._replace(query='', fragment='').geturl()
+    if not w3lib.html.get_base_url(out['html']):
```
kmike: Does it add `<base>` element? if so, maybe move this code to a function named appropriately (e.g. `_html_add_base`)?

splash/benchmark/download_sites.py
kmike: why isn't charset fixed if there is a `<base>` element?

immerrr: Fixed, and fixed.

splash/benchmark/download_sites.py
immerrr: This function does several things, I've updated the docstring to spell them out for clarity.

immerrr:
@kmike is there anything else that prevents merging this?

kmike:
I can't get it work :) 

* splash.benchmark and splash.tests imports are undefined when you execute code from benchmarks folder because these packages are not distributed with Splash.

* if splash.benchmark and splash.tests are added to packages in setup.py download_sites.py raises an error while trying to start proxy server (Splash starts fine with `python -m splash.server`):

```
[-] Starting factory <splash.proxy_server.SplashProxyServerFactory instance at 0x114109710>
  File "/usr/local/Cellar/python/2.7.5/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/local/Cellar/python/2.7.5/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/kmike/envs/splash/lib/python2.7/site-packages/splash-1.4-py2.7.egg/splash/tests/mockserver.py", line 725, in <module>
    run(opts.http_port, opts.https_port, opts.proxy_port, not opts.quiet)
  File "/Users/kmike/envs/splash/lib/python2.7/site-packages/splash-1.4-py2.7.egg/splash/tests/mockserver.py", line 698, in run
    sslport = reactor.listenSSL(sslport_num, factory, ssl_factory())
  File "/Users/kmike/envs/splash/lib/python2.7/site-packages/splash-1.4-py2.7.egg/splash/tests/mockserver.py", line 673, in ssl_factory
    return ssl.DefaultOpenSSLContextFactory(pem, pem)
  File "/Users/kmike/envs/splash/lib/python2.7/site-packages/twisted/internet/ssl.py", line 104, in __init__
    self.cacheContext()
  File "/Users/kmike/envs/splash/lib/python2.7/site-packages/twisted/internet/ssl.py", line 113, in cacheContext
    ctx.use_certificate_file(self.certificateFileName)
OpenSSL.SSL.Error: [('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'FILE_CTRL', 'system lib'), ('SSL routines', 'SSL_CTX_use_certificate_file', 'system lib')]
```

* I'm not sure how I did it, but download_sites.py was working before I installed Splash, BUT it put nothing to `splash/benchmark/sites/localhost_8806/` folder, so benchmark found zero websites.

I think we shouldn't distribute splash.benchmark with other Splash code (it shouldn't be installed by installing Splash). What does it take to move benchmarks folder to the root?

immerrr:
> it shouldn't be installed by installing Splash

if not, I'm not sure about organizing the code. The necessary subset of SplashServer functionality can probably be copy-pasted. But for the rest, there are three separate runnable scripts there: the downloader, the file server & the benchmark runner. they all use code from the file server module, so at least that module must be importable.

immerrr:
it worked ok for me in `pip install -e .` mode.

PR #187, opened 2015-03-05
Dockerfile: put provisioning into a separate script

immerrr:
This way the commands can be easier executed manually.

I am not sure what to do with `provision.sh` in the container, so I've just put into `/tmp`.


kmike:
LGTM.
Could you please also move the Dockerfile to /dockerfiles/splash folder?

dockerfiles/splash/provision.sh
```diff
@@ -0,0 +1,123 @@
...
+remove_builddeps -- remove build-dependencies
+remove_extra -- remove files that are unnecessary to run Splash
+
+EOF
+}
+
+prepare_install () {
+    # Prepare docker image for installation of packages, docker images are
+    # usually stripped and aptitude doesn't work immediately.
+    sed 's/main$/main universe/' -i /etc/apt/sources.list && \
```
dangra: A better bash idioms to fail fast is `set -e` at the top of the script, instead of ` && \ ` at the end of each line.

immerrr: The short answer is that it is not guaranteed to work inside functions. The long answer I have found is here: http://lists.gnu.org/archive/html/bug-bash/2012-12/msg00093.html

I suppose I could drop the functions, however...

immerrr:
>  Could you please also move the Dockerfile to /dockerfiles/splash folder?

done

dockerfiles/splash/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:12.04
+ENV DEBIAN_FRONTEND noninteractive
+
+# Install, use dev tools, and then clean up in one RUN transaction
+# to minimize image size.
+
+# software-properties-common contains "add-apt-repository" command for PPA conf
+# ppa:pi-rho/security is a repo for libre2
```
kmike: this comment is not meaningful here

immerrr: thanks, updated

kmike:
This change broke Docker Hub - it can't build an image. I've updated settings to point to the new Dockerfile, but when Docker Hub tries to build an image an error happens:

> dockerfiles/splash/provision.sh: no such file or directory

I've sent a request to Docker support and now I'm waiting for the reply. We may have to revert this PR if Docker Hub automated builds can't handle it.

immerrr:
Hmm.. I see the errors in the build log [here](https://registry.hub.docker.com/u/scrapinghub/splash/builds_history/16922/), but I can't see the actual errors, I guess they are only available for the project owners.  Can we add a debug-echo to the dockerfile temporarily to see where in the directory structure `docker build` is executed?

kmike:
The answer from the support team was the following:

> The automated build system does the equivalent to:
> cd dockerfiles/splash && docker build -t splash-local .

They don't support `-f` option in DockerHub at the moment. Suggested workaround:

> One workaround that may or may not work might be to split your build into two images. The first image would be triggered when the git repo is updated. It would take care of the ADD . portion. It would have a webhook for a secondary build that would build from dockerfiles/splash/Dockerfile, which would build FROM the original base image.

kmike:
It looks like we're out of luck if we want to have a Dockerfile in a subfolder and ADD something.
splash-jupyter Dockerfile works only because it doesn't use ADD.

So it was my suggestion to move all files in a subfolder which caused the problems.

Issue #191, opened 2015-03-10
Allow to enable/disable JavaScript 


kmike:
See https://github.com/scrapinghub/splash/blob/a18dbac13c8b13afe8c744d928650cbb2ce0b03f/splash/browser_tab.py#L110

This should be exposed as a Lua commands - `splash:enable_js()` and `splash:disable_js()`?

kmike:
+ we should add an argument to the HTTP server.

kmike:
It also worths it to expose other options; it could be a separate issue though.

HotSushi:
Please test this patch for this issue(#191)  
https://gist.github.com/HotSushi/4b5189e4d626f6484717

HotSushi:
Output of following Lua script
```
function main(splash)
    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
    png=splash:png()
    return png
end
```

![Image](https://cloud.githubusercontent.com/assets/6441597/6604868/991325ba-c851-11e4-9d29-06cfea5fc4f0.png)

HotSushi:
After calling splash:disable_js() ..
```
function main(splash)
    splash:disable_js()
    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
    png=splash:png()
    return png
end
```
![imagedisablejs](https://cloud.githubusercontent.com/assets/6441597/6604920/06863cfe-c852-11e4-81f1-55c9f6b9fac2.png)

HotSushi:
The function only works, if it called in the beginning not later!
```
function main(splash)
    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
    splash:disable_js()
    png=splash:png()
    return png
end
```
The output png will say 'Is Javascipt Enabled? YES' 

Any solution?

kmike:
Hey @HotSushi,

Your patch is in a right direction, but it has some issues:

1. the indentation is off (we use 4 spaces consistently, as per PEP8);
2. these methods should be documented like all other methods (see https://github.com/scrapinghub/splash/blob/master/docs/scripting-ref.rst);
3. why is `self.web_view.setPage(self.web_page)` needed?
4. there should be automated tests for this feature (see https://github.com/scrapinghub/splash/blob/master/splash/tests/test_execute.py).

> The output png will say 'Is Javascipt Enabled? YES' 

The JS code which produces YES is executed before the JS was turned off. So this example doesn't show that the function is not working.

[...]

PR #192, opened 2015-03-14
Allow enabling/disabling JavaScript via Lua commands.

sunu:
Fixes issue #191

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -339,31 +339,16 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
-.. _splash-disable_js:
+.. _splash-js-enabled:
 
-splash:disable_js
+splash.js_enabled
 -----------------
 
-Disable execution of JavaSript code embedded in the page.
+Enable or disable execution of JavaSript code embedded in the page.
 
-**Signature:** ``splash:disable_js()``
-
-Should be called before ``splash:go(...)``.
-
-See also: :ref:`splash-enable_js`.
-
-.. _splash-enable_js:
-
-splash:enable_js
-----------------
-
-Enable execution of JavaSript code embedded in the page.
-
-**Signature:** ``splash:enable_js()``
+**Signature:** ``splash.js_enabled = true/false``
 
 JavaScript execution is enabled by default.
-
-See also: :ref:`splash-disable_js`.
 
 .. _splash-runjs:
 
@@ -1205,22 +1190,14 @@ Example:
          ]]
      end
 
-.. _splash-set-images-enabled:
+.. _splash-images-enabled:
 
-splash:set_images_enabled
--------------------------
+splash.images_enabled
+---------------------
 
 Enable/disable images.
 
-**Signature:** ``splash:set_images_enabled(enabled)``
-
-**Parameters:**
-
-* enabled - ``true`` to enable images, ``false`` to disable them.
-
-**Returns:** nil.
-
-**Async:** no.
+**Signature:** ``splash.images_enabled = true/false``
 
 By default, images are enabled. Disabling of the images can save a lot
 of network traffic (usually around ~50%) and make rendering faster.
@@ -1240,7 +1217,7 @@ Example:
 .. code-block:: lua
 
      function main(splash)
-         splash:set_images_enabled(false)
+         splash.images_enabled = false
          assert(splash:go("http://example.com"))
          return {png=splash:png()}
      end
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -339,6 +339,32 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
+.. _splash-disablejs:
+
+splash:disablejs
+-------------
+
+Disable execution of JavaSript code embedded in the page.
+
+**Signature:** ``splash:disablejs()``
+
+Should be called before ``splash:go(...)``.
+
+See also: :ref:`splash-enablejs`.
+
+.. _splash-enablejs:
+
+splash:enablejs
+-------------
+
+Enable execution of JavaSript code embedded in the page.
+
+**Signature:** ``splash:enablejs()``
+
+JavaScript execution is enabled by default.
+
+See also: :ref:`splash-disablejs`.
+
 .. _splash-runjs:
 
 splash:runjs
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -339,31 +339,31 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
-.. _splash-disablejs:
+.. _splash-disable_js:
 
-splash:disablejs
--------------
+splash:disable_js
+-----------------
 
 Disable execution of JavaSript code embedded in the page.
 
-**Signature:** ``splash:disablejs()``
+**Signature:** ``splash:disable_js()``
 
 Should be called before ``splash:go(...)``.
 
-See also: :ref:`splash-enablejs`.
+See also: :ref:`splash-enable_js`.
 
-.. _splash-enablejs:
+.. _splash-enable_js:
 
-splash:enablejs
--------------
+splash:enable_js
+----------------
 
 Enable execution of JavaSript code embedded in the page.
 
-**Signature:** ``splash:enablejs()``
+**Signature:** ``splash:enable_js()``
 
 JavaScript execution is enabled by default.
 
-See also: :ref:`splash-disablejs`.
+See also: :ref:`splash-disable_js`.
 
 .. _splash-runjs:
 
--- a/splash/lua_modules/splash.lua
+++ b/splash/lua_modules/splash.lua
@@ -106,13 +106,16 @@ end
 -- error handling.
 --
 local Splash = {}
-Splash.__index = Splash
 
 local private = {}
 local PRIVATE_PREFIX = "private_"
 
 function Splash._create(py_splash)
-  local self = {args=py_splash.args}
+  local self = {
+    args=py_splash.args,
+    _setters={},
+    _getters={}
+  }
   setmetatable(self, Splash)
 
   -- Create Lua splash:<...> methods from Python Splash object:
@@ -145,6 +148,14 @@ function Splash._create(py_splash)
     end
   end
 
+  -- Set attribute handler functions
+  for key, opts in pairs(py_splash.lua_properties) do
+    local func = drops_self_argument(py_splash[key])
+    self._getters[opts.name] = drops_self_argument(
+                                            py_splash[opts.getter_method])
+    self._setters[opts.name] = func
+  end
+
   return self
 end
 
@@ -166,12 +177,12 @@ Request.__index = Request
 function Request._create(py_request)
   local self = {info=py_request.info}
   setmetatable(self, Request)
-  
+
   -- copy informational fields to the top level
   for key, value in pairs(py_request.info) do
     self[key] = value
   end
-  
+
   for key, opts in pairs(py_request.commands) do
     local command = py_request[key]
     command = drops_self_argument(command)
@@ -182,7 +193,7 @@ function Request._create(py_request)
 
     self[key] = command
   end
-  
+
   return self
 end
 
@@ -193,5 +204,20 @@ function Splash:on_request(cb)
   end)
 end
 
+function Splash:__newindex( index, value )
+  if self._setters[index] then
+    self._setters[index](self, value)
+  else
+    rawset(Splash, index, value)
+  end
+end
+
+function Splash:__index(index)
+  if self._getters[index] then
+    return self._getters[index](self)
+  else
+    return rawget(Splash, index)
+  end
+end
 
 return Splash
--- a/splash/qtrender_lua.py
+++ b/splash/qtrender_lua.py
@@ -69,6 +69,19 @@ def command(async=False, can_raise_async=False, table_argument=False,
     return decorator
 
 
+def lua_property(name):
+    """ Decorator for marking methods that make attributes available to Lua """
+    def decorator(meth):
+        def setter(method):
+            meth._setter_method = method.__name__
+            return method
+        meth._is_lua_property = True
+        meth._name = name
+        meth.lua_setter = setter
+        return meth
+    return decorator
+
+
 def emits_lua_objects(meth):
     """
     This decorator makes method convert results to
@@ -103,6 +116,11 @@ def first_argument_from_storage(meth):
 def is_command(meth):
     """ Return True if method is an exposed Lua command """
     return getattr(meth, '_is_command', False)
+
+
+def is_lua_property(meth):
+    """ Return True if method is exposed to an Lua attribute """
+    return getattr(meth, '_is_lua_property', False)
 
 
 def can_raise(meth):
@@ -164,6 +182,24 @@ def get_commands(obj):
     return commands
 
 
+def get_lua_properties(obj):
+    """
+    Inspect a Python object and get a dictionary of all lua properties, their
+    getter and setter methods which were made available to Lua using
+    @lua_property and @<getter_method_name>.lua_setter decorators.
+    """
+    lua_properties = {}
+    for name in dir(obj):
+        value = getattr(obj, name)
+        if is_lua_property(value):
+            setter_method = getattr(value, '_setter_method')
+            lua_properties[setter_method] = {
+                'name': getattr(value, '_name'),
+                'getter_method': name,
+            }
+    return lua_properties
+
+
 class _WrappedJavascriptFunction(object):
     """
     JavaScript functions wrapper. It allows to call JS functions
@@ -222,7 +258,8 @@ class Splash(object):
     (wrapped in 'Splash' Lua object; see :file:`splash/lua_modules/splash.lua`).
     """
     _result_content_type = None
-    _attribute_whitelist = ['commands', 'args', 'tmp_storage']
+    _attribute_whitelist = ['commands', 'args', 'tmp_storage',
+                            'lua_properties']
 
     def __init__(self, lua, tab, render_options=None):
         """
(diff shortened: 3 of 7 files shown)


docs/scripting-ref.rst
```diff
@@ -331,6 +331,32 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
+.. _splash-disablejs:
+
+splash:disablejs
```
immerrr: Do we need separate `disable`/`enable` funcs? There's already a `splash:set_images_enabled`, so `set_js_enabled(true/false)` seems more natural.

immerrr: Oh, I see they are mentioned like that in #191, @kmike do you think it's better to have them as separate functions?

sunu: @immerrr Yeah, I agree that having separate `disable` and  `enable` functions is a bit redundant. Should I update the pr or wait for @kmike's response?

kmike: That's a good question. We have 3 options:

1. splash:enablejs() / splash:disablejs()
2. splash:set_js_enabled(...)
3. splash.js_enabled = ...

I proposed (1) because it looks nicer in user scripts than (2) - splash:set_js_enabled(false) is longer and less readable.

There is indeed splash:set_images_enabled already, and it is better to be consistent though.

(3) also provides a nice API (even nicer). It is shorter and allows to read the value as well as write to it. A problem with (3) is that it is a bit harder to implement (sandbox code should be changed), and that it is harder to indicate errors or make action async this way. There shouldn't be any errors, and the operation is not async, so (3) looks good to me.

kmike: I like (3) most, next comes (1), the worst is (2).

kmike: The docs are incorrect, by the way: methods are named enable_js/disable_js in the code, but enablejs/disablejs in the docs. Also, underline is too short.

docs/scripting-ref.rst
```diff
@@ -331,6 +331,32 @@ Compare:
...
+-------------
+
+Disable execution of JavaSript code embedded in the page.
+
+**Signature:** ``splash:disablejs()``
+
+Should be called before ``splash:go(...)``.
```
kmike: Why? What happens if it is called after splash:go()? It'd be good to test that.

splash/tests/test_execute.py
```diff
@@ -332,6 +332,47 @@ def test_pcall_wrong_keyword_arguments(self):
         self.assertEqual(data["ok"], False)
 
 
+class EnableDisableJSTest(BaseLuaRenderTest):
```
kmike: good tests :+1:

splash/tests/test_execute.py
```diff
@@ -332,6 +332,70 @@ def test_pcall_wrong_keyword_arguments(self):
...
+            return html
+        end
+        """, {
+            'url': self.mockurl('jsrender'),
+        })
+        self.assertStatusCode(resp, 200)
+        self.assertIn(u'After', resp.text)
+
+    def test_disablejs_after_splash_go(self):
+        # disable_js doesn't work when called after splash.go(...)
```
kmike: The test doesn't show that. HTML may have been changed while JS was still enabled, but splash:disable_js() might still disable further JS processing.

sunu:
@kmike Fixed the docs and tests for case (3). I tried implementing (1) with an attribute with a setter in Splash class, but couldn't get it to work. Any pointers on how to implement that?

kmike:
@sunu Splash creates a Lua [runtime](https://github.com/scrapinghub/splash/blob/master/splash/lua_runtime.py) for each script. Then it finds "main" Lua function in the script. This "main" function receives "splash" object. To build this object we first create and instance of "[splash.qtrender_lua.Splash](https://github.com/scrapinghub/splash/blob/a18dbac13c8b13afe8c744d928650cbb2ce0b03f/splash/qtrender_lua.py#L171)". We don't pass `splash.qtrender_lua.Splash` object directly to Lua script for several reasons: it is not good to expose internals to the script, and by wrapping it in another object we can provide a better API. So this object is wrapped in a Lua object - see [here](https://github.com/scrapinghub/splash/blob/a18dbac13c8b13afe8c744d928650cbb2ce0b03f/splash/lua_modules/splash.lua#L90). The wrapper exposes some methods and attributes of the original Python object, but not all of them, and with a slightly different API.

Currently access to Python-level objects is limited, based on a whitelist. This is implemented [here](https://github.com/scrapinghub/splash/blob/a18dbac13c8b13afe8c744d928650cbb2ce0b03f/splash/lua_runtime.py#L105) using [lupa](https://github.com/scoder/lupa#restricting-lua-access-to-python-objects) access restriction feature. Lua code can only call some whitelisted methods of a single Python object (`splash.qtrender_lua.Splash`), and read a single whitelisted attribute of this object.

So in order to provide `.js_enabled` property you need to 

1. extend the sandbox to allow to write to selected properties of selected objects;
2. add `.js_enabled` property to "Splash" object (with getter and setter);
3. expose this property to Lua script, similar to how is splash.args property exposed, but with write access enabled.

kmike:
@immerrr are you OK with .js_enabled property instead of `.set_js_enabled(true/false)`?

immerrr:
Sure. It looks nicer.

splash/tests/test_execute.py
sunu: OK. The script is processed asynchronously. Got it.

sunu:
@kmike Thank you very much for the pointers. I have made some changes but I still can't get it working. I'll update the PR with my unfinished changes now. Hopefully you can show me the right way forward.

splash/lua_modules/splash.lua
```diff
@@ -98,7 +98,7 @@ local Splash = {}
 Splash.__index = Splash
 
 function Splash.private_create(py_splash)
-  local self = {args=py_splash.args}
+  local self = {args=py_splash.args, js_enabled=py_splash.js_enabled}
```
kmike: Here you're creating spalsh.js_enabled variable by copying py_splash.js_enabled value to it. So changes to splash.js_enabled won't affect py_splash.js_enabled. `args` is different because it is a dict.

splash/tests/test_execute.py
```diff
@@ -332,6 +332,67 @@ def test_pcall_wrong_keyword_arguments(self):
...
+            splash.js_enabled = true
+            splash:go(splash.args.url)
+            local html = splash:html()
+            return html
+        end
+        """, {
+            'url': self.mockurl('jsrender'),
+        })
+        self.assertStatusCode(resp, 200)
+        self.assertIn(u'After', resp.text)
```
HotSushi: @sunu , I think you should **assertNotIn(u'Before', resp.text)** rather than **assertIn(u'After', resp.text)**, because 'After' will be present in the html's script tag anyways.

kmike:
> What's the proper way to debug the lua script execution process btw? I tried using pdb but couldn't find anything useful. 

I use prints :)

splash/tests/test_execute.py
sunu: Thanks for the catch. Fixed 'em now.

sunu:
Hey @kmike,
I think I have implemented the expected behavior now and got the tests to pass. But I'm not sure how clean my way of implementation is. Please take a look and let me know.

sunu:
Ah, didn't run the full test suite. Sorry.

sunu:
Fixed now. :)

kmike:
Hey @sunu,

I like your idea of creating properties in a Lua wrapper - it makes the sandboxing code simpler.

splash/lua_modules/splash.lua
```diff
@@ -98,7 +98,7 @@ local Splash = {}
 Splash.__index = Splash
 
 function Splash.private_create(py_splash)
-  local self = {args=py_splash.args}
+  local self = {args=py_splash.args, members={}, set_js_enabled=drops_self_argument(py_splash['set_js_enabled'])}
```
kmike: Could you please split this into multiple lines:

```
local self = {
    args=py_splash.args,
    members={},
    ... 
}
```

The line is too long :) there is no pep8 for Lua code, but it is better to be consistent when possible.

splash/lua_modules/splash.lua
```diff
@@ -132,5 +132,21 @@ function Splash:jsfunc(...)
...
+function Splash:__newindex( index, value )
+  if index == "js_enabled" then
+    self.members[index] = value
+    self:set_js_enabled(value)
+  else
+    rawset(Splash, index, value)
+  end
+end
+
+function Splash:__index(index)
```
kmike: `Splash.__index` was assigned to another value above. It is better to delete that assignment if you're changing what's `Splash.__index`.

splash/qtrender_lua.py
```diff
@@ -174,7 +174,7 @@ class Splash(object):
     (wrapped in 'Splash' Lua object; see :file:`splash/lua_modules/splash.lua`).
     """
     _result_content_type = None
-    _attribute_whitelist = ['commands', 'args']
+    _attribute_whitelist = ['commands', 'args', 'set_js_enabled']
```
kmike: Why not use `@command`? I don't like putting `set_js_enabled` to `_attribute_whitelist` because all other functions are declared as commands.

splash/lua_modules/splash.lua
```diff
@@ -132,5 +132,21 @@ function Splash:jsfunc(...)
   return unwraps_errors(func)
 end
 
+function Splash:__newindex( index, value )
+  if index == "js_enabled" then
+    self.members[index] = value
+    self:set_js_enabled(value)
+  else
```
kmike: It'd be nice to add `splash.images_enabled` as well, for consistency, and maybe also `splash.cookies`, `splash.result_content_type`, `splash.navigation_locked`, etc. But it is not easy to add them right now - names of these attributes have to be listed in several places, etc. I'm not asking to add them all in this PR, but the code should allow that. There should be an obvious way to add an attribute (like we can add a new method by creating a Python method and marking it as `@command`).

sunu:
@kmike updated. Let me know what you think.
I have only changed set_images_enabled for now. I can do the same for the other methods in a separate PR.

kmike:
@sunu this looks better, but if I'm not mistaken this PR currently doesn't allow to read attribute values.

splash/qtrender_lua.py
```diff
@@ -209,6 +209,15 @@ def __init__(self, lua, tab, render_options=None):
                     'can_raise_async': getattr(value, '_can_raise_async', False),
                 })
         self.commands = self.lua.python2lua(commands)
+
+        lua_attr_to_functions = {
+            'js_enabled': 'set_js_enabled',
+            'images_enabled': 'set_images_enabled'
```
kmike: Could you please make it a decorator similar to `@command`?

sunu: Ok. I will.

sunu:
Yeah, the attribute values can be read only after they are set first. So the default values can't be read. I'll try to fix that.

sunu:
Hey @kmike! Does this look ok now?

kmike:
Hi @sunu,

It seems something went wrong, and there are now extra commits in the PR. Could you please rebase it on master? I'll be back to reviewing after that.

sunu:
Hey @kmike !
I rebased it on master and tweaked it a bit so that the `set_images_enabled` change is now backwards-compatible. Please review :)

splash/qtrender_lua.py
```diff
@@ -221,6 +248,10 @@ def init_dispatcher(self, return_func):
         """
         self._return = return_func
 
+    @attr_function(attr_name='js_enabled', default=True)
```
kmike: Hi @sunu,

I think having default values this way is not good; it is better for property to be able to read an actual property value. This would be less error prone. Currently if we e.g. add a command-line option to use js_enabled=False by default, default value here will have to be changed, but it is not clear how. Also, other commands may change these properties; in this case the read value will be incorrect.

splash/tests/lua_modules/emulation.lua
```diff
@@ -26,7 +26,7 @@ function Splash:go_and_wait(args)
     error("non-zero 'wait' is required when rendering whole page")
   end
 
-  self:set_images_enabled(self.args.images)
+  self.images_enabled = self.args.images
```
kmike: This was the only place set_images_enabled function is tested; if you're removing it another testshould be aded to ensure the function remains working.

docs/scripting-ref.rst
```diff
@@ -339,6 +339,32 @@ Compare:
...
+----------------
+
+Enable execution of JavaSript code embedded in the page.
+
+**Signature:** ``splash:enable_js()``
+
+JavaScript execution is enabled by default.
+
+See also: :ref:`splash-disable_js`.
+
```
kmike: the docs are outdated.

splash/qtrender_lua.py
sunu: Yes, I agree. This kind of felt like the wrong way to specify default value to me too, but I couldn't think of a proper alternative.

Would it be better if I set these attributes and then explicitly set the default values in `__init__()`
like this?
``` python   
def __init__(....):
    self.images_enabled = True   # Specify the default value
    ....
    self.set_images_enabled(self.images_enabled)  # Set default value explicitly
```

kmike: It should be a Python property, or do something similar to what are Python properties doing, but in Lua. 

So instead of a single set_js_enabled method you can have one more method, get_js_enabled, call it from Lua and expose the result as an attribute. Instead of these two methods we can also have a Python property, but it'll require some changes to the sandboxing code.

sunu:
@kmike PR updated :)

docs/scripting-ref.rst
```diff
@@ -339,6 +339,17 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
+.. _splash-js_enabled:
```
kmike: the convention is to use "-" in reference names: _splash-js-enabled

docs/scripting-ref.rst
```diff
@@ -1181,20 +1192,12 @@ Example:
 
 .. _splash-set-images-enabled:
```
kmike: reference should be renamed

splash/qtrender_lua.py
```diff
@@ -54,6 +54,16 @@ def decorator(meth):
...
 
 
+def attr_function(attr_name, gettr_method):
+    """ Decorator for marking methods that make attributes available to Lua """
+    def decorator(meth):
+        meth._is_attr_func = True
+        meth._attr_name = attr_name
+        meth._gettr_method = gettr_method
+        return meth
+    return decorator
```
kmike: What about renaming this decorator to `@lua_property`? It is similar to `@property` decorator, but creates a property available to Lua code.

kmike: attributes like _is_attr_func should be also renamed if we rename the decorator

splash/qtrender_lua.py
```diff
@@ -221,6 +250,13 @@ def init_dispatcher(self, return_func):
         """
         self._return = return_func
 
+    @attr_function(attr_name='js_enabled', gettr_method='get_js_enabled')
```
kmike: I think https://docs.python.org/2/library/functions.html#property syntax is better - instead of passing method name it is better to expose `@<meth_name>.setter` decorator.

sunu:
@kmike Updated PR. I have left `set_images_enabled` there for backward compatibility because I couldn't find a generic way to do it with the @property syntax.

splash/qtrender_lua.py
```diff
@@ -221,6 +238,15 @@ def init_dispatcher(self, return_func):
         """
         self._return = return_func
 
+    @property
+    def js_enabled(self):
+        return self.tab.get_js_enabled()
```
kmike: Hi @sunu! I liked your previous implementation more :) My idea was not to use built-in Python `@property` decorator, but to provide another syntax for specifying the setter. The syntax should look like this:

```
@lua_property(name='js_enabled')
def get_js_enabled(self):
    # ...

@get_js_enabled.setter
def set_js_enabled(self, value):
    # ...
```

immerrr: big :+1 on this

sunu:
@kmike Sorry for the confusion. Fixed the syntax now.

kmike:
Hi @sunu - it starts to look good :+1: 
.. but I broke your PR with recent changes. Could you please 

* rebase it to make mergeable;
* clean up the code in `Splash.__init__` - see how it is done [now](https://github.com/scrapinghub/splash/blob/3ddd263fb0c5e85e4d421b05e2d4732181cc583f/splash/qtrender_lua.py#L149) for commands;
* use `@func.setter` syntax instead of `@func.setter()` - parentheses are not needed, and builtin `@property` decorator doesn't use them;
* change `@lua_property(attr_name='foo')` to `@lua_property(name='foo')` or just `@lua_property('foo')`. The problem with `attr_name` name is that we'll use two words to describe the same: "attr" and "property", and it is confusing a reader may think they mean something different. `@lua_property(property_name='foo')` is redundant => `@lua_property(name='foo')`.

sunu:
Hey @kmike,
Updated the PR. Take a look.

kmike:
Hey @sunu,

Could you please check the test failures? https://travis-ci.org/scrapinghub/splash/jobs/63351491 - it seems that a raw unwrapped Python object (a method) is available to user Lua code. We shouldn't expose Python methods to scripts directly.

sunu:
@kmike fixed the failure.

splash/lua_modules/splash.lua
```diff
@@ -106,13 +106,13 @@ end
...
 local Splash = {}
-Splash.__index = Splash
 
 local private = {}
 local PRIVATE_PREFIX = "private_"
 
 function Splash._create(py_splash)
-  local self = {args=py_splash.args}
+  local self = {args=py_splash.args, attr_to_settr_map={},
+                attr_to_gettr_map={}}
```
kmike: What about making attr_to_getter_map and attr_to_setter_map private? Currently they will be suggested by the autocompletion engine if you use Splash-Jupyter, and they shouldn't. Adding `_` to the name may be enough to hide them from autocompleter - see [here](https://github.com/scrapinghub/splash/blob/fe80d031bf4d45e6f8b341f643a60018607fef9c/splash/kernel/completer.py#L94). This won't hide them from user, but I don't think it is a problem. An alternative is to put these maps outside Splash object (like it is done for `private`), or to protect against accessing them in `__newindex` / `__index`. The way private attributes are hidden is not perfect because it requires at most 1 Splash object per Lua runtime; we'll have to refactor this code if we start supporting multiple browser tabs in a single Lua script.

We should also rename these methods: `settr` vs `setter` saves only a single character :) Just `_setters` and `_getters` sounds better to me, but if you prefer longer names that's fine.

Could you please also indent it like dicts are indented in Splash Python code?

```lua
local self = {
  args=py_splash.args, 
  _setters={},
  _getters={},
}
```

sunu: Yes, makes sense. Will update the code.

sunu:
@kmike PR updated. I've set the `set_images_enabled` method only as `@command` because it's the only one we need to be exposed as a method. Let me know if I missed anything.

kmike:
Good work @sunu!

PR #193, opened 2015-03-15
Allow to enable/disable JavaScript  #191

HotSushi:
- browser_tab supports enable/disable Javascript. (`enable_js()`,` disable_js()`)
- qtrender_lua can invoke enable_js,disable_js
- documentation
- tests written (successful)

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -331,6 +331,44 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
+.. _splash-enable_js:
+
+splash:enable_js
+------------
+
+Enables javascript for the browser.
+
+**Signature:** ``splash:enable_js()``
+
+Example:
+
+.. code-block:: lua
+
+    splash:enable_js()
+    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
+
+See also: :ref:`splash-disable_js`, :ref:`splash-runjs`, :ref:`splash-jsfunc`,
+:ref:`splash-evaljs`.
+
+.. _splash-enable_js:
+
+splash:disable_js
+------------
+
+Disables javascript for the browser.
+
+**Signature:** ``splash:disable_js()``
+
+Example:
+
+.. code-block:: lua
+
+    splash:disable_js()
+    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
+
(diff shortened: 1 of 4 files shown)


docs/scripting-ref.rst
```diff
@@ -331,6 +331,44 @@ Compare:
 See also: :ref:`splash-runjs`, :ref:`splash-jsfunc`,
 :ref:`splash-wait-for-resume`, :ref:`splash-autoload`.
 
+.. _splash-enable_js:
+
+splash:enable_js
+------------
```
kmike: the underline is too short

docs/scripting-ref.rst
```diff
@@ -331,6 +331,44 @@ Compare:
...
+Enables javascript for the browser.
+
+**Signature:** ``splash:enable_js()``
+
+Example:
+
+.. code-block:: lua
+
+    splash:enable_js()
+    assert(splash:go('https://www.whatismybrowser.com/is-javascript-enabled'))
```
kmike: This example looks redundant because JS is enabled by default

splash/browser_tab.py
```diff
@@ -548,6 +548,16 @@ def runjs(self, js_source, handle_errors=True):
         js_source = "%s;undefined" % js_source
         self.evaljs(js_source, handle_errors=handle_errors)
 
+    def enable_js(self):
+        """ Enables Javascript """
```
kmike: from PEP257:

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

splash/tests/test_execute.py
```diff
@@ -2017,6 +2017,47 @@ def test_related_resources(self):
...
+                    </body></html>
+                ]],
+                baseurl = splash.args.base,
+            }
+            return splash:png()
+        end
+        """
+        resp = self.request_lua(script, {"base": self.mockurl("")})
+        self.assertStatusCode(resp, 200)
+        img = Image.open(StringIO(resp.content))
```
kmike: I think this way to check if JS is enabled is too indirect; it is better to check just the resulting HTML than to render an image and encode the information in pixel colors.

kmike:
See also discussion in https://github.com/scrapinghub/splash/pull/192 - we should check what happens when splash:disable_js or splash:enable_js is called after page is loaded.

HotSushi:
@kmike, I will close this pull request as sunu is already working on it, thanks for your comments I will keep them in mind during my next commit..

Issue #180, opened 2015-02-06
improve argument validation


kmike:
When 'wait=foobar' is passed, Splash returns a stacktrace. It makes sense to return some validation error. The issue could be not specific to 'wait' argument.

kmike:
Fixed by https://github.com/scrapinghub/splash/commit/470dffd39af0edcc2d852bba04412c2b9685020a.

PR #198, opened 2015-03-27
improved argument validaton

jp111:
Please review Issue #180


kmike:
Hi @jp111,

Could you please add a test for it?

jp111:
Hi @kmike ,
I have placed the test under Base class so that it runs for all children classes.

kmike:
Thanks @jp111! 
I've cherry-picked your changes because it seems something is not right with the PR.

Issue #186, opened 2015-03-02
tests are unstable


kmike:
Some tests (notably, redirect tests which rely on timeouts, and some else) often fail. 
I wonder if we can use something like https://github.com/klrmn/pytest-rerunfailures for them (the repo is unsupported though), or just mark them using some pytest marker and skip by default.

PR #222, opened 2015-05-27
[Work In Progress] Use PyQt5 instead of PyQt4.

sunu:


sunu:
Hmm. Need to figure out how to install PyQt5 on Travis.

sunu:
@kmike Hey, I have run into a problem here. Installing qt5.4 on Travis seems to be quite difficult. [The ppa that Quamash is using](https://github.com/harvimt/quamash/blob/master/.travis.yml#L24) seems to be offline. And there doesn't seem to be any qt5.4 ppa for Ubuntu 12.04. Travis times out if I try to compile things from source even if I try to compile only the required modules. There is a ppa for Qt5.1 though. But we would have to use a lower version of PyQt5 as well because PyQt5.4 (the Webkit modules iirc) doesn't compile with it.

Also Travis [doesn't seem](https://github.com/travis-ci/travis-ci/issues/2046)  to support 14.04 yet.

So I looked for alternative CI services. [Shippable](https://www.shippable.com/) seems nice. How do you feel about moving there? They support docker images. So we can use 14.04. And they seem to parse travis.yml out of the box too. 

After trying out for some time, I have the tests running on Ubuntu 14.04 there. Here's the link to the [current build](https://app.shippable.com/builds/556b075e475bca0b0007f07f). 7 tests are failing for now.

kmike:
Hey @sunu! The updated link to the PPA Quamash is using is https://launchpad.net/~beineri/+archive/ubuntu/opt-qt541. I'd prefer to keep using Travis CI if possible.

In http://pyqt.sourceforge.net/Docs/PyQt5/introduction.html#an-explanation-of-version-numbers it is said that one should always use latest PyQT version, and it will still support earlier Qt versions. Did you find it to be not true?

I think it'd be helpful to support at least qt 5.2, because qt 5.2 is what you find in latest Ubuntu LTS (14.04), and it'd be good to allow Splash to run on stock software as much as possible.

There are PPAs for qt5.3 and qt5.2 in beineri's repo: see e.g. https://launchpad.net/~beineri/+archive/ubuntu/opt-qt532 and https://launchpad.net/~beineri/+archive/ubuntu/opt-qt521, so it should be possible to test Splash against multiple QT versions automatically if we find out that the differences between them are significant and regressions are common.

.travis.yml
```diff
@@ -9,15 +9,51 @@ branches:
 
 env:
     matrix:
-        - SYSTEM_PACKAGES=false
-        - SYSTEM_PACKAGES=true
+        - SYSTEM_PACKAGES=false QT=qt5.2
+        - SYSTEM_PACKAGES=true  QT=qt5.2
+        - SYSTEM_PACKAGES=false QT=qt5.4
+        - SYSTEM_PACKAGES=true  QT=qt5.4
```
kmike: Let's use 3 builds instead of 4; I think there is no need to test SYSTEM_PACKAGES=false with two different Qt versions. It should be fine to test it only with qt5.2 or qt5.4.

sunu:
There seem to be random failures with SYSTEM_PACKAGES=true sometimes.

kmike:
There are known random failures related to timeouts - see https://github.com/scrapinghub/splash/issues/186; these failures were not related to SYSTEM_PACKAGES=true/false. I've tried to make tests more stable with https://github.com/box/flaky - it worked, but I haven't decided if allowing tests to fail sometimes is a right thing to do. For now, in case of random test failures please rerun tests - for me it usually worked on second pass.

splash/tests/test_execute.py
```diff
@@ -559,22 +559,14 @@ def test_dateobj(self):
...
     def test_regexp(self):
         self.assertEvaljsResult(
-            '/my-regexp/i',
-            {
-                u'_jstype': u'RegExp',
-                'caseSensitive': False,
-                'pattern': u'my-regexp'
-            },
+            'var a = "123A"; a.match(/[a]/i);',
+            ['A'],
```
kmike: This is a backwards-incompatible change; it requires changes to docs. It is fine to remove this feature (it is not that useful), but could you please check why does it work differently with pyqt5/qt5?

splash/qtutils.py
```diff
@@ -175,8 +175,8 @@ def qt2py(obj, max_depth=100):
 
     # print(obj, obj.__class__)
 
-    if isinstance(obj, QString):
-        return unicode(obj)
+    # if isinstance(obj, QString):
+    #     return unicode(obj)
```
kmike: why is it needed?

sunu:
@kmike This is ready for review.

kmike:
I've merged your changes to https://github.com/scrapinghub/splash/tree/qt5 branch.

Issue #106, opened 2014-07-22
Filter responses based on content-type


kmike:
Idea was proposed by Pablo in https://github.com/scrapinghub/splash/issues/68#issuecomment-32474061.

kmike:
Fixed by #225 and #246.

PR #225, opened 2015-05-31
Allow aborting requests depending on the content-type

Youwotma:
Issue: #106 

This can be used to block images, videos, css and other unnecessary resources from loading.


kmike:
Hi @YouWoTMA,

+1 to add this feature, but I think Splash should allow to set these filters per-request, not at startup.

YouWoTMA:
Yes, I agree. I'll update the pull request in the following days :)

kmike:
I think the most general way to do it is to implement #231. But passing these filters in application/json POST requests also sounds good to me.

YouWoTMA:
I have changed it to a per-request option, I'm not familiar with lua :)

splash/request_middleware.py
```diff
@@ -72,6 +72,59 @@ def process(self, request, render_options, operation, data):
...
+        """
+        Split a mime string into type and subtype: 'text/html; charset=utf-8' -> ('text', 'html')
+        """
+        separator = mime.find(';')
+        if separator > 0:
+            mime = mime[:separator].strip()
+        parts = mime.split("/")
+        if len(parts) != 2:
+            return None
+        return parts[0], parts[1]
```
kmike: Could you please make these static methods regular functions?
It'd be also nice to add doctests for them. For example, you can convert your existing example to a doctest which will run automatically:

```
Split a mime string into type and subtype

>>> split_mime('text/html; charset=utf-8')
('text', 'html')
```

splash/request_middleware.py
```diff
@@ -72,6 +72,59 @@ def process(self, request, render_options, operation, data):
         return request
 
 
+class ContentTypeMiddleware(object):
```
kmike: it is odd to have a response middleware in a file named request_middleware. Let's move it to response_middleware.py

splash/request_middleware.py
```diff
@@ -72,6 +72,59 @@ def process(self, request, render_options, operation, data):
...
+    def __init__(self, verbosity=0):
+        self.verbosity = verbosity
+
+    @staticmethod
+    def contains(mime_set, mime):
+        for part1 in mime[0], '*':
+            for part2 in mime[1], '*':
+                if (part1, part2) in mime_set:
+                    return True
+        return False
```
kmike: While globs are nice, I'm a bit worried that we have so many different matching languages in use: regexes for proxy profiles, AdBlockPlus filters for request filters, and now globs for Content-Type filters. What about using regexes?

If we go with globs, it is better to use https://docs.python.org/2/library/fnmatch.html.

splash/network_manager.py
```diff
@@ -363,9 +367,23 @@ def __init__(self, filters_path, allowed_schemes, verbosity):
...
 
+        self.response_middlewares.append(ContentTypeMiddleware(self.verbosity))
+
+    def run_response_middlewares(self):
+        reply = self.sender()
+        render_options = self._getRenderOptions(reply.request())
+        if render_options:
+            for middleware in self.response_middlewares:
+                middleware.process(reply, render_options)
+        reply.metaDataChanged.disconnect(self.run_response_middlewares)
```
kmike: It could be better to disconnect the callback in a 'finally' block. 
By the way, why is disconnecting needed?

kmike:
The implementation looks good to me. I've left some comments, but they are minor.
To merge it we need tests and docs :)

splash/network_manager.py
YouWoTMA: Because it metaDataChanged can be fired multiple times if the response has trailing headers, we probably don't want to run the middlewares multiple times or abort the download once it's completed.

kmike: Never heard about these headers, thanks for the insight.

splash/request_middleware.py
YouWoTMA: I tried to copy the http [Accept header](http://tools.ietf.org/html/rfc7231#section-5.3.2) syntax, but I have changed it to fnmatch because it seems to me like it handles the accept header syntax the same but it's more powerful.

kmike: Aha, this makes sense.

splash/request_middleware.py
kmike: Actually I'm fine with keeping static methods, but they need tests :)

kmike:
Thanks @Youwotma!

Issue #188, opened 2015-03-08
Allow to set result HTTP header from a Splash script


kmike:

pawelmhm:
are we talking about setting headers of response from target website in splash? Can this be implemented by adding :on_response method to splash, similar to :on_request method that we have? Or is this issue about setting HTTP header of response returned by splash?

kmike:
> Or is this issue about setting HTTP header of response returned by splash to splash client?

Yes, this issue is about setting HTTP header of response returned by splash to splash client.

pawelmhm:
ok @kmike I upvote both senses :)

kmike:
Second sense is fixed by #227; for the first sense there is now #230.

PR #227, opened 2015-06-01
[Issue #188][lua_scripting] allow to set result header

pawelmhm:
Adds method to lua splash objects allowing for setting response headers. 

https://github.com/scrapinghub/splash/issues/188

### DIFF
--- a/splash/qtrender_lua.py
+++ b/splash/qtrender_lua.py
@@ -260,7 +260,6 @@ class Splash(object):
     _result_content_type = None
     _attribute_whitelist = ['commands', 'args', 'tmp_storage',
                             'lua_properties']
-    _result_headers = []
 
     def __init__(self, lua, tab, render_options=None):
         """
@@ -301,6 +300,8 @@ class Splash(object):
 
         wrapper = self.lua.eval("require('splash')")
         self._wrapped = wrapper._create(self)
+
+        self._result_headers = []
 
     def init_dispatcher(self, return_func):
         """
@@ -555,6 +556,12 @@ class Splash(object):
     def set_result_header(self, name, value):
         if not all([isinstance(h, basestring) for h in [name, value]]):
             raise ScriptError("splash:set_result_header() arguments must be strings")
+
+        try:
+            name = name.decode('utf-8').encode('ascii')
+            value = value.decode('utf-8').encode('ascii')
+        except UnicodeEncodeError:
+            raise ScriptError("splash:set_result_header() arguments must be ascii")
 
         header = (name, value)
         self._result_headers.append(header)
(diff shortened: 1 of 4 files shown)


splash/tests/test_execute.py
```diff
@@ -187,6 +187,28 @@ def test_bad_content_type_func(self):
...
+        self.assertEqual(resp.headers.get("foo"), "bar")
+
+    def test_bad_result_header_set(self):
+        resp = self.request_lua("""
+        function main(splash)
+            splash:set_result_header({}, [])
+            return "hi!"
+        end
+        """)
+        self.assertStatusCode(resp, 400)
```
kmike: it is better to use self.assertErrorLineNumber

splash/resources.py
```diff
@@ -105,8 +105,10 @@ def _writeOutput(self, data, request, content_type=None):
             data = json.dumps(data, cls=SplashJSONEncoder)
             return self._writeOutput(data, request, "application/json")
 
-        if isinstance(data, tuple) and len(data) == 2:
-            data, content_type = data
+        if isinstance(data, tuple) and len(data) == 3:
+            data, content_type, headers = data
+            for header in headers:
+                request.setHeader(str(header[0]), str(header[1]))
```
kmike: What happens if non-ascii header values are passed? `str(header[0])` will encode them to bytes using the encoding which depends on user locale, etc. I think it is better to be more explicit and use either ascii or latin1 (I prefer ascii slightly more), and raise an error if a value can't be encoded.

splash/resources.py
```diff
@@ -105,8 +105,10 @@ def _writeOutput(self, data, request, content_type=None):
             data = json.dumps(data, cls=SplashJSONEncoder)
             return self._writeOutput(data, request, "application/json")
 
-        if isinstance(data, tuple) and len(data) == 2:
-            data, content_type = data
+        if isinstance(data, tuple) and len(data) == 3:
+            data, content_type, headers = data
+            for header in headers:
```
kmike: `for name, value in headers` is easier to read

splash/qtrender_lua.py
```diff
@@ -260,6 +260,7 @@ class Splash(object):
     _result_content_type = None
     _attribute_whitelist = ['commands', 'args', 'tmp_storage',
                             'lua_properties']
+    _result_headers = []
```
kmike: It should be set in `__init__` - otherwise you'll be updating a class variable shared by all Splash instances.

kmike:
Thanks @pawelmhm! Could you please also add some docs [here](https://github.com/scrapinghub/splash/blob/master/docs/scripting-ref.rst#splashset_result_content_type), near splash:set_result_content_type? It'd be good to add "See also: splash:set_result_content_type" note to set_result_header, and vice versa.

splash/tests/test_execute.py
kmike: `splash:set_result_header({}, [])` is not valid Lua - there is no `[]`, so you'll get 400 error because of a syntax error, not because of validation error.

kmike:
:+1:

PR #228, opened 2015-06-01
Fixed improper detection of HTTP errors

kmike:
See https://github.com/scrapinghub/splash/issues/161#issuecomment-107433312


kmike:
HTTP errors were not detcted correctly for 4xx or 5xx responses with empty bodies.

Issue #231, opened 2015-06-02
provide splash:on_response_headers method


kmike:
It should allow to register a callback called when response headers are received. This callback should allow to filter out responses based on their headers, URL and status code.

See also: https://github.com/scrapinghub/splash/pull/225, https://github.com/scrapinghub/splash/issues/106. 

This method should be called before splash:on_response proposed in #230; its purpose is to allow to filter out responses without having to download their bodies.

pawelmhm:
I started impleneting this but I found out that there are obstacles if you want to actually modify QNetworkReply from python code, if you try to do something like

```python
reply = self.sender() # returns QNetworkReply object
reply.setRawHeader("foo", "bar")
```
you will get
```
RuntimeError('no access to protected functions or signals for objects not created from Python',)
```
this seems to mean that c++ objects such as QNetworkReply cannot be modified from Python, I looked into docs and setter methods like [setHeader or setUrl are protected](http://doc.qt.io/qt-5/qnetworkreply.html), public methods of reply are for reading only. 

I'm not exactly sure how we can bypass this, I don't see any code in splash that would actually modify responses. Perhaps we could recreate QNetworkReply object in python?

pawelmhm:
I see access to protected functions is even listed as gotcha on PyQt docs page: http://pyqt.sourceforge.net/Docs/PyQt4/gotchas.html#access-to-protected-member-functions

kmike:
@pawelmhm thanks for looking at it - hmm, yes, it could be tricky. You can try creating a reply copy [here](https://github.com/scrapinghub/splash/blob/aec50a784e3ccf920c3acfa1e5ce26d273da4887/splash/network_manager.py#L380) and return it, but I'm not sure Python object will be preserved after Python -> C++ -> Python roundtrip. The same applies to a subclass - maybe it won't work; there was a similar issue with QNetworkRequest, and subclassing it didn't help - bare QNetworkRequest was returned instead of a subclass after going through C++ grounds.

pawelmhm:
to be clear we can still add splash:on_response_headers callback and have response object in lua with method abort() and some ways of accessing headers, this would still be useful for filtering out responses based on some criteria.

 Perhaps we can do this in first step, and try dealing with problem of protected methods later?

kmike:
Yes, response.abort() and response.get_header() are both very useful.

[...]

PR #246, opened 2015-06-25
On response headers

pawelmhm:
Adds support for Lua callback `on_response_headers`. Callback will accept response object as argument, for now this object will have two methods: `get_header` and `abort()`. Using callback will allow users to abort requests if some header is present in response (e.g. content_type). 

https://github.com/scrapinghub/splash/issues/231

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -1563,7 +1563,7 @@ request to Splash:
 .. _splash-on-response-headers:
 
 splash:on_response_headers
------------------
+--------------------------
 
 Register a function to be called after response headers are received, before 
 response body is read.
@@ -1582,14 +1582,23 @@ response body is read.
 * ``info`` - a table with response data in `HAR response`_ format
 * ``request`` - a table with request information 
 
+
 These fields are for information only; changing them doesn't change
 response received by splash. ``response`` has following methods:
 
 * ``response:abort()`` - aborts reading of response body
 
 A callback passed to :ref:`splash-on-response-headeers` can't call Splash
-async methods like :ref:`splash-wait` or :ref:`splash-go`. ``response`` cannot
-is cleared after exiting from callback, so you cannot use it outside callback.
+async methods like :ref:`splash-wait` or :ref:`splash-go`. ``response`` object
+is deleted after exiting from callback, so you cannot use it outside callback.
+
+``response.request`` available in callback contains following attributes:
+
+* ``url`` - requested URL - can be different from response URL in case there is
+  redirect
+* ``headers`` - HTTP headers of request
+* ``method`` HTTP method of request
+* ``cookies`` - cookies in .har format
 
 Example 1 - log content-type headers of all responses received while rendering
 
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -1560,6 +1560,80 @@ request to Splash:
 
     `splash:on_request` method doesn't support named arguments.
 
+.. _splash-on-response-headers:
+
+splash:on_response_headers
+-----------------
+
+Register a function to be called after response headers are received, before 
+response body is read.
+
+**Signature:** ``splash:on_response_headers(callback)``
+
+**Returns:** nil.
+
+**Async:** no.
+
+:ref:`splash-on-response-headers` callback receives a single ``response`` argument.
+``response`` contains following fields:
+
+* ``url`` - requested URL;
+* ``headers`` - HTTP headers of response
+* ``info`` - a table with response data in `HAR response`_ format
+* ``request`` - a table with request information 
+
+These fields are for information only; changing them doesn't change
+response received by splash. ``response`` has following methods:
+
+* ``response:abort()`` - aborts reading of response body
+
+A callback passed to :ref:`splash-on-response-headeers` can't call Splash
+async methods like :ref:`splash-wait` or :ref:`splash-go`. ``response`` cannot
+is cleared after exiting from callback, so you cannot use it outside callback.
+
+Example 1 - log content-type headers of all responses received while rendering
+
+.. code-block:: lua
+
+    function main(splash)
+        local all_headers = {}
+        splash:on_response_headers(function(response)
+            local content_type = response.headers["Content-Type"]
+            all_headers[response.url] = content_type
+        end)
+        assert(splash:go(splash.args.url))
+        return all_headers
+    end
+    
+Example 2 - abort reading body of all responses with content type ``text/css``
+
+.. code-block:: lua
+
+    function main(splash)
+        splash:on_response_headers(function(response)
+            local content_type = response.headers["Content-Type"]
+            if content_type == "text/css" then
+                response.abort()
+            end
+        end)
+        assert(splash:go(splash.args.url))
+        return splash:png()
+    end
+
+Example 3 - extract all cookies set by website without reading response body
+
+.. code-block:: lua
+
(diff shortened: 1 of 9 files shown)


splash/network_manager.py
```diff
@@ -266,9 +268,22 @@ def _handleFinished(self):
         self.log("Finished downloading {url}", reply)
 
     def _handleMetaData(self):
+        """Signal emitted before response body is reader, after
```
kmike: ready?

kmike:
WrappedRequest has one more feature - it disallows user to use itself outside of the callback (and breaks a GC cycle by removing a reference to LuaRuntime). Users can copy necessary request parameters if they need them. 

Do you think WrappedResponse shouldn't do it? There are two parts of the question:

1) we must ensure objects are not kept in memory longer than needed;
2) we should decide if using response outside callbacks is fine. `response.abort()` won't necessarily work outside a callback (response may be already downloaded), and keeping them for longer could lead to increased memory usage.

It could make sense to add _WrappedResponse to https://github.com/scrapinghub/splash/blob/aec50a784e3ccf920c3acfa1e5ce26d273da4887/splash/utils.py#L48 - this way _debug endpoint will be able to find alive _WrappedResponse objects.

splash/qtrender_lua.py
```diff
@@ -723,6 +737,38 @@ def set_header(self, name, value):
...
+        yield res
+
+
+class _WrappedResponse(object):
+    _attribute_whitelist = ['commands', "response"]
+
+    def __init__(self, lua, response):
+        self.lua = lua
+        self.response = response
+        self.headers = dict([(str(k), str(v)) for k, v in response.rawHeaderPairs()])
```
kmike: Splash requires Python 2.7+, so it can be written more succinctly: 
`{str(k): str(v) for k,v in response.rawHeaderPairs()}`

While header->name mapping is convenient, it is not always correct - there could be several response headers with the same name, it is allowed by HTTP (see e.g. http://stackoverflow.com/questions/4371328/are-duplicate-http-response-headers-acceptable).

kmike:
Hm, is there a way to get all response headers? To use response:get_header users need to know header names.

pawelmhm:
> Hm, is there a way to get all response headers? To use response:get_header users need to know header names.

ATM no, but this is easy to add  and definitely worth adding so will add that to this PR.

splash/qtrender_lua.py
pawelmhm: > While header->name mapping is convenient, it is not always correct 

this is valid point, but I see that PyQt4 QNetworkReply already handles this for us, if there are multiple headers with same key PyQT concatenates them together, you can test this like this

```python
def test_multiple_response_headers(self):
        resp = self.request_lua("""
        function main(splash)
            local h = nil
            splash:on_response_headers(function(response)
                h = response.get_headers()
            end)
            splash:http_get(splash.args.url)
            return h
        end
        """, {'url': "http://httpbin.org/response-headers?Set-Cookie=foo&Set-Cookie=bar&Set-Cookie=forest"})
```

above should return
```
{u'Access-Control-Allow-Credentials': u'true',
 u'Access-Control-Allow-Origin': u'*',
 u'Connection': u'keep-alive',
 u'Content-Length': u'130',
 u'Content-Type': u'application/json',
 u'Date': u'Tue, 07 Jul 2015 13:09:46 GMT',
 u'Server': u'nginx',
 u'Set-Cookie': u'foo\nbar\nforest'}
```
I see chrome devtools show multiple headers for this url:
```
Set-Cookie:foo
Set-Cookie:bar
Set-Cookie:forest
```

but they are concatenated by PyQt

kmike: That's actually great. Could you please add a test for it?

kmike:
This looks good! Could you please add

1) docs;
2) a test for `response:get_headers()`;
3) a test for multiple headers with the same name.

pawelmhm:
> 3) a test for multiple headers with the same name.

actually this is surprisingly difficult to test. We use twisted for mock resources but it looks like twisted does not allow to set multiple headers on response. [`request.setHeader(k, v)`](https://twistedmatrix.com/documents/current/api/twisted.web.http.Request.html#setHeader) keeps headers as dictionary, so if you try to set many headers with same key only last one will end up in response. To actually set multiple headers we could either find a way to do this with twisted mock resource (maybe somehow write raw response directly, but how?), or use remote resource for example httpbin.

pawelmhm:
hey @kmike can you check recent changes, I updated getting headers, removed method get_headers() and just added headers attribute to wrapped lua object making it available in lua script as a table (I think this should be easier for users). Also added basic response attributes: url, status and har info. If you think all ok I'll add docs.

splash/lua_modules/splash.lua
```diff
@@ -223,5 +223,42 @@ function Splash:on_request(cb)
...
+        url=py_reply.url,
+        status=py_reply.status,
+        info=py_reply.info
+    }
+
+    setmetatable(self, Response)
+
+    for key, value in pairs(py_reply.headers) do 
+        self[key] = value
+    end
```
kmike: Does it mean that if there is 'url' http header is the response user won't be able to access response.url attribute?

kmike:
The PR looks good, please go on :)

splash/lua_modules/splash.lua
pawelmhm: good catch looks like my mistake, will fix that

pawelmhm:
ok added docs and added request in callback @kmike

docs/scripting-ref.rst
```diff
@@ -1560,6 +1560,80 @@ request to Splash:
 
     `splash:on_request` method doesn't support named arguments.
 
+.. _splash-on-response-headers:
+
+splash:on_response_headers
+-----------------
```
kmike: the underline is too short, it won't render properly

splash/lua_modules/splash.lua
```diff
@@ -223,5 +223,52 @@ function Splash:on_request(cb)
...
+    local self = {
+        headers=py_reply.headers,
+        info=py_reply.info,
+        request=py_reply.request
+    }
+
+    setmetatable(self, Response)
+    
+    -- convert har headers to something more convenient
+    _request_headers = {}
```
kmike: Shouldn't it be local?

docs/scripting-ref.rst
```diff
@@ -1560,6 +1560,80 @@ request to Splash:
...
+
+**Async:** no.
+
+:ref:`splash-on-response-headers` callback receives a single ``response`` argument.
+``response`` contains following fields:
+
+* ``url`` - requested URL;
+* ``headers`` - HTTP headers of response
+* ``info`` - a table with response data in `HAR response`_ format
+* ``request`` - a table with request information
```
kmike: I think we should document what's available in response.request table.

splash/lua_modules/splash.lua
```diff
@@ -223,5 +223,52 @@ function Splash:on_request(cb)
...
+    for name, value in pairs(py_reply.request["headers"]) do
+        _request_headers[value["name"]] = value["value"]
+    end
+    py_reply.request["headers"] = _request_headers
```
kmike: Maybe we should provide this table for `splash:on_request` as well? It may be a separate ticket though, because for `splash:on_request` we need to make `request.headers` writeable.

kmike:
@pawelmhm there are some minor comments, but the PR looks solid :+1:

PR #251, opened 2015-07-14
[Work In Progress] Python 3 porting.

sunu:


sunu:
Hey @kmike !

This PR contains all the changes I have made so far. 

For now, I have only made just enough changes to make the all tests in `tests/test_execute.py` pass on Python 3 without any failures on Python 2.

I'm using my local fork of twisted for now which has a ported twisted.web.proxy for Python 3 (A patch for this is currently up for review on twisted issuetracker).

All tests are passing on Python 2. On Python 3, 8 tests in `test_execute.py` are failing. Here's a [gist of the test results](https://gist.github.com/sunu/3ae778f15de318908172) Out of them 7 are obvious (6 of them are due to the `u` prefix used in Python 2 to represent unicode strings. Another one is due to different error messages that Python 2 and Python 3 produce.) I'm thinking to use something like

```python
if _PY3:
    # compare py3 equivalent string
else:
    # compare py2 equivalent string
```
to test these cases. Is there a better way to do it?

The remaining is a test for timeout which fails and I haven't been able to find the reason yet :(

splash/browser_tab.py
```diff
@@ -24,6 +24,11 @@
 from splash.render_options import validate_size_str
 from splash.qwebpage import SplashQWebPage, SplashQWebView
 
+try:
+    unicode = unicode
+except NameError:
+    unicode = str
```
kmike: let's use six.text_type

splash/compat.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+
+import sys
+
+if sys.version_info < (3, 0):
+    _PY3 = False
+else:
+    _PY3 = True
```
kmike: we can use six.PY3

splash/har/log.py
```diff
@@ -8,6 +8,10 @@
 from PyQt5.QtWebKit import qWebKitVersion
 
 from .utils import get_duration, format_datetime, without_private
+from splash.compat import _PY3
+
+if _PY3:
+    basestring = (str, bytes)
```
kmike: I'm not sure we should accept both str and bytes in Python 3

splash/har/qt.py
```diff
@@ -114,7 +119,11 @@ def reply2har(reply, include_content=False, binary_content=False):
 
     status_text = reply.attribute(QNetworkRequest.HttpReasonPhraseAttribute)
     if status_text is not None:
-        res["statusText"] = bytes(status_text).decode('latin1')
+        try:
+            res["statusText"] = bytes(status_text, 'latin1')
+        except TypeError:
+            res["statusText"] = bytes(status_text).decode('latin1')
```
kmike: This code is not clear: what should be the type of `res['statusText']` - bytes or unicode? If it should be bytes then the second branch is wrong, if it should be unicode the first branch is wrong. I think in most cases the type of a value should be the same in Python 3 and Python 2.

splash/lua.py
```diff
@@ -203,7 +208,7 @@ def python2lua(lua, obj, max_depth=100):
     if isinstance(obj, dict):
         return lua.table_from({
             python2lua(lua, key, max_depth-1): python2lua(lua, value, max_depth-1)
-            for key, value in obj.iteritems()
+            for key, value in list(obj.items())
```
kmike: One option is six.iteritems, another is just `obj.items()`. `obj.items` will be a bit less efficient in Python 2.x. `list(obj.items())` looks pointless.

splash/cookies.py
```diff
@@ -78,23 +78,23 @@ def add(self, cookie):
     @classmethod
     def har_cookie2qt(cls, cookie):
         qcookie = QNetworkCookie()
-        qcookie.setName(cookie["name"])
-        qcookie.setValue(cookie["value"])
+        qcookie.setName(cookie[b"name"])
+        qcookie.setValue(cookie[b"value"])
```
kmike: It makes more sense to use unicode for key names, unless it is hard for some reason.

splash/proxy.py
```diff
@@ -11,12 +11,19 @@
 from __future__ import absolute_import
 import re
 import os
-import ConfigParser
+try:
+    import configparser
+except ImportError:
+    import ConfigParser as configparser
```
kmike: it should be in six.moves

splash/qtrender_lua.py
```diff
@@ -558,8 +568,12 @@ def set_result_header(self, name, value):
             raise ScriptError("splash:set_result_header() arguments must be strings")
 
         try:
-            name = name.decode('utf-8').encode('ascii')
-            value = value.decode('utf-8').encode('ascii')
+            try:
+                name = name.decode('utf-8').encode('ascii')
+                value = value.decode('utf-8').encode('ascii')
+            except AttributeError:
```
kmike: Why is it necessary? I.e. why can name and value be of different types, can't they have a single type both in Python 3.x and 2.x?

splash/qtrender_lua.py
```diff
@@ -783,8 +797,8 @@ def start(self, lua_source, sandboxed, lua_package_path,
 
         try:
             main_coro = self.get_main(lua_source)
-        except (ValueError, lupa.LuaSyntaxError, lupa.LuaError) as e:
-            raise ScriptError("lua_source: " + repr(e))
+        except Exception as e:
```
kmike: Why is a more broad except clause needed? 
Previously exceptions not from a list were propogated to the caller. I'm not sure caller was handling them though :)

splash/utils.py
```diff
@@ -30,14 +36,31 @@ def as_b64(self):
 class SplashJSONEncoder(json.JSONEncoder):
     def default(self, o):
         if isinstance(o, BinaryCapsule):
+            print("yup")
```
kmike: yup!

splash/utils.py
```diff
@@ -77,7 +100,7 @@ def get_ru_maxrss():
     return size
 
 
-def truncated(text, max_length=100, msg='...'):
+def truncated(text, max_length=100, msg=b'...'):
```
kmike: I think this function should work on unicode

sunu:
Thanks for the review @kmike. I have got almost all the tests passing on python3 now. Will polish things up and update the PR tomorrow :)

sunu:
Not ready for review yet. My harddrive is behaving strangely. I just pushed all of the changes just to be on the safer side in case it crashes. Will let you know when it's ready for next review.

splash/qtrender_lua.py
sunu: I guess I added it while debugging and forgot to remove.

sunu:
Hey @kmike !
It's ready for review. All tests passing on both Python 2 and 3 :D

splash/browser_tab.py
```diff
@@ -24,6 +25,8 @@
 from splash.render_options import validate_size_str
 from splash.qwebpage import SplashQWebPage, SplashQWebView
 
+unicode = six.text_type
```
kmike: let's just use six.text_type - it is confusing when builtins are redefined

splash/kernel/inspections.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, folder=None):
         for name in files:
             full_name = os.path.join(folder, name)
             with open(full_name, "rb") as f:
-                info = json.load(f, encoding='utf8')
+                info = json.loads(f.read().decode('utf-8'), encoding='utf8')
```
kmike: In Python 2 f.read() will return bytes, so calling .decode is not correct here

kmike:
Could you please fix merge conflicts?

splash/kernel/inspections.py
sunu: Umm.. sorry, I don't understand the issue. `decode` is converting the returned bytes to unicode. Why is that incorrect here?

sunu:
Will do.

splash/browser_tab.py
sunu: OK

sunu:
@kmike fixed the conflicts.

sunu:
Hey @kmike !

I have a question. According to [this question on SO](http://stackoverflow.com/questions/4400678/http-header-should-use-what-character-encoding), in case of headers, ASCII is the standard encoding. So should I use ASCII instead of UTF-8 in places like [this](https://github.com/scrapinghub/splash/pull/251/files#diff-3cb74b2bdc3baec8d0368069cdcab2fdR552) for example?

splash/har/log.py
```diff
@@ -6,10 +6,14 @@
 import splash
 from PyQt5.QtCore import PYQT_VERSION_STR, QT_VERSION_STR
 from PyQt5.QtWebKit import qWebKitVersion
+import six
 
 from .utils import get_duration, format_datetime, without_private
 
 
+if six.PY3:
+    basestring = str
```
kmike: It is not good to redefine builtins. You can use six.string_types here.

splash/har/qt.py
```diff
@@ -114,13 +115,17 @@ def reply2har(reply, include_content=False, binary_content=False):
 
     status_text = reply.attribute(QNetworkRequest.HttpReasonPhraseAttribute)
     if status_text is not None:
-        res["statusText"] = bytes(status_text).decode('latin1')
+        try:
+            res["statusText"] = bytes(status_text, 'latin1').decode('latin1')
+        except TypeError:
+            res["statusText"] = bytes(status_text).decode('latin1')
```
kmike: What is the purpose of this code? If status_text is unicode there is no need to convert it to bytes and then back to unicode.

splash/kernel/inspections.py
kmike: yep, you're right

splash/qtrender_lua.py
```diff
@@ -533,7 +536,11 @@ def delete_cookies(self, name=None, url=None):
     @command()
     def add_cookie(self, name, value, path=None, domain=None, expires=None,
                    httpOnly=None, secure=None):
-        cookie = dict(name=name, value=value)
+        name = name
+        value = value
+        cookie = dict()
+        cookie['name'] = name
+        cookie['value'] = value
```
kmike: why is this change needed?

splash/qtrender_lua.py
```diff
@@ -784,7 +791,7 @@ def start(self, lua_source, sandboxed, lua_package_path,
         try:
             main_coro = self.get_main(lua_source)
         except (ValueError, lupa.LuaSyntaxError, lupa.LuaError) as e:
-            raise ScriptError("lua_source: " + repr(e))
+            raise ScriptError(u"lua_source: " + repr(e))
```
kmike: Why is 'u' needed here?

splash/qtrender_png.py
```diff
@@ -6,15 +6,20 @@
...
 
 from PIL import Image
 from PyQt5.QtCore import QBuffer, QPoint, QRect, QSize, Qt
 from PyQt5.QtGui import QImage, QPainter, QRegion
+import six
 
 from splash import defaults
 
+
+xrange = six.moves.range
```
kmike: Let's not redefine builtins. If values are small & bounded we can use range, if not - six.moves.range.

splash/render_options.py
```diff
@@ -2,8 +2,15 @@
...
 import os
 import json
+from io import StringIO
+
+import six
+
 from splash import defaults
 
+if six.PY3:
+    basestring = (str, bytes)
```
kmike: we shouldn't redefine builtins; there is six.string_types

splash/render_options.py
```diff
@@ -28,21 +35,23 @@ def fromrequest(cls, request, max_timeout):
...
                 request.content.seek(0)
 
                 # 2. application/json POST data
-                if 'application/json' in content_type:
+                if 'application/json' in content_type.decode('utf-8'):
                     try:
-                        data.update(json.load(request.content, encoding='utf8'))
+                        content = request.content.read()
+                        data.update(json.load(
+                            StringIO(content.decode('utf-8')), encoding='utf8'))
```
kmike: Why is StringIO needed here? 
If you have a content as string you can use json.loads. 
Why isn't json.load working?

This may not be a problem here, but io.StringIO is worse than cStringIO.StringIO because it makes a copy of the data.

splash/render_options.py
```diff
@@ -28,21 +35,23 @@ def fromrequest(cls, request, max_timeout):
...
                     except ValueError as e:
-                        raise BadOption("Invalid JSON: '{}'".format(e.message))
+                        raise BadOption("Invalid JSON: '{}'".format(str(e)))
 
                 # 3. js_source from application/javascript POST requests
-                if 'application/javascript' in content_type:
-                    data['js_source'] = request.content.read()
+                if 'application/javascript' in content_type.decode('utf-8'):
```
kmike: I like `b'application/javascript' in content_type` more because in all other places we use bytes constants for header names

splash/render_options.py
```diff
@@ -52,25 +61,31 @@ def fromrequest(cls, request, max_timeout):
...
 
-        data['uid'] = id(request)
+        data[b'uid'] = id(request)
         return cls(data, max_timeout)
 
     def get(self, name, default=_REQUIRED, type=str, range=None):
         value = self.data.get(name)
+        if value is None and isinstance(name, bytes):
+            # In case of json post data the keys are unicode
+            value = self.data.get(name.decode('utf-8'))
```
kmike: it can be more clean to normalize `name` type before calling self.data.get

splash/render_options.py
```diff
@@ -169,6 +193,7 @@ def get_headers(self):
             raise BadOption("'headers' must be either JSON array of (name, value) pairs or JSON object")
 
         if isinstance(headers, (list, tuple)):
+            # import ipdb; ipdb.set_trace()
```
kmike: stray comment

splash/resources.py
```diff
@@ -38,7 +39,7 @@ def render(self, request):
             return Resource.render(self, request)
         except BadOption as e:
             request.setResponseCode(400)
-            return str(e) + "\n"
+            return str(e).encode('utf-8') + b"\n"
```
kmike: `str(e).encode('utf8')` is not correct in Python 2: you're calling .encode method for bytestring, and in thiscase Python first tries to convert it to unicode using sys.getdefaultencoding (usually ascii), and only then encode to the specified encoding. So it can fail with `UnicodeDecodeError` if str(e) contains non-ascii data.

splash/resources.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, pool, max_timeout, is_proxy_request=False):
         self.max_timeout = max_timeout
 
     def render_GET(self, request):
-        #log.msg("%s %s %s %s" % (id(request), request.method, request.path, request.args))
+        # log.msg("%s %s %s %s" % (id(request), request.method, request.path, request.args))
```
kmike: Here we followed a convention: for comments use `# comment`, but for commented out code use `#code`. I don't mind changing it though.

splash/resources.py
```diff
@@ -158,7 +165,7 @@ def _internalError(self, failure, request):
     def _badRequest(self, failure, request):
         failure.trap(BadOption)
         request.setResponseCode(400)
-        request.write(str(failure.value) + "\n")
+        request.write(str(failure.value).encode('utf-8') + b"\n")
```
kmike: `str(..).encode(..)` is a bad pattern in Python 2.x

kmike:
Hey @sunu! I think the main issue is that we're doing encoding and decoding here and there; it doesn't look consistent. I think we need to define places where we get bytes, convert data to unicode there and work on unicode most of time. 

So, the question is where can we get bytes from. One place is Twisted HTTP API - it makes sense to decode the data in Twisted resources before passing it to all other components. I'm not sure about Lua/lupa and Qt - do we get bytes or unicode from them, and what do we sending? Once we have answers to these questions it should become more clear where should we get bytes and where should we get unicode data.

sunu:
Hey @kmike,
Twisted HTTP API basically expects bytes and returns bytes in most cases. PyQt mostly deals with unicode. And lua2python also returns bytes since we use binary=True.
Let me take a better look to decide where the encoding/decoding should occur.

splash/har/qt.py
sunu: status_text is not unicode. It's a Qt byte array. So it needs to be converted into bytes and then decoded into unicode.

splash/qtrender_lua.py
sunu: I made the change when I was using bytes as keys. Now that we're using unicode keys this is no longer required. I'll fix it.

splash/render_options.py
sunu: I think json.load doesn't work here because request.content contains bytes. I will just use json.loads with the decoded string instead.

splash/render_options.py
sunu: Yes. makes sense.

kmike:
Hey @sunu,

Sorry, I missed your question:

> I have a question. According to this question on SO, in case of headers, ASCII is the standard encoding. So should I use ASCII instead of UTF-8 in places like this for example?

In tests it doesn't matter :) In general, it depends. Cookies are often encoded to UTF-8. Other headers are often encoded to latin1. Using ASCII when sending data is a good way to ensure compatibility.

splash/resources.py
sunu: What's most elegant way to do this? I'm thinking of just leaving out the encode part on python 2 since str(e) is essentially just bytes.

kmike: Yes, the way to go is to leave out .encode in Python 2.x. If this needs to be done several times you can e.g. add a function like similar to Scrapy's [to_bytes](https://github.com/scrapy/scrapy/blob/936266a61a1ae621e10d48ceb9633a7e245d017b/scrapy/utils/python.py#L110).

sunu:
@kmike So I looked at where we should be doing all the encoding/decoding.

* resources.py and mockserver.py - Twisted response etc
* render_options.py - extracting parameters from request
* proxy_server.py - for header manipulation

Other than these, there are places where we have to convert Qt bytearrays or the return value from lua_to_python to unicode value in qtrender_lua.py

kmike:
@sunu sounds good! 

So BrowserTab should only work with unicode and raise an exception if bytes data is passed?

sunu:
@kmike The failing tests on Python3 should pass after [this Twisted ticket](https://twistedmatrix.com/trac/ticket/7981) is fixed.
Apart from that, I did some clean up for the encode/decode stuff. Should I explicitly check for unicode in BrowserTab? It would be really helpful if you could take a look at the current changes and suggest whether any changes are necessary. :)

.travis.yml
```diff
@@ -51,9 +57,9 @@ before_install:
     # Builds Complete
     - popd
     # pip installs
-    - pip install psutil requests jsonschema strict-rfc3339 xvfbwrapper pytest funcparserlib
-    - pip install git+https://github.com/sunu/qt5reactor
-    - "if [[ $SYSTEM_PACKAGES == 'false' ]]; then pip install -UI Pillow twisted service_identity lupa; fi"
+    - pip install psutil requests jsonschema strict-rfc3339 xvfbwrapper pytest funcparserlib qt5reactor-fork six
+    - if [[ $SYSTEM_PACKAGES == 'false' ]]; then pip install -UI Pillow service_identity lupa; fi
+    - if [[ $TRAVIS_PYTHON_VERSION == 3.4 ]]; then git clone https://github.com/twisted/twisted; pushd twisted; python setup.py install; popd; else pip install twisted; fi
```
kmike: I think `pip install https://github.com/twisted/twisted/archive/trunk.zip#egg=twisted` should work

splash/browser_tab.py
```diff
@@ -672,7 +671,7 @@ def html(self):
         """ Return HTML of the current main frame """
         self.logger.log("getting HTML", min_level=2)
         frame = self.web_page.mainFrame()
-        result = bytes(frame.toHtml().encode('utf-8'))
+        result = frame.toHtml().encode('utf-8')
```
kmike: Do you think we should return the contents as bytes here?

splash/har/qt.py
kmike: this makes sense, thanks for the explanation

splash/proxy.py
```diff
@@ -104,6 +105,8 @@ def _getFilterParams(self, profile_name=None):
         Return (blacklist, whitelist, proxy_list) tuple
         loaded from profile ``profile_name``.
         """
+        if isinstance(profile_name, bytes):
```
kmike: When is profile_name bytes?

splash/qtrender.py
```diff
@@ -229,5 +229,5 @@ def get_result(self):
 
 class HarRender(DefaultRenderScript):
     def get_result(self):
-        return json.dumps(self.tab.har())
+        return json.dumps((self.tab.har()))
```
kmike: extra () are not needed

.travis.yml
sunu: I did try `pip install git+https://github.com/twisted/twisted` and for some reason it didn't work.

splash/browser_tab.py
sunu: Hmm .. as far as I understand, we feed this result into python2lua which ultimately converts unicode into bytes. So returning `frame.toHtml()` as is should be ok and more consistent.

splash/render_options.py
```diff
@@ -51,30 +57,35 @@ def fromrequest(cls, request, max_timeout):
...
         return cls(data, max_timeout)
 
     def get(self, name, default=_REQUIRED, type=str, range=None):
         value = self.data.get(name)
         if value is not None:
             if type is not None:
-                if type is str and isinstance(value, unicode):
-                    value = value.encode('utf8')
+                if type is str and isinstance(value, six.text_type):
+                    value = value.encode('utf-8')
```
kmike: What about making RenderOptions work on unicode, i.e. accepting unicode and returning unicode? HTTP resources can be in charge of decoding/encoding data.

splash/render_options.py
```diff
@@ -170,14 +193,20 @@ def get_headers(self):
...
-                if not (isinstance(el, (list, tuple)) and len(el) == 2 and all(isinstance(e, basestring) for e in el)):
+                if not (isinstance(el, (list, tuple)) and len(el) == 2 and all(
+                        isinstance(e, (six.string_types, six.binary_type)) for e
+                        in el)):
                     raise BadOption("'headers' must be either JSON array of (name, value) pairs or JSON object")
-
+        try:
+            headers = bytes_to_unicode(headers, encoding='ascii')
+        except UnicodeDecodeError:
+            raise BadOption("headers can not contain non-ascii characters.")
```
kmike: What if they must contain these characters, i.e. what if user wants to send non-ascii cookies? Many servers accept utf8-encoded cookies.

splash/resources.py
```diff
@@ -102,6 +105,7 @@ def _writeOutput(self, data, request, content_type=None):
             content_type = self.content_type
 
         if isinstance(data, (dict, list)):
+            data = bytes_to_unicode(data)
```
kmike: why can data contain bytes here?

splash/resources.py
```diff
@@ -340,10 +344,10 @@ def render_GET(self, request):
         url = params['url']
         if not url.lower().startswith('http'):
             url = 'http://' + url
-        url = url.encode('utf8')
-        params = {k:v for k,v in params.items() if v is not None}
+        params = {k: v for k, v in params.items() if v is not None}
 
-        request.addCookie('phaseInterval', 120000)  # disable "phases" HAR Viewer feature
+        request.addCookie('phaseInterval',
+                          120000)  # disable "phases" HAR Viewer feature
```
kmike: I prefer to move the comment above this line instead of wraping it like this

splash/server.py
```diff
@@ -86,11 +87,12 @@ def parse_opts():
...
-    if opts.logfile:
-        logfile = DailyLogFile.fromFullPath(opts.logfile)
-    else:
-        logfile = sys.stderr
+    # TODO: fix this.
+    # from twisted.python.logfile import DailyLogFile
+    # if opts.logfile:
+    #     logfile = DailyLogFile.fromFullPath(opts.logfile)
+    # else:
+    logfile = sys.stderr
```
kmike: I think we should raise an exception in Python 3 if user uses --logfile option instead of silently ignoring it. Just moving import under `if` could work: 

```
if opts.logfile:
    from twisted.python.logfile import DailyLogFile
    logfile = DailyLogFile.fromFullPath(opts.logfile)
else:
    logfile = sys.stderr
```

splash/proxy.py
sunu: profile_name is always bytes whenever it is provided as the proxy keyword. I should move the decoding into render_options.py#125

splash/tests/test_execute.py
```diff
@@ -2166,8 +2190,8 @@ def test_get_perf_stats(self):
         end
         """
         out = self.request_lua(func).json()
-        self.assertItemsEqual(out.keys(),
-                              ['walltime', 'cputime', 'maxrss'])
+        self.assertEqual(list(out.keys()).sort(),
```
kmike: `[].sort()` returns None, so this test now checks nothing. It should be `sorted(out.keys())` or `set(out.keys())`.

splash/tests/test_render.py
```diff
@@ -38,14 +39,15 @@ def wrapper(self, *args, **kwargs):
...
-    endpoint = "render.html"
+    endpoint = u"render.html"
 
     def __init__(self, ts):
         self.ts = ts
 
     @property
     def host(self):
-        return "localhost:%s" % self.ts.splashserver.portnum
+        print(u"localhost:%s" % self.ts.splashserver.portnum)
```
kmike: stray print statement?

splash/utils.py
```diff
@@ -84,19 +164,21 @@ def get_total_phymem():
...
+    except AttributeError:  # psutil < 2.0
         return psutil.phymem_usage().total
 
 
-def truncated(text, max_length=100, msg='...'):
+def truncated(text, max_length=100, msg=u'...'):
     """
     >>> truncated("hello world!", 5)
-    'hello...'
+    u'hello...'
```
kmike: This doctest won't work in Python 3.x.

splash/utils.py
```diff
@@ -84,19 +164,21 @@ def get_total_phymem():
     """ Return the total amount of physical memory available. """
     try:
         return psutil.virtual_memory().total
-    except AttributeError: # psutil < 2.0
+    except AttributeError:  # psutil < 2.0
         return psutil.phymem_usage().total
 
 
-def truncated(text, max_length=100, msg='...'):
+def truncated(text, max_length=100, msg=u'...'):
```
kmike: I think it doesn't matter if we leave `msg='...'`, but it could make doctest work in 2.x and 3.x

splash/browser_tab.py
kmike: It is a bit strange that some methods return bytes and some methods return unicode. E.g. `set_contents` requires unicode, but `html`returns bytes. 

toHtml returns QString, which is basically qt's unicode, right? Why are you encoding it to utf8 here?

splash/resources.py
sunu: data almost always contains bytes because lupa always returns bytes, not unicode.

splash/resources.py
sunu: weird auto folding by PyCharm. WIll fix.

splash/server.py
sunu: Hmm makes sense. The twisted.python.logfile porting is stuck on some Windows only corner case.

kmike:
hey @sunu - the PR starts to look good, good job!

Some extra notes:

1. I don't like bytes_to_unicode and unicode_to_bytes functions; it looks like they shouldn't be necessary if we are handling unicode consistently. I may be wrong, but they look like a band-aid solution which one can always apply to fix bytes/unicode inconsistency, but we shouldn't have this inconsistency in first place.

2. It also makes sense to convert all the input to unicode earlier, likely in RenderOptions.fromrequest.

3. Unfortunately (or fortunately) there is quite a few new commits in Splash master, and https://github.com/scrapinghub/splash/pull/255 is going to be merged soon; I'll merge these new changes to qt5 branch, so likely this PR will require a rebase or merge to take new changes in account.

splash/render_options.py
kmike: or RenderOptions.fromrequest method

splash/browser_tab.py
sunu: toHtml returns unicode, not even Qstring. So yeah, encoding it is pointless.

splash/tests/test_execute.py
sunu: Oops. :D

splash/utils.py
sunu: Hmm. Makes sense.

splash/resources.py
kmike: What about fixing it with `splash.lua.lua2python` function in appropriate places? We should apply `splash.lua.lua2python` with care because Lua <-> Python converison is not loseless, but there shouldn't be any problems if we apply this function e.g. to the result of Lua's `main` function (if it is not applied yet).

sunu:
Umm .. the rebase didn't work for some reason :confused:

.travis.yml
kmike: `pip install https://github.com/twisted/twisted/archive/trunk.zip#egg=twisted` is better because it doesn't even require git. Anyways, they released twisted 15.3.0 yesterday

sunu: The ported twisted.web.proxy is not yet included in the release :(

kmike: ouch, good to know

sunu:
All tests pass on my machine now, but travis is acting up.

kmike:
Hey @sunu - it looks like there are still extra commits in this branch. Could you please try clean it up?

sunu:
@kmike which one exactly? The "remove jpeg scaling tests" commit some how is placed far back in history due the rebase mess up. (I didn't notice that you have already merged master into qt5 and rebased it :(). Changes from that commit are getting overwritten by changes from other commits later in log.

sunu:
@kmike should be fixed now. Seems like missed those changes which merging.

kmike:
@sunu I mean that "authored by .." commits shouldn't be in this PR - we'll have 2 slightly different copies of e.g. https://github.com/sunu/splash/commit/66fcdb1bc56153df7d77c98dbd470c02e7e129c3 if this PR will be merged as-is.

We have 2 branches - qt5 and py3:
```
---QQQQ
   \PPP
```

After rebase it should look like this:

```
---QQQQ
       \PPP
```

If you prefer merge it should look like this (an extra merge commit):
```
---QQQQ
   \PPP\M
```

But now it looks like this - commits from qt5 branch are duplicated:
```
---QQQQ
   \PPPqqqq
```

I guess you need to checkout https://github.com/sunu/splash/commit/f43d6c204893cfdfc97641f973f946ecad269ca9, redo the rebase (or do a merge) and then cherry-pick https://github.com/sunu/splash/commit/313ffdeb5590fd4777f6aeadf1a0728c87e0bb7e. I'm far from being a git guru, so maybe there is another way to fix it.

kmike:
I think tests were failing because I broke qt5 branch by not pulling https://github.com/scrapinghub/splash/commit/e6b050750280e12d0fa1e77f025f6f8f346382c3. This is fixed now.

sunu:
Whew ..  looks like it's all fixed now.

splash/lua.py
```diff
@@ -203,14 +204,14 @@ def python2lua(lua, obj, max_depth=100):
     if isinstance(obj, dict):
         return lua.table_from({
             python2lua(lua, key, max_depth-1): python2lua(lua, value, max_depth-1)
-            for key, value in obj.iteritems()
+            for key, value in obj.items()
```
kmike: let's use six.iteritems here

splash/qtrender_image.py
```diff
@@ -67,10 +68,17 @@ def qimage_to_pil_image(self, qimage):
...
+        try:
+            return Image.frombytes(
+                self.pillow_image_format,
+                self._qsize_to_tuple(qimage.size()),
+                buf, 'raw', self.pillow_decoder_format)
+        except AttributeError:
+            return Image.fromstring(
+                self.pillow_image_format,
+                self._qsize_to_tuple(qimage.size()),
+                buf, 'raw', self.pillow_decoder_format)
```
kmike: I think it is better to do something like https://github.com/kmike/psd-tools/blob/7efb5c6342373bf5a0f663892de98f0fb07f33db/src/psd_tools/user_api/pil_support.py#L14 to avoid duplicating all the arguments

splash/qtrender_image.py
```diff
@@ -457,13 +465,13 @@ def crop(self, rect):
         self.img = self.img.crop((left, top, right, bottom))
 
     def to_png(self, complevel=defaults.PNG_COMPRESSION_LEVEL):
-        buf = StringIO()
+        buf = BytesIO()
```
kmike: Could you please check that this change doesn't make performance worse in Python 2.x?

kmike: Using cStringIO in Python 2.x is a safe choice

splash/render_options.py
```diff
@@ -27,54 +30,55 @@ def fromrequest(cls, request, max_timeout):
...
                 for value in values
-            ]
+                ]
             data.setdefault('headers', headers)
-            data.setdefault('http_method', request.method)
+            data.setdefault('http_method', request.method.decode('utf-8'))
 
             request.content.seek(0)
-            data.setdefault('body', request.content.read())
+            data.setdefault('body', request.content.read().decode('utf-8'))
```
kmike: I'm not sure what implications does this change have for POST requests with binary bodies. I'm also not sure if we should care about it ;)

splash/render_options.py
```diff
@@ -169,13 +176,16 @@ def get_headers(self):
...
+                "'headers' must be either JSON array of (name, value) pairs or JSON object")
 
         if isinstance(headers, (list, tuple)):
             for el in headers:
-                if not (isinstance(el, (list, tuple)) and len(el) == 2 and all(isinstance(e, basestring) for e in el)):
-                    raise BadOption("'headers' must be either JSON array of (name, value) pairs or JSON object")
-
+                if not (isinstance(el, (list, tuple)) and len(el) == 2 and all(
+                        isinstance(e, six.string_types) for e
+                        in el)):
```
kmike: I think it is better to move `all(isinstance(e, six.string_types) for e in el)` to a variable - formatting look weird here

splash/resources.py
```diff
@@ -112,21 +115,26 @@ def _writeOutput(self, data, request, content_type=None):
...
         request.write(data)
 
     def _logStats(self, request):
         stats = {
-            "path": request.path,
-            "args": request.args,
+            # Anything we retrieve from Twisted request object contains bytes.
+            # We have to convert it to unicode first for json.dump to succeed.
+            "path": request.path.decode('utf-8'),
+            "args": bytes_to_unicode(request.args),
```
kmike: This is the only place `bytes_to_unicode` is used, and we don't need to do a recursive decoding here. I think we should drop `bytes_to_unicode` and do decoding the same way as in `RequestOptions.fromrequest`. You can add a (much smaller) utility function to convert Twisted request.args to unicode or just duplicate the code.

splash/tests/mockserver.py
```diff
@@ -15,7 +17,8 @@
 
 _REQUIRED = object()
 
-def getarg(request, name, default=_REQUIRED, type=str):
+
+def getarg(request, name, default=_REQUIRED, type=bytes):
```
kmike: Can code be simplified if we keep using `str` here instead of `bytes`? I see results of `getarg` are often decoded.

kmike:
Hey @sunu! I think we're close :) 

Could you please check cases like [this](https://github.com/scrapinghub/splash/blob/64014968fa6e9d5a88fd131b246ddfc875a50039/splash/qtrender.py#L28) - Python 3 has a different syntax for metaclasses, so we need to use something like six.with_metaclass.

I made 1.7 release today, and I hope it'll be the last qt4-based release (unless we face some big issues with qt5). Thanks for the great work so far!

kmike:
@sunu please be careful with merging/rebasing today's qt5 branch changes - see https://github.com/scrapinghub/splash/pull/260#issuecomment-128521056.

sunu:
@kmike Should I just skip the hanging tests after merging?

splash/qtrender_image.py
sunu: Hey @kmike . I ran some performance checks and didn't see much difference. The time and memory consumed remained almost the same. Am I doing something wrong? Here is what I did.
```
In [5]:

from PIL import Image
In [3]:

from io import BytesIO
def to_png_bytesio(complevel=1):
    img = Image.open("pillow.jpg")
    buf = BytesIO()
    img.save(buf, 'png', compress_level=complevel)
    return buf.getvalue()
In [24]:

%timeit -n 1000 to_png_bytesio()
1000 loops, best of 3: 3.89 ms per loop
In [25]:

from cStringIO import StringIO
def to_png_stringio(complevel=1):
    img = Image.open("pillow.jpg")
    buf = StringIO()
    img.save(buf, 'png', compress_level=complevel)
    return buf.getvalue()
In [26]:

%timeit -n 1000 to_png_stringio()
1000 loops, best of 3: 3.9 ms per loop
In [1]:

%load_ext memory_profiler
In [8]:

from benchmark import to_png_bytesio
In [9]:

%mprun -f to_png_bytesio to_png_bytesio()
('',)

Filename: benchmark.py
​
Line #    Mem usage    Increment   Line Contents
================================================
    11     32.1 MiB      0.0 MiB   def to_png_bytesio(complevel=1):
    12     32.2 MiB      0.0 MiB       img = Image.open("pillow.jpg")
    13     32.2 MiB      0.0 MiB       buf = BytesIO()
    14     32.3 MiB      0.1 MiB       img.save(buf, 'png', compress_level=complevel)
    15     32.3 MiB      0.0 MiB       return buf.getvalue()
In [11]:

from benchmark import to_png_stringio
In [12]:

%mprun -f to_png_stringio to_png_stringio()
('',)

Filename: benchmark.py
​
Line #    Mem usage    Increment   Line Contents
================================================
     4     32.3 MiB      0.0 MiB   def to_png_stringio(complevel=1):
     5     32.3 MiB      0.0 MiB       img = Image.open("pillow.jpg")
     6     32.3 MiB      0.0 MiB       buf = StringIO()
     7     32.4 MiB      0.1 MiB       img.save(buf, 'png', compress_level=complevel)
     8     32.4 MiB      0.0 MiB       return buf.getvalue()
```

kmike: It could be the case difference is non-existing in practice, but could you please try it with a larger image, something like 1024x40000px? Such images are common when render_all=1 is used.

sunu: Still not much of a difference. Tried with http://imgur.com/q4pPirs
```
In [1]:

from PIL import Image
In [2]:

from io import BytesIO
def to_png_bytesio(complevel=1):
    img = Image.open("huge.png")
    buf = BytesIO()
    img.save(buf, 'png', compress_level=complevel)
    return buf.getvalue()
In [3]:

%timeit to_png_bytesio()
1 loops, best of 3: 1.57 s per loop
In [4]:

from cStringIO import StringIO
def to_png_stringio(complevel=1):
    img = Image.open("huge.png")
    buf = StringIO()
    img.save(buf, 'png', compress_level=complevel)
    return buf.getvalue()
In [5]:

%timeit to_png_stringio()
1 loops, best of 3: 1.56 s per loop
In [6]:

%load_ext memory_profiler
In [7]:

from benchmark import to_png_bytesio
In [8]:

%mprun -f to_png_bytesio to_png_bytesio()
('',)

Filename: benchmark.py
​
Line #    Mem usage    Increment   Line Contents
================================================
    11     30.9 MiB      0.0 MiB   def to_png_bytesio(complevel=1):
    12     30.9 MiB      0.0 MiB       img = Image.open("huge.png")
    13     30.9 MiB      0.0 MiB       buf = BytesIO()
    14    189.1 MiB    158.2 MiB       img.save(buf, 'png', compress_level=complevel)
    15    189.1 MiB      0.0 MiB       return buf.getvalue()
In [9]:

from benchmark import to_png_stringio
In [11]:

%mprun -f to_png_stringio to_png_stringio()
('',)

Filename: benchmark.py
​
Line #    Mem usage    Increment   Line Contents
================================================
     4     31.0 MiB      0.0 MiB   def to_png_stringio(complevel=1):
     5     31.0 MiB      0.0 MiB       img = Image.open("huge.png")
     6     31.0 MiB      0.0 MiB       buf = StringIO()
     7    189.1 MiB    158.1 MiB       img.save(buf, 'png', compress_level=complevel)
     8    189.6 MiB      0.5 MiB       return buf.getvalue()
In [ ]:

​
```

kmike: ok, cool, so we are fine here. Thanks for checking it!

kmike:
@sunu yes, I think we should skip tests for now, to wrap up QT5 / Python 3 port. But at the same time I'm not comfortable with making qt5 default (i.e. merging qt5 branch into master) until the issue is fixed.

Could you please 

1. Submit a PR against qt5 branch to skip these tests;
2. Merge or rebase Python 3 port on top of updated qt5 branch after I merge this PR?

sunu:
@kmike Just sent the PR skipping the `resource_timeout` test. #272 
Regarding not merging qt5 into master, yes that makes perfect sense.

dockerfiles/splash/provision.sh
```diff
@@ -65,23 +81,23 @@ install_builddeps () {
 
 install_python_deps () {
     # Install python-level dependencies.
-    pip install -U pip && \
-    /usr/local/bin/pip install --no-cache-dir \
-        Twisted==15.2.1 \
-        qt4reactor==1.6 \
+    pip3 install -U pip && \
+    pip3 install -U https://github.com/twisted/twisted/archive/trunk.zip#egg=twisted && \
```
kmike: If a latest commit is needed we should pin package versions in Dockerfile, e.g. `https://github.com/twisted/twisted/archive/a5108c0db829847bbfd4bf427868eb0b13da0465.zip`.

kmike:
Thanks @sunu for good work! It is continued here: https://github.com/scrapinghub/splash/pull/288

Issue #214, opened 2015-04-28
add JPEG support


kmike:

pawelmhm:
+1

We need to convert images from png to jpeg (customer wants jpeg). It would be much nicer to just get JPEG from Splash.  

is anything blocking this?

kmike:
Nothing blocks it, just a lack of implementation.

kmike:
Fixed by #255.

PR #255, opened 2015-07-27
Add JPEG endpoint

chekunkov:
Addresses #214

TODO:

- [x] rename qtrender_png.py to qtrender_image.py
- [x] wrap most of the functions in qtrender_image.py in QtImageRenderer class
- [x] add support for RGB image format (JPEG doesn't work well with ARGB)
- [x] add appropriate resource, renderer and url mapping
- [x] check Qt performance with images in RGB format
- [x] add tests
- [x] add documentation
- [x] add jpeg support in render.json

Work on this endpoint was started during EuroPython 2015 sprints


kmike:
Hey @chekunkov,

I'm going to make Splash 1.7 release soon; likely it will be the last qt4-based Splash release. Do you want JPEG-related changes to be in this release?

Also, in Pillow 2.9 there is a function for converting QImage to Pillow Image (see https://github.com/python-pillow/Pillow/pull/1217) - I wonder if we should switch to it?

splash/browser_tab.py
```diff
@@ -674,32 +675,52 @@ def html(self):
         self.store_har_timing("_onHtmlRendered")
         return result
 
-    def png(self, width=None, height=None, b64=False, render_all=False,
-            scale_method=None):
-        """ Return screenshot in PNG format """
-        self.logger.log("Getting PNG: width=%s, height=%s,"
-                        " render_all=%s, scale_method=%s" %
-                        (width, height, render_all, scale_method), min_level=2)
+    @contextmanager
```
immerrr: Why is contextmanager needed here?

splash/render_options.py
```diff
@@ -248,11 +251,23 @@ def get_common_params(self, js_profiles_path):
...
+            'scale_method': self.get_scale_method()
+        }
+
     def get_png_params(self):
-        return {'width': self.get_width(), 'height': self.get_height(),
-                'scale_method': self.get_scale_method()}
+        return self.get_image_params()
+
+    def get_jpeg_params(self):
+        params = {'quality': self.get_quality()}
```
immerrr: This seems a bit tricky-ish. Why do we want quality for one of the image rendering methods and disregard it for the other one?

splash/browser_tab.py
chekunkov: check how it's used. it allows conveniently reuse code in both png and jpeg methods

splash/render_options.py
chekunkov: I noticed that png and jpeg parameters are the same except jpeg also should allow `quality` parameter which png endpoint does not support. That's why I created method with common parameters (looks like width, height and scale_method apply to any image format) and added specific parameter to jpeg endpoint

chekunkov:
> Also, in Pillow 2.9 there is a function for converting QImage to Pillow Image (see python-pillow/Pillow#1217) - I wonder if we should switch to it?

yep, I discovered this function at Saturday night, but I believe it's safe to add it in a separate PR, current implementation works quite well for me, but I still need to compare performance.

> I'm going to make Splash 1.7 release soon; likely it will be the last qt4-based Splash release. Do you want JPEG-related changes to be in this release?

definitely. do I have time until the beginning of the next week?

kmike:
@chekunkov yes, beginning of the next week sounds good

splash/browser_tab.py
chekunkov: to be more precise - because originally this method covered with `try...finally` all image conversion flow. Now it does the same, but along with code reuse allows some endpoint-specific things, like different logging message and different har timing event name

immerrr: Yes, but the try/finally contains only a rollback of setViewportSize, so the only difference between contextmanager/yield/with and function/return is when that happens. It seems that it is ok to do it either way and contextmanager can be dropped, or can it not?

splash/render_options.py
immerrr: Ah, ok. I skimmed through the sources and remembered that both png and jpeg functions had some argument to apply (for png it is complevel), and it felt somehow "unfair" to only have one of those functions to have this parameter overridable.

docs/scripting-ref.rst
```diff
@@ -917,6 +917,87 @@ If your script returns the result of ``splash:png()`` in a top-level
...
+.. code-block:: lua
+
+     function main(splash)
+         assert(splash:go(splash.args.url))
+         return {jpeg=splash:jpeg()}
+     end
+
+If your script returns the result of ``splash:jpeg()`` in a top-level
+``"jpeg"`` key (as we've done in a previous example) then Splash UI
+will display it as an image.
```
kmike: Is this true? It seems there are no changes to Splash UI (see resources.py DemoUI).

chekunkov: oops, not sure, I need to check that. I mostly copied png description :)

splash/browser_tab.py
chekunkov: ur right, it can be dropped as resizing still happens in try...finally, fixed

docs/scripting-ref.rst
chekunkov: removed this sentence from docs

chekunkov:
I'm happy to announce that this PR is ready for a review. According to benchmarks rendering JPEG is 1.5x-2x faster than rendering PNG.

docs/scripting-ref.rst
```diff
@@ -917,6 +917,84 @@ If your script returns the result of ``splash:png()`` in a top-level
...
+*width* parameter sets the width of the resulting image.  If the viewport has a
+different width, the image is scaled up or down to match the specified one.
+For example, if the viewport is 1024px wide then ``splash:jpeg{width=100}`` will
+return a screenshot of the whole viewport, but the image will be downscaled to
+100px width.
+
+*height* parameter sets the height of the resulting image.  If the viewport has
+a different height, the image is trimmed or extended vertically to match the
+specified one without resizing the content.  The region created by such
+extension is transparent.
```
kmike: what does 'The region created by such extension is transparent.' mean for JPEG?

splash/qtrender_image.py
```diff
@@ -0,0 +1,471 @@
+
+"""
+This module handles rendering QWebPage into PNG images.
```
kmike: PNG or JPEG?

splash/tests/lua_modules/emulation.lua
```diff
@@ -102,4 +102,19 @@ function emulation.render_png(splash)
...
+    local render_all = (splash.args.render_all or
+            splash.args.viewport == "full")
+    return splash:jpeg{
+        width=splash.args.width,
+        height=splash.args.height,
+        render_all=render_all,
+        scale_method=splash.args.scale_method,
+        quality=splash.args.quality,
+    }
+end
```
kmike: It seems this function is unused in tests. Could you please add a test to test_execute_emulation?

Proxy interface also look untested; to enable it add a test case to test_proxy_server (no need to test with use_gzip=True).

docs/scripting-ref.rst
immerrr: @kmike Nice catch :)

docs/scripting-ref.rst
```diff
@@ -917,6 +917,84 @@ If your script returns the result of ``splash:png()`` in a top-level
...
+* render_all - optional, if ``true`` render the whole webpage;
+* scale_method - optional, method to use when resizing the image, ``'raster'``
+  or ``'vector'``
+* quality - optional, quality of JPEG image, integer in range from ``0`` to ``100``
+
+**Returns:** JPEG screenshot data.
+
+**Async:** no.
+
+Without arguments ``splash:jpeg()`` will take a snapshot of the current viewport.
```
immerrr: I wonder if we actually need all of that documentation for JPEG endpoint. I mean, if anything changes we'll have one more place to fix it. Maybe we could refer to relevant portions of `splash:png` and just underline the differences?

kmike: I also don't like the duplication, but full docstring is nice e.g. for autocompletion results.

One option is to provide something like `splash:screenshot{format='jpeg'}` function, and document splash:png and splash:jpeg only as shortcuts.

chekunkov: Yes, main reason why I just copy almost everything is that autocompletion shows all this documentation when you're using Splash with Jupyter.

I can probably remove two last blocks which start from ```If the result of ``splash:jpeg()`` ``` and just mention that it works the same as `splash:png()`. 

It's true that 4 parameters have almost the same documentation, but still there are some specifics for each format - like @kmike noticed extended height would be transparent for PNG and white for JPEG. I'm not sure we can completely omit this duplication problem, and while we have only 2 image formats - it should be fine to have some duplication in docs.

> One option is to provide something like splash:screenshot{format='jpeg'} function, and document splash:png and splash:jpeg only as shortcuts.

this method should support 2 different signatures - `quality` for jpeg only, also `transparency` for png only if we consider to add it (as discussed with @kmike) - wouldn't it be counterintuitive to have method with such dynamic signature? Documentation for this method will have to much ifs - like `use this parameter if format is 'jpeg'`. Also it would require much better input validation in lua than we have now.

`splash:screenshot` -> `splash:image{format='jpeg'}` as a shorter alternative

splash/tests/lua_modules/emulation.lua
chekunkov: Fixed

splash/qtrender_image.py
chekunkov: AND :)

docs/scripting-ref.rst
chekunkov: 'is white'. Fixed

docs/scripting-ref.rst
kmike: Another possible name is `splash:render{'jpeg', ...}`. 

+1 to keeping docstrings duplicated for now. I think that this documentation issue is not that important, a new method will require more discussion, and de-duplicating the docstrings without creating a new method won't make things better.

chekunkov: There's one more trick we can do with docs for these 2 methods - we can create some section related to image processing in docs with common howtos and limitations, remove some verbose parts from parameters docs in script reference docs - leave only concise one line description - and reference newly created section where it applies.

+1 to moving this problem out of the current PR's scope

kmike:
:balloon: :balloon: :balloon:

immerrr:
:tada:

PR #273, opened 2015-08-11
pytest.skip doesn't do what we want.

sunu:
Hey @kmike,
#272 actually doesn't do what I expected. `pytest.skip` actually skips all the tests in the file. It's an issue in pytest https://github.com/pytest-dev/pytest/issues/607. I managed to missed that somehow.


kmike:
Ouch, thanks. I was bitten by this myself in past.

Issue #262, opened 2015-08-04
improve splash logging 


pawelmhm:
We are currently using splash /execute endpoint making POST requests, and it is has become very difficult to monitor splash and understand how it is being used.  If you make GET to Splash you can at least see url and params in query string, but if you make POST to splash server only data that goes to logs is following:

```
2015-08-04 17:46:47.024152 [pool] [139898212308896] SLOT 0 is closing <splash.qtrender.HtmlRender object at 0x7f3c9740b390>
2015-08-04 17:46:47.024187 [render] [139898212308896] close is requested by a script
2015-08-04 17:46:47.024324 [pool] [139898212308896] SLOT 0 done with <splash.qtrender.HtmlRender object at 0x7f3c9740b390>
2015-08-04 17:46:47.024606 [stats] {"maxrss": 103376, "load": [0.36, 0.41, 0.41], "fds": 28, "qsize": 0, "rendertime": 0.4805610179901123, "active": 0, "path": "/render.html", "args": {}, "_id": 139898212308896}
2015-08-04 17:46:47.024821 [-] "127.0.0.1" - - [04/Aug/2015:15:46:46 +0000] "POST /render.html HTTP/1.1" 200 495 "-" "curl/7.35.0"
2015-08-04 17:46:47.024942 [pool] SLOT 0 is available

```
there is mention of splash stats, but no mention of url being crawled or parameters for splash. It would be cool if splash could print this information in logs when making POST.

kmike:
+1 to fix it somehow.

kmike:
Fixed by #274.

The logging can be improved further, e.g. some information should be printed when Splash receives the request, not only when it finishes it. Currently if a request is failing there is not much information in logs. But it is a separate issue.

PR #274, opened 2015-08-13
[logging] improved logging

pawelmhm:
* message with splash.args should be logged for all requests, both for GET and POST
requests to splash
* adds method, user-agent and timestamp to stats message


pawelmhm:
https://github.com/scrapinghub/splash/issues/262

kmike:
Thanks! There is more to improve in Splash logging, but I'm not closing #262 for now.

Issue #226, opened 2015-06-01
HTTP error responses should be JSON


kmike:
I think error responses should be JSON instead of plaintext. This way it'll be easier for clients to parse error responses automatically and decide which error happened.

kmike:
Fixed by #275 and #307.

PR #275, opened 2015-08-17
Show errors in the requested format

Youwotma:
render.html -> HTML error
render.png or jpeg -> Image error
All other endpoints -> JSON error (#226)

### DIFF
--- a/splash/resources.py
+++ b/splash/resources.py
@@ -37,8 +37,19 @@ class _ValidatingResource(Resource):
         try:
             return Resource.render(self, request)
         except BadOption as e:
-            request.setResponseCode(400)
-            return str(e) + "\n"
+            return self._writeError(request, 400, str(e))
+
+    def _writeErrorContent(self, request, code, content, content_type='text/plain'):
+        request.setHeader("content-type", content_type)
+        request.setResponseCode(code)
+        request.write(content)
+
+    def _writeError(self, request, code, message):
+        """Can be overridden by subclasses format errors differently"""
+        return self._writeErrorContent(request, code, json.dumps({
+            "error": code,
+            "message": message,
+        }), content_type="application/json")
 
 
 class BaseRenderResource(_ValidatingResource):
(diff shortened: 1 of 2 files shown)


immerrr:
The original issue AFAIU was about parsing and finding out what's happened. I don't think that presenting the error in image format is better than plaintext in that regard. Or is it?

pawelmhm:
I remember there was this suggestion from @kmike https://github.com/scrapinghub/splash/issues/226

Youwotma:
My thoughts to show an image instead of text was indeed to make it easy to
see what happened in case somebody uses 
Or saves the result in a file without checking the response code and
displays it somewhere.

Of course api consumers should always check the responses status code, but
if you forget or don't bother to do it the browser will only display a
broken image with no other clues of what went wrong.

immerrr:
@pawelmhm yes, @kmike's suggestion was about the fact that json is better than plaintext, which i can relate to.
@Youwotma i'd wildly guess that splash is used programmatically more often than visually and it's quite tricky to show images in, for example, spider log. Besides, images embedded as `` sometimes result in http errors too, and I cannot remember from the top of my head anyone serving a pretty "404 Not Found" image instead of a plain HTTP error.

kmike:
You both have a point, but for backwards compatibility it is better to keep png/jpeg error responses simple by default.

In /execute endpoint one can get a screenshot or html of a page even if it returns an error. What about adding an option to do it in render.xxx endpoints, something like `ignore_errors=1`? If this option is active the result will have the same status code, but splash will still render it and return page's html or screenshot. It may also allow to say "I want to render as much as possible in 30s and get a result; I don't care if it is incomplete".

kmike:
@Youwotma could you please add a test which will check the error result explicitly, i.e. that it is a JSON response with 'error' and 'message' fields?

Youwotma:
Yep, will do it tomorrow

kmike:
Thanks!

PR #276, opened 2015-08-18
POST method for outgoing Splash requests

pawelmhm:
it adds support for passing http_method argument to browser tab go() method. 

This allows to pass http_method and body in both HTTP API and in lua scripting. In Lua this can be done with `splash:go(url, http_method=method_name, body=raw_body)`. I also added optional argument `formdata` for `splash:go()`, it can be Lua table with form data to post. 

Does it make sense to add formdata param to HTTP API? 

Theoretically ATM uses could also probably do POST with files, but I'm not really sure how this would work, I'm not sure if it is worth checking that, maybe that could be some extra feature for later.


pawelmhm:
ok will fix tests tomorrow

splash/browser_tab.py
```diff
@@ -797,6 +804,11 @@ def request_obj(self, url, headers=None):
             self.web_page.skip_custom_headers = True
             self._set_request_headers(request, headers)
 
+        if (body and not headers) or (body and not request.hasRawHeader("content-type")):
```
kmike: Why is first condition needed?

splash/browser_tab.py
```diff
@@ -827,11 +839,21 @@ def get(self, url, callback, headers=None, follow_redirects=True):
...
                       headers=None):
-        # XXX: The caller must ensure self._delete_reply is called in a callback.
-        if method != 'GET':
-            raise NotImplementedError()
-        request = self.request_obj(url, headers=headers)
-        reply = self.network_manager.get(request)
+        # XXX: The caller must ensure self._delete_reply is called in a callback
+        request = self.request_obj(url, headers=headers, body=body)
+
+        operation = getattr(self.network_manager, method.lower(), None)
```
kmike: I think we should be explicit here, i.e. no use getattr. `method` can be validated somewhere earlier, but if not we'll allow to call any network manager method here.

splash/qtrender_lua.py
```diff
@@ -343,10 +344,26 @@ def error(error_info):
...
+            body = self.lua.lua2python(formdata)
+            if isinstance(body, dict):
+                body = urlencode(body)
+            else:
+                raise ScriptError("formdata argument for go() must be Lua table")
+
+        elif body:
+            body = self.lua.lua2python(body)
+            if not isinstance(body, basestring):
+                raise ScriptError("request body must be string")
```
kmike: I think we should raise an error if both formdata and body are passed

splash/qtrender_lua.py
```diff
@@ -343,10 +344,26 @@ def error(error_info):
         ))
 
     @command(async=True)
-    def go(self, url, baseurl=None, headers=None):
+    def go(self, url, baseurl=None, headers=None, http_method="GET", body=None, formdata=None):
+
         if url is None:
             raise ScriptError("'url' is required for splash:go")
 
+        if http_method not in OPERATION_QT_CONSTANTS:
```
kmike: let's be more permissive and allow both lower-case and upper-case methods

splash/qtrender_lua.py
```diff
@@ -343,10 +344,26 @@ def error(error_info):
...
+            raise ScriptError("incorrect HTTP method: {}".format(http_method))
+
+        if formdata:
+            body = self.lua.lua2python(formdata)
+            if isinstance(body, dict):
+                body = urlencode(body)
+            else:
+                raise ScriptError("formdata argument for go() must be Lua table")
```
kmike: What if user passes `{'foo', 'bar', 'baz'}`? It is still a Lua table, is it accepted?

kmike: Also, shouldn't we add max_depth argument to lua2python here? Arbitrary nesting shouldn't be supported for form data, it is better to fail earlier.

splash/render_options.py
```diff
@@ -138,7 +140,10 @@ def get_quality(self):
         return self.get("quality", defaults.JPEG_QUALITY, type=int, range=(0, 100))
 
     def get_http_method(self):
-        return self.get("http_method", "GET")
+        method = self.get("http_method", "GET")
+        if method not in OPERATION_QT_CONSTANTS:
```
kmike: allow lower-case?

splash/browser_tab.py
pawelmhm: good point, not needed, removed this line

splash/browser_tab.py
pawelmhm: so you think we should raise exception here if method is invalid? 

We do allow any valid network manager with this code. We do not allow invalid network_manager methods, like 'GETX' for example. If we get invalid method we default to GET.

splash/qtrender_lua.py
pawelmhm: ok

splash/qtrender_lua.py
pawelmhm: yeah good idea about max_depth, I think max_depth=3 should be ok? this is what we use when converting headers

pawelmhm: > What if user passes {'foo', 'bar', 'baz'}? It is still a Lua table, is it accepted?

this will be converted to {1: "foo", 2:"bar", 3:"baz"} probably not what user will expect, but should not fail in splash I suppose. you think we should check for cases like this? Can be difficult to detect

splash/qtrender_lua.py
pawelmhm: ok added this

pawelmhm:
should we add option to have formdata in HTTP api here? probably worth having in api too what do you think @kmike

splash/browser_tab.py
kmike: right, but there are many more NAM methods than just get/post/.., we shouldn't allow user to call them

pawelmhm: ok, updated

splash/browser_tab.py
```diff
@@ -827,11 +839,21 @@ def get(self, url, callback, headers=None, follow_redirects=True):
...
-            raise NotImplementedError()
-        request = self.request_obj(url, headers=headers)
-        reply = self.network_manager.get(request)
+        # XXX: The caller must ensure self._delete_reply is called in a callback
+        request = self.request_obj(url, headers=headers, body=body)
+
+        if method.upper() not in OPERATION_QT_CONSTANTS:
+            raise ValueError("Invalid HTTP method: '{}'".format(method))
+
+        operation = getattr(self.network_manager, method.lower())
```
kmike: I think this is not correct: there is no .options method, and a method for DELETE is called `deleteResource`, not 'delete`. Maybe we can use [sendCustomRequest](http://doc.qt.io/qt-5/qnetworkaccessmanager.html#sendCustomRequest) instead?

splash/browser_tab.py
```diff
@@ -287,6 +289,11 @@ def go(self, url, callback, errback, baseurl=None, http_method='GET',
         """
         self.store_har_timing("_onStarted")
 
+        if body and http_method not in ["POST", "PUT"]:
+            # if someone gives body but forgets method
+            # QT will raise exceptions
+            http_method = "POST"
```
munro: Wouldn't it be better to throw the exception if the programmer made a mistake, instead of guessing?  Also it looks like it's doing that below with `ValueError("Request with method {} should not have body".format(method))`

pawelmhm: yeah you're probably right, not needing to pass "method" everytime could sometimes be a "feature" (you can only pass body and it will default to POST) but on the other hand it can make some programmer mistakes difficult to spot - someone will pass wrong method something could go wrong but he will not know why. 

So will raise 400 with bad method here.

splash/browser_tab.py
pawelmhm: I'm not sure about this method, docs say:

> This method provides means to send verbs other than the common ones provided via get() or post() etc., for instance sending an HTTP OPTIONS command.

so I guess this would be for some uncommon HTTP like OPTIONS.

We could use [createRequest method of network manager](http://doc.qt.io/qt-5/qnetworkaccessmanager.html#createRequest) but then again I wonder if we really should add support for all methods in this one PR.

I initially thought that it could be useful and not difficult to support all verbs, but now I see there are many edges cases here. Aside from one you mentioned seems like DELETE methods [may have a body](http://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request).

What do you think about limiting this PR to adding POST only? I would refactor this PR and limit allowed HTTP verbs to POST and GET. We could create new PR for handling all other HTTP methods, this way we could also add separate unit tests for all other methods and we could look into all edge cases related to other HTTP methods.

kmike: I'm fine with limiting this PR to POST requests, it is an useful addition.

splash/browser_tab.py
```diff
@@ -829,11 +834,18 @@ def get(self, url, callback, headers=None, follow_redirects=True):
...
+        # XXX: The caller must ensure self._delete_reply is called in a callback
+        request = self.request_obj(url, headers=headers, body=body)
+        method = method.upper()
+
+        if method not in ["POST", "GET"]:
             raise NotImplementedError()
-        request = self.request_obj(url, headers=headers)
-        reply = self.network_manager.get(request)
+
+        if body and method == "POST":
```
kmike: is it possible to send a POST request with empty body?

pawelmhm: hmm I assumed current master code is right when it assumes we can't have POST without body:https://github.com/scrapinghub/splash/blob/master/splash/browser_tab.py#L405 but now I checked this and it seems like this is actually allowed, PyQt does not complain when you pass body=None and it performs POST request okay.

Perhaps in some cases POST without body will be user error (user forgetting to pass body), but given that it is not forbidden by HTTP standard and allowed by PyQt we should probably allow it.

kmike: Yeah, I think we should allow to send a POST requests with empty ("") body.

splash/browser_tab.py
```diff
@@ -537,6 +537,13 @@ def http_get(self, url, callback, headers=None, follow_redirects=True):
             follow_redirects=follow_redirects
         )
 
+    def http_post(self, url, callback, headers=None, follow_redirects=True, body=None):
```
kmike: What happens if `follow_redirects` is True (default) and a response for our POST request is a redirect? We shouldn't send the same POST response again to the new location, it should become a GET request. See https://github.com/scrapy/scrapy/blob/181f8d27d782cb433de5f78770ee1446c62991cd/scrapy/downloadermiddlewares/redirect.py#L54

splash/qtrender_lua.py
```diff
@@ -473,24 +499,46 @@ def errback(msg, raise_):
...
+    @command(async=True)
+    def http_post(self, url, headers=None, follow_redirects=True, body=None):
+        """
+
+        :param url: string with url to fetch
+        :param headers: dict, if None {"content-type": "application/x-www-form-urlencoded"} will be added later
+        :param follow_redirects: boolean
+        :param body: string with body to be sent in request
+        :return: AysncBrowserCommand http_post
+        """
```
kmike: Could you please add docs to https://github.com/scrapinghub/splash/blob/master/docs/scripting-ref.rst?

pawelmhm: ok docs added

splash/browser_tab.py
pawelmhm: ah, you're right, but this would require us to add some redirect middleware to Splash?

kmike: No, I think we should change _on_request_finished implementation or maybe change the code calling _on_request_finished.

pawelmhm: seems like redirects after POST are already handled by PyQt if you use `go()` but indeed if we use `http_post` they are not handled, and request method after redirect is still POST not GET

pawelmhm: fixed in recent commit

docs/scripting-ref.rst
```diff
@@ -832,6 +837,53 @@ To load a webpage to the browser use :ref:`splash-go`.
 
 .. _HAR response: http://www.softwareishard.com/blog/har-12-spec/#response
 
+.. _splash-http-post:
+
+splash:http_post
+---------------
```
kmike: underline is too short

splash/browser_tab.py
```diff
@@ -528,6 +528,13 @@ def http_get(self, url, callback, headers=None, follow_redirects=True):
...
+        self.http_client.post(url,
+                              callback=callback,
+                              headers=headers,
+                              follow_redirects=True,
```
kmike: why is there `follow_redirects` argument if it is not passed to self.http_client?

splash/qtrender_lua.py
```diff
@@ -415,16 +416,39 @@ def error(error_info):
...
+
+        elif formdata:
+            body = self.lua.lua2python(formdata, max_depth=3)
+            if isinstance(body, dict):
+                body = urlencode(body)
+            else:
+                raise ScriptError("formdata argument for go() must be Lua table")
+
+        elif body:
+            body = self.lua.lua2python(body, max_depth=3)
```
kmike: why is lua2python call needed here?

splash/qtrender_lua.py
```diff
@@ -415,16 +416,39 @@ def error(error_info):
...
+
         if url is None:
             raise ScriptError({
                 "argument": "url",
                 "message": "'url' is required for splash:go",
             })
 
+        http_method = http_method.upper()
+        if http_method not in ["POST", "GET"]:
+            raise ScriptError("Unsupported HTTP method {}".format(http_method))
```
kmike: Sorry, I changed the way errors are reported in master. The way to go now is to instantiate ScriptError with a dict (check other methods for examples). Could you please update this?

splash/browser_tab.py
pawelmhm: good catch

splash/qtrender_lua.py
pawelmhm: well, someone could pass anything as body, if user passes Lua table or function or something else unexpected as body, it could cause trouble somewhere later, so I thought it's better raise explicit error here as soon as we know something is wrong. But to know what type we have we need to convert from lua to python.

kmike: Yeah, but the only supported type is a string, and if I'm not mistaken lua2python is not needed to get a Python string from a Lua string because they are converted automatically. So you can check for a string and discard all other types without trying to convert them.

splash/qtrender_lua.py
pawelmhm: ok I updated this in qrender lua and in render_options

splash/qtrender_lua.py
pawelmhm: :+1:

kesavkolla:
Please merge this branch.  It's a quite useful feature.  Specially form submissions on many sites are POST requests.

splash/tests/test_jsonpost.py
```diff
@@ -157,6 +157,36 @@ def test_cookie(self):
         self.assertStatusCode(r, 200)
         self.assertIn("bar", r.text)
 
+    def test_http_POST_request_from_splash(self):
+        formbody = {"param1": "one", "param2": "two"}
+
+        r = self.request({
+            "url": self.mockurl("postrequest"),
+            "formdata": formbody,
+            "body": urllib.urlencode(formbody),
```
kmike: Why are both formdata and body passed here? If formdata is supported it could make sense to raise an error, if it is not supported removing it will make the test more clear.

kmike:
@kesavkolla yeah, this PR is going to merged before next release.

splash/tests/test_jsonpost.py
pawelmhm: good point, fixed

kmike:
Thanks @pawelmhm!

PR #277, opened 2015-08-19
Disable localStorage

Youwotma:
A lot of websites use localStorage like this:

```
if(window.localStorage) {
    localStorage.completely_useless_persistent_information = '...'
} else {
    // Use alternative storage method (cookies, server side) or don't
    // store data at all
}
loadTheRestOfTheApplication();
```
At the moment in splash localStorage exists and is an object but trying
to set a property will raise an exception and prevent the application to
load. This disables localStorage (window.localStorage is null) and will
make code like the above work


kmike:
It makes sense, thanks!

PR #280, opened 2015-08-24
Add support for specifying the proxy URL directly.

### LINKED ISSUE #160: How can i crawler a url use a custom proxy

Youwotma:
Yes, I saw the other PR.

This allows passing either an URL or a profile name as the proxy parameter

Fixes  #160

### DIFF
--- a/splash/proxy.py
+++ b/splash/proxy.py
@@ -11,6 +11,7 @@ an HTTP proxy (see :mod:`splash.proxy_server`).
 from __future__ import absolute_import
 import re
 import os
+import urlparse
 import ConfigParser
 
 from PyQt4.QtNetwork import QNetworkProxy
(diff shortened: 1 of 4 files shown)


splash/proxy.py
```diff
@@ -160,6 +161,50 @@ def _parseIni(self, ini_path):
...
+    If GET parameter is a fully qualified URL, use the specified proxy.
+    The syntax to specify the proxy is:
+    [protocol://][user:password@]proxyhost[:port])
+
+    Where protocol is either ``http`` or ``socks5``. If port is not specified,
+    it's assumed to be 1080.
+    """
+    def __init__(self, proxy):
+        url = urlparse.urlparse(proxy)
+        print proxy, url, url.scheme, url.hostname
```
kmike: stray print?

Youwotma: Yeah, good eye. I have amended the commit

kmike:
Thanks!

PR #287, opened 2015-08-28
Add splash:get_version()

Youwotma:


docs/scripting-ref.rst
```diff
@@ -1822,6 +1822,30 @@ Example 3 - extract all cookies set by website without reading response body
...
     end
 
+.. _splash-version:
+
+splash:get_version
+------------------
+
+Get Splash mayor and minor version.
+
+**Signature:** ``splash:get_version()``
```
kmike: major, minor = splash:get_version()

kmike:
I think 'major' is common spelling. Otherwise, +1 to merge.

Youwotma:
Right, commit amended!

kmike:
Hey @Youwotma,

What do you think about returning a Lua table instead of 2 values? We can extend table with more information: major/minor/point, "dev" suffix, qt version, python version.

PR #260, opened 2015-08-03
[Don't merge] Qt5 port

kmike:
A PR with qt5/pyqt5 Splash port. It is opened to make it easier to review changes, and to run Travis tests for qt5 branch automatically.

PR #288, opened 2015-09-01
Qt 5 + Python 3

kmike:
This is a follow-up to https://github.com/scrapinghub/splash/pull/260 and https://github.com/scrapinghub/splash/pull/251.


kmike:
This PR contains qt5-based Splash version compatible with Splash 1.8. It won't be updated.

PR #311, opened 2015-10-01
Add a --disable-private-mode command line flag

Youwotma:


kmike:
It makes sense, thanks.

PR #314, opened 2015-10-13
Fix double-encoding of URL Paths

Youwotma:
This is a bug only present on the qt4 version, this change tries to
emulate what seems to be the default behaviour in qt5.

### DIFF
--- a/splash/qtutils.py
+++ b/splash/qtutils.py
@@ -8,7 +8,6 @@ import itertools
 import sys
 import time
 
-
 from PyQt4.QtCore import (QAbstractEventDispatcher, QDateTime, QObject,
                           QRegExp, QString, QUrl, QVariant)
 from PyQt4.QtGui import QApplication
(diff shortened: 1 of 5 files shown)


splash/qtutils.py
```diff
@@ -141,9 +140,19 @@ def qurl2ascii(url):
...
 
 
+def to_qurl(s):
+    if isinstance(s, QUrl):
+        return s
+
+    if isinstance(s, unicode):
+        s = s.encode('utf-8')
+
+    return QUrl.fromEncoded(s)
```
kmike: [Docs](http://doc.qt.io/qt-5/qurl.html#fromEncoded) say "Parses input and returns the corresponding QUrl. input is assumed to be in encoded form, containing only ASCII characters.", but we're not escaping non-ASCII chars. Is it an issue with Qt documentation, or a feature they don't officially support?

Youwotma: Not sure, my guess is documentation error. I tested different ways of
converting to a QUrl and this is the only one that did the right thing
every time.

Youwotma: Examining the source here:
http://cep.xray.aps.anl.gov/software/qt4-x11-4.8.6-browser/d4/dd4/qurl_8cpp_source.html#l04443
it manipulates a char array and percent-encode any non-safe characters, so it will work if you pass a utf-8 encoded char array.

kmike:
Thanks!

Issue #238, opened 2015-06-09
Does Splash handle javascript set cookies?


mappelgren:
I'm interested in building a scraper that goes through a site and finds what cookies are being set both 1st party and 3rd party. Scrapy's built in cookie middleware only deals with cookies set in http headers (set-cookies). I'm interested in cookies set through other methods, for example javascript. Therefore I wonder, how does Splash/scrapyjs handle cookies set through other means than http headers? Does it set javascript cookies or handle them in any way? Or are they just ignored?

Youwotma:
I just checked and they don't get saved or reported in the har file, so pretty much ignored.

mappelgren:
Ok, thanks! Thats too bad, do you think it would be possible to modify something to make them at least acknowledge that they are being created?

Youwotma:
As a workarround you could [inject a script](https://splash.readthedocs.org/en/latest/api.html#execute-javascript) to read the contents of document.cookie.
I don't know what endpoint you are using, but for example you could save it as an attribute of the body element with this script.
```
document.body.setAttribute("data-cookie", document.cookie);
```

mappelgren:
Ok, so document.cookie would contain the cookies but they wouldn't be set or forwarded with requests. That makes sense. Thanks a lot for your help!

kmike:
Thanks for the discussion! See also: https://github.com/scrapinghub/splash/issues/135. We need to fix it somehow.

lucaswxp:
Hey guys.

For me, document.cookie appears to be always empty, try this request, for instance:
```sh
curl -i -X POST -H 'content-type: application/json' -d '{"js_source": "document.body.setAttribute(\"data-cookie\", document.cookie);", "wait": 3, "url": "http://www.americanas.com.br"}'  http://localhost:8050/render.html
```
Then `body[data-cookie]` will be empty.

But if I do a raw curl request, I get the cookies set back:
```sh
curl -I http://www.americanas.com.br
> ...
Set-Cookie: MobileOptOut=1; path=/; domain=americanas.com.br
Set-Cookie: b2wChannel=ACOM; path=/; domain=americanas.com.br
Set-Cookie: akaau=1438897818~id=b874481c11a6576d5e992cfba8510912; path=/
```

Any guidance in how do I get all cookies from the response within splash?

herlon214:
:+1:

Youwotma:
[...]

Issue #135, opened 2014-12-22
cookies modified from JS code


kmike:
We should check and document what happens when cookies are modified in JS code, but no further requests are sent.

PR #316, opened 2015-10-15
JS cookie support ( #238, #135)

Youwotma:
JS cookies are sent in the http headers, and header cookies are visible
to JavaScript.

This doesn't fix the cookie sharing problem between different sessions.

### DIFF
--- a/splash/network_manager.py
+++ b/splash/network_manager.py
@@ -195,11 +195,12 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
             request.setRawHeader(name, value)
 
     def _handle_request_cookies(self, request):
-        jar = QNetworkCookieJar()
-        self.setCookieJar(jar)
         cookiejar = self._get_webpage_attribute(request, "cookiejar")
         if cookiejar is not None:
             cookiejar.update_cookie_header(request)
+            self.setCookieJar(cookiejar)
+        else:
+            self.setCookieJar(QNetworkCookieJar())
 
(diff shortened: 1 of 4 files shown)


splash/network_manager.py
```diff
@@ -195,11 +195,12 @@ def _handle_custom_headers(self, request):
             request.setRawHeader(name, value)
 
     def _handle_request_cookies(self, request):
-        jar = QNetworkCookieJar()
-        self.setCookieJar(jar)
         cookiejar = self._get_webpage_attribute(request, "cookiejar")
         if cookiejar is not None:
             cookiejar.update_cookie_header(request)
+            self.setCookieJar(cookiejar)
```
kmike: Hmm, who owns/deletes cookiejars? 

A cookiejar is created in SplashQWebPage. Then it get re-parented to our network manager (according to setCookieJar docs). But we may have several web pages alive (one per slot), but only a single network access manager. When you set a new cookiejar using setCookieJar old cookiejar is deleted (see https://github.com/qtproject/qtbase/blob/bafbf3b872103409370fbda43a2ce3ad153a33f4/src/network/access/qnetworkaccessmanager.cpp#L677). I wonder if it may cause https://github.com/scrapinghub/splash/issues/321.

Youwotma:
I got the same error toady too, so it probably it's causing it. You can revert this PR if you want.

kmike:
@Youwotma I've reverted the fix for now (until we find a better solution), but kept the tests.

PR #336, opened 2015-11-10
cleanup splash:wait_for_resume

kmike:
this seems to work fine in qt5

### DIFF
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -620,13 +620,6 @@ class BrowserTab(QObject):
         one argument. The first argument will be an object with `resume()`
         and `error()` methods. The code _must_ call one of these functions
         before the timeout or else it will be canceled.
-
-        Note: this cleans up the JavaScript global variable that it creates,
-        but QT seems to notice when a JS GV is deleted and it destroys the
-        underlying C++ object. Therefore, we can only delete the JS GV _after_
-        the user's code has called us back. This should change in QT5, since
-        it will then be possible to specify a different object ownership
-        policy when calling addToJavaScriptWindowObject().
         """
 
         frame = self.web_page.mainFrame()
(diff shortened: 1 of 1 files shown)


splash/browser_tab.py
```diff
@@ -620,13 +620,6 @@ def wait_for_resume(self, js_source, callback, errback, timeout):
         one argument. The first argument will be an object with `resume()`
         and `error()` methods. The code _must_ call one of these functions
         before the timeout or else it will be canceled.
-
-        Note: this cleans up the JavaScript global variable that it creates,
-        but QT seems to notice when a JS GV is deleted and it destroys the
-        underlying C++ object. Therefore, we can only delete the JS GV _after_
-        the user's code has called us back. This should change in QT5, since
-        it will then be possible to specify a different object ownership
-        policy when calling addToJavaScriptWindowObject().
```
kmike: @mehaase - do you recall details about this issue? I can't reproduce it in qt5.5.1, but I wonder if that's because the issue is fixed or because examples are not complex enough to trigger the bug. Did Splash testing suite segfault, or did you have other segfaulting examples back then?

mehaase: Hmm, I think I was actually talking about PyQT. The PyQT4 `addToJavaScriptWindowObject()` method doesn't have [the ownership argument](https://github.com/Werkov/PyQt4/blob/master/sip/QtWebKit/qwebframe.sip). That argument was [added in PyQT5](https://github.com/baoboa/pyqt5/blob/1e1d8a3ba677ef3e47b916b8a5b9c281d0f8e4b5/sip/QtWebKitWidgets/qwebframe.sip).

As I recall, it was pretty easy to trigger. I triggered it a lot (maybe every time?) with just the basic scripts (in the old headless horseman repo) I was using to evaluate the feature.

kmike:
I've tried it with examples from https://github.com/TeamHG-Memex/scrash-lua-examples and it was not crashing => merge, but keep an eye on it.

Issue #203, opened 2015-04-12
disable browser/webkit caching ?


nwohaibi:
Hi,
Thanks for the wonderful work on Spalsh

I just wanted to know if there is any way to disable browser caching of files?
Or maybe return all HTTP requests made in har/log/entries, not just the ones with 200 http status ?

thanks in advance

kmike:
Hi @nwohaibi,

Thanks!

> I just wanted to know if there is any way to disable browser caching of files?

There is a way to do it in QWebKit (see http://doc.qt.io/qt-4.8/qnetworkrequest.html#CacheLoadControl-enum), but currently this option is not exposed by Splash. It is a good feature to have, but we need to design a public API for it and implement it.

> Or maybe return all HTTP requests made in har/log/entries, not just the ones with 200 http status ?

HAR entries already contain all HTTP requests, not just the ones with 200 http status code. In case of cache some records may be missing because they are not requested at all. It should be possible to add them to the output as well, but I haven't checked the details; implementation may be not so straightforward.

nwohaibi:
Thanks for taking the time to clarify :)
Since I already have Splash in production, i might tackle the issue by modifying cache-control headers in HTTP responses. This way, WebKit would assume all resources are not to be cached.  
let me know if I can be of any help
and thanks again

starrify:
Hi @kmike 

> There is a way to do it in QWebKit (see http://doc.qt.io/qt-4.8/qnetworkrequest.html#CacheLoadControl-enum), but currently this option is not exposed by Splash.

I used to believe that, and I even tried to make a PR that way. However later I realized that it is not the case. (Proved by local testings)

The `QNetworkRequest::CacheLoadControl` attribute shall be set for request instances, and it is Qt's network manager to decide whether to use a disk cache. However in the current implement of `splash`, caching in the network managers is not enabled at all (please check https://github.com/scrapinghub/splash/blob/master/splash/network_manager.py#L42)

As WebKit also has its own in-memory cache (for scripts, stylesheets, images, etc.), that is believed to be the real cause. In some specific scenarios it's required to strictly disable any kind of caching. Thus I made PR #339 for this.

nirvana-msu:
I think I'm facing a similar issue, though I'm having trouble tracing it to the exact route cause. I use `scrapy-splash` (`aquarium`, in fact), and I need to scrape two URLs - both identical apart from URL fragment. So the two URLs are similar to:
http://example.com/#1
http://example.com/#2
[...]

PR #339, opened 2015-11-19
added: New API endpoint `_cache_control`. Solves #203

starrify:


kmike:
Hey @starrify,

The feature makes sense, but I think it is not good to allow to set these options via HTTP API: they should be set at startup like other global Splash options.

nramirezuy:
hey @starrify, was this working properly before? Because right now it seems to be ignoring `QSettings`.

Issue #344, opened 2015-11-24
splash not setting proper proxy authentication for requests


pawelmhm:
It seems like Splash is not setting proper "authentication" headers for requests. I have following Lua script

```lua
function main (splash)
    local host = 'proxy_url'
    local port = 8010
    local user = splash.args.proxy_user
    local password = splash.args.proxy_user
    splash:on_request(function (request)
        request:set_proxy{host, port, username=user, password=password}
    end)
    splash:go(splash.args.url)
    return splash.har()
end
```

when I submit this script to Splash I get 407 from my proxy even though proxy params are correct. When debugging this I see that har log does not have basic_auth header in request headers, and this is the reason why my proxy responds with 407. 

There was related issue #302 but mostly about POST requests.

pawelmhm:
after doing some more research I found out that

1) order of statements in Lua matters:

```lua
    splash:on_request(function (request)
        request:set_proxy{host, port, username=user, password=password}
    end)
    splash:go(splash.args.url)
``` 

fails, but

```lua
splash:go(splash.args.url)    
splash:on_request(function (request)
        request:set_proxy{host, port, username=user, password=password}
    end)
``` 
works.

2) QT Network Access Manager emits signal if proxy rejects credentials: http://doc.qt.io/qt-4.8/qnetworkaccessmanager.html#proxyAuthenticationRequired 

3) QT Network Access Manager caches credentials.

> QNetworkAccessManager will cache the credentials internally. The next time the proxy requests authentication, QNetworkAccessManager will automatically send the same credential without emitting the proxyAuthenticationRequired signal again.

I'm not yet sure what it means for situations when there are multiple requests in splash using different proxies? How this cache is stored? Is it is stored per proxy host? What if there is same proxy host and but it has different usernames and passwords?

From my experience it seems that after first request accepted by proxy others are accepted too. This could mean that network access manager is using cached credentials for further requests with same proxy without checking them. 

I wonder if setting proxy per request is actually good idea, perhaps we should consider moving setProxy to splash object instance instead of to on_request callback? docs say

> setProxy() Sets the proxy to be used in future requests to be proxy. This does not affect requests that have already been sent. The proxyAuthenticationRequired() signal will be emitted if the proxy requests authentication.

so it seems that setProxy should be set per network access manager and not per request.

Youwotma:
I had this problem too, solved it by setting the 'Proxy-Authentication' header manually

pawelmhm:
[...]

PR #346, opened 2015-11-25
[#344] set proxy credentials on proxyAuthenticationRequired signal

pawelmhm:
seems like this signal is emitted the very first time user is trying to access proxy.
Subsequent request to same proxy dont need filling out credentials. Qt docs say that
credentials are cached. It seems like they are cached also between Splash restarts.

### DIFF
--- a/splash/network_manager.py
+++ b/splash/network_manager.py
@@ -64,10 +64,10 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
         super(ProxiedQNetworkAccessManager, self).__init__()
         self.sslErrors.connect(self._on_ssl_errors)
         self.finished.connect(self._on_finished)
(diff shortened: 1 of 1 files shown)


pawelmhm:
docs for this signal are here, http://doc.qt.io/qt-4.8/qnetworkaccessmanager.html#proxyAuthenticationRequired

> This signal is emitted whenever a proxy requests authentication and QNetworkAccessManager cannot find a valid, cached credential. The slot connected to this signal should fill in the credentials for the proxy proxy in the authenticator object.

> QNetworkAccessManager will cache the credentials internally. The next time the proxy requests authentication, QNetworkAccessManager will automatically send the same credential without emitting the proxyAuthenticationRequired signal again.

if we're ok with this change I'll add some unit tests for it

PR #347, opened 2015-12-01
Private mode

pawelmhm:
Allow users to enable/disable private_mode from Lua Scripts with a call to 

```lua
splash.private_mode_enabled = false/true -- similar to splash.js_enabled
```

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -35,6 +35,19 @@ Enable or disable execution of JavaSript code embedded in the page.
 
 JavaScript execution is enabled by default.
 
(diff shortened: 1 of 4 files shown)


kmike:
Hey @pawelmhm,

:+1: 

Could you please add docs?

kmike:
An argument to regular HTTP API would be also nice, but it can be done in a separate PR if you're not up to implementing it now.

pawelmhm:
thanks for review @kmike ! I updated docs, if we could add HTTP API changes in other PR it would be cool

PR #348, opened 2015-12-02
fix proxy authentication #344

pawelmhm:
this just adds test for referenced issue

### DIFF
--- a/splash/network_manager.py
+++ b/splash/network_manager.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+import base64
 import itertools
 import functools
 from datetime import datetime
@@ -176,6 +177,11 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
         # proxies set in on_request
         if hasattr(request, 'custom_proxy'):
             self.setProxy(request.custom_proxy)
+            user, password = request.custom_proxy.user(), request.custom_proxy.password()
+            if not user and not password:
+                return
+            auth = b"Basic " + base64.b64encode("{}:{}".format(user, password).encode("utf-8"))
+            request.setRawHeader(b"Proxy-Authorization", auth)
 
     def _handle_custom_headers(self, request):
         if self._get_webpage_attribute(request, "skip_custom_headers"):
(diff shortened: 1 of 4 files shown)


kmike:
I've definitely seen proxy auth working. Do you have an idea when it works, so that we can add a test for that as well?

pawelmhm:
> I've definitely seen proxy auth working. 

sure I've seen it working too. It worked pretty well. I need to check since when it started to malfunction, I'm guessing that some changes in Qt5 perhaps? Have you seen proxy auth working under Qt5?

pawelmhm:
@kmike I added fix for missing authentication headers, I went with manually setting auth headers in createRequest in NAT. 

Alternative approach is setting authentication on ProxyAuthenticationRequired as in  here: https://github.com/scrapinghub/splash/pull/346 but as @Youwotma noted this way we can't deal with invalid proxy credentials being passed. If there are invalid credentials passed in Signal QT will just keep emitting signal again and again, filling up retrying request all the time with bad credentials. We probably dont want that. 

Drawback of current solution is that it handles only Basic authentication and not other forms of proxy-auth, but at least it does handle the most common form of authentication. 

TODO list would involve adding tests for proxy authentication in HTTP API, I'm not sure if this works ok now, but if we can do review and testing of this on other PR that would be cool.

splash/network_manager.py
```diff
@@ -177,6 +178,13 @@ def _handle_custom_proxies(self, request):
         if hasattr(request, 'custom_proxy'):
             self.setProxy(request.custom_proxy)
 
+            user, password = request.custom_proxy.user(), request.custom_proxy.password()
+            if not user and password:
```
Youwotma: Note that this is evaluated as `if (not user) and password:` so if there is no user and no password still sets the header

pawelmhm: stupid typo on my side sorry

pawelmhm: some proxies allow setting just user with blank password, so I think we should make check for "user" only.

Youwotma: I think having and empty user is also allowed, we should return only if there is no user and no password (`if not user and not password`)

pawelmhm: hmm are you sure it is allowed to have blank user and non-blank password? how proxy can determine password validity without user string? Given bit of confusion around this perhaps safest option is just base64 encoding "user:passowrd" and then setting header if result of b64 is nonempty string.

pawelmhm: or maybe 

```python 

if not "".join((user, password)):
    return
```
this will make sure there is either user or password

kmike:
Thanks @pawelmhm and @Youwotma!

Issue #232, opened 2015-06-02
add /_ping endpoint


kmike:
It should return "ok", the current max memory usage and maybe some other lightweight stats (no heavy things like gc.collect). Load balancers can use `/_ping` to check if a Splash instance is alive; memory usage info is helpful if load balancer itself wants to stop Splash based on its RAM usage.

kfei:
:+1:

kmike:
Fixed by #350.

PR #350, opened 2015-12-11
Add ping endpoint

rasmusx:
Hello,

Here is MR for adding a "/_ping" endpoint that returns OK and maxrss.

Aims to solve issue: #232

### DIFF
--- a/splash/resources.py
+++ b/splash/resources.py
@@ -347,6 +347,18 @@ class ClearCachesResource(Resource):
         }).encode('utf-8')
 
 
+class PingResource(Resource):
+    isLeaf = True
+
+    def render_GET(self, request):
+        request.setHeader(b"content-type", b"application/json")
+        return json.dumps({
+            "status": "ok",
+            "maxrss": resource.getrusage(resource.RUSAGE_SELF).ru_maxrss,
+        })
+
+
+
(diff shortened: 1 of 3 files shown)


splash/resources.py
```diff
@@ -347,6 +347,18 @@ def render_POST(self, request):
...
 
 
+class PingResource(Resource):
+    isLeaf = True
+
+    def render_GET(self, request):
+        request.setHeader(b"content-type", b"application/json")
+        return json.dumps({
+            "status": "ok",
+            "maxrss": resource.getrusage(resource.RUSAGE_SELF).ru_maxrss,
```
kmike: This has a problem - on OS X RUSAGE_SELF is in bytes, but on Linux it is in KBs. Check https://github.com/scrapinghub/splash/blob/3c9970ac96096a444ae9eb345d9670adc72bfbef/splash/utils.py#L114.

kmike:
Nice, thanks! Could you please add tests and docs?

rasmusx:
Thanks for pointing out my mistake.

I updated the code. Max RSS should be fixed now (always in bytes). I also updated docs and added a test case. Let me know if this correct/enough.

Thanks :)

kmike:
Thanks @rasmusx!

PR #357, opened 2015-12-23
[resources/logging] log args as JSON, log message on errors

pawelmhm:
follow up on #274 args message should be proper JSON, commit also adds logging of parsed url representation so that it should be easy to extract netloc and other url components for example using logstash.

### DIFF
--- a/splash/resources.py
+++ b/splash/resources.py
@@ -198,9 +198,15 @@ class BaseRenderResource(_ValidatingResource):
             "timestamp": int(time.time()),
             "user-agent": (request.getHeader(b"user-agent").decode('utf-8')
                            if request.getHeader(b"user-agent") else None),
-            "args": repr(options)
+            "args": options
         }
-        log.msg(json.dumps(msg), system="events")
+        try:
+            msg = json.dumps(msg)
+        except TypeError:
+            # splash-as-proxy options contain byte objects which are not
+            # serializable
+            pass
+        log.msg(msg, system="events")
 
     def _on_timeout_error(self, failure, request, timeout=None):
         failure.trap(defer.CancelledError)
--- a/splash/resources.py
+++ b/splash/resources.py
@@ -47,16 +47,24 @@ class _ValidatingResource(Resource):
         try:
             return Resource.render(self, request)
         except BadOption as e:
-            return self._write_error(request, 400, e)
+            self._write_error(request, 400, e)
+            self._log_stats(request, {}, error=self._format_error(400, e))
+            return b"\n"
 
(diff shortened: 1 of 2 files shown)


splash/resources.py
```diff
@@ -201,9 +201,27 @@ def _log_stats(self, request, options):
...
+    def detailed_options(self, options):
+        url = options.get("url")
+        if not url:
+            return options
+        url = six.moves.urllib.parse.urlparse(url)
+        options.update({
+            "netloc": url.netloc,
+            "query": url.query,
+            "path": url.path
+        })
```
kmike: I think we shouldn't parse URLs here - Splash script users may pass anything in 'url' argument, it doesn't have a special meaning. Is itmust be possible to write these filters in Kibana instead?

pawelmhm: actually parsing url in kibana could be tricky, i can do some research though. Alternative apprach would be: If you think users could pass anything maybe we could just do some try - except, and try parsing url only if it is valid url?

redapple: I agree with @kmike , `url` could be anything. I would rather do the extra parsing outside of Splash, in log analyzers, like Logstash for example, which can be extended to do URL parsing, or may have that already.

I believe leaving this PR's purpose of making parsing of arguments easier as proper JSON is already a big win, without analyzing them fully at this stage.

pawelmhm: @redapple @kmike ok I updated PR

pawelmhm:
hey @kmike do you think we should also log Splash status code and client ip in "events"?

pawelmhm:
6e20f66eba93f63888ea573a74b adds proper JSON logging of errors

pawelmhm:
ping @kmike do you think its ok?

splash/resources.py
```diff
@@ -201,9 +203,19 @@ def _log_stats(self, request, options):
...
         }
-        log.msg(json.dumps(msg), system="events")
+        if error:
+            msg["error"] = error
+        try:
+            msg = json.dumps(msg)
+        except TypeError:
+            # splash-as-proxy options may contain byte objects which are not
+            # serializable
+            pass
```
kmike: splash-as-a-proxy feature is dropped - is this still required?

kmike:
Yeah, I think this feature is good :+1:

splash/resources.py
pawelmhm: good point will check that and update PR

pawelmhm:
rebased against recent master (without splash-as-proxy) and ammended recent commit to log error in consistent format (same format that goes to response will be logged) and without ugly try-except

PR #358, opened 2015-12-26
backwards-incompatible: protection for splash:evaljs, splash:jsfunc and splash:wait_for_resume results

### LINKED ISSUE #343: evaljs with jquery causes Splash to crash
I have following innocently looking Lua script using evaljs:

```lua
function main(splash)
    splash:go("http://www.ica.se/handla/ica-nara-ovrells-id_01739/")
    splash:set_viewport_full()
    splash:wait(1)
    splash:evaljs("$('#search').val('cola');")
    splash:evaljs("e = $.Event('keypress'); e.which = 13; $('#search').trigger(e)")
    splash:wait(1)
    return splash:jpeg()
end
```

this crashes Splash every time I run it. I found warning in docs about eval js

> Don’t use splash:evaljs when the result is not needed - it is inefficient and could lead to problems; use splash:runjs instead. For example, the following innocent-looking code (using jQuery) may fail A gotcha is that to allow chaining jQuery $ function returns a huge object, splash:evaljs tries to serialize it and convert to Lua. It is a waste of resources, and it could trigger internal protection measures; splash:runjs doesn’t have this problem.

it's nice we warn users about evaljs but I think we should do something about it, it seems bad that evaljs can actually crash your splash server. What internal protection measure do we have now? Is there a way to improve those protection measures to avoid problem like this?

kmike:
A PR to fix #343.

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -45,7 +45,7 @@ Enable or disable browser's private mode (incognito mode).
 **Signature:** ``splash.private_mode_enabled = true/false``
 
 Private mode is enabled by default unless you pass flag ``--disable-private-mode`` at Splash startup.
-Note that if you disable private mode browsing data such as cookies or items kept in local 
+Note that if you disable private mode browsing data such as cookies or items kept in local
 storage may persist between requests.
 
 .. _splash-resource-timeout:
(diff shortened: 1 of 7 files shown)


docs/scripting-ref.rst
```diff
@@ -346,30 +346,47 @@ a table to an Array use :ref:`treat-as-array`.
...
+================  =================
 
 .. note::
 
     The rule of thumb: if an argument or a return value can be serialized
     via JSON, then it is fine.
 
+Note that currently you can't return DOM Elements, JQuery $ results and
+similar structures from JavaScript to Lua; to pass data you have to
+extract their attributes of interest as plain strings/numbers/objects/arrays:
```
pawelmhm: so we can't return circular objects and functions from js calls now, is this the only backward incompatible change added by this PR?

kmike: Yeah, Splash now doesn't return circular objects; instead of circular object detection there is now a depth limit. I went with depth limit because JS engine in Webkit doesn't have a Map object which can be used for detection of duplicate object references. Even https://github.com/douglascrockford/JSON-js/blob/master/json2.js uses O(N) lookups for cycle detection, which makes serialization O(N^2), I can't stand it :)

Another backwards-incompatible change is that previously Splash tried to serialize e.g. DOM elements with all their properties; it worked for a single element on a simple page, but didn't work if there are many of them. Now it fails fast and only supports explicit objects/arrays. 

It was not possible to return functions even before this PR - they were returned just as empty tables which is not useful. 

Another backwards-incompatible change is that splash:evaljs error info is slightly changed ('type' value is upper-cased).

PR #359, opened 2015-12-27
Correctly handling unsupported content

starrify:
For now I cannot think of a good design for returning the downloaded unsupported content back to the client via existing endpoints (e.g. `/html`, `/png`, and `/har`). The good news:
1. The server won't hang and return a 504 timeout error now.
2. At least the `/har` endpoint now works as expected.

Maybe later we can introduce something like `splash:raw_content` in the scripting support to deliver the content.


splash/browser_tab.py
```diff
@@ -512,6 +540,28 @@ def _on_url_changed(self, url):
...
+            # original loadFinished signal emits. In such cases we do not want
+            # the same signal twice.
+            if not self._unsupported_content_reply.error():
+                self.web_page.mainFrame().loadFinished.emit(True)
+            else:
+                self.web_page.mainFrame().loadFinished.emit(False)
+
+    def _on_unsupported_content(self, reply):
+        self.logger.log('Unsupported content detected', min_level=3)
+        self._is_unsupported_content = True
```
pawelmhm: this attribute is set per browser tab and browser tab can be used to download multiple resources, if you set this attribute here will it affect other requests going through same webpage? e.g. script doing

```lua
splash:go("url_with_unsupported_content") # this will set _is_unsupported_content to True
splash:go("url with supported content") # will this go through?
```

other thing worth noting is that there could be multiple resources downloaded when rendering so e.g.

```lua
splash:go("some page")
```
may issue 10 requests, first 6 will have supported content, but 7th will have unsupported content, what will happen with resources 7-10? will they be downloaded all right?

splash/tests/test_render.py
```diff
@@ -208,6 +208,21 @@ def test_invalid_wait(self):
                                   'wait': wait})
                 self.assertStatusCode(r, 400)
 
+        def test_unsupported_content(self):
+            cases = [
+                # Short body (Can be received together with the headers)
+                ("raw-bytes?length=16", 200),
+                # Short body with error
+                ("raw-bytes?length=16&claim_length=32&abort=1", 502),
```
pawelmhm: is this something webkit cannot handle? why? is it because of different content-length and actual length of response?

I read in docs that 
> This signal is emitted when WebKit cannot handle a link the user navigated to or a web server's response includes a "Content-Disposition" header with the 'attachment' directive. 

so perhaps also worth adding test for content-disposition?

I also wonder about 502 status code, this is Bad Gateway so if there is some load balancer in front of splash, e.g. nginx, user may confuse 502 from splash with 502 from some load balancer. Is 502 best choice for status code? Are there any alternatives?

PR #360, opened 2016-01-07
Several improvements to the web interface

Youwotma:
Several improvements to the web interface:
- Autocomplete `splash:` methods
- Display several images and complex structures in the response
- Allow returning directly splash.png/jpeg()
- Load harviewer lazyly and allow seeing the rest of the response if it doesn't load
- Allow downloading big string as a file (e.g: `splash:html()`)

Looking for feedback and testing

Autocomplete:
splash

New response view:
Default script result
splash

Return array
splash

Return object
splash

Return string
splash

Return long string (`splash:html()`)
splash

Return har
splash

Return image
splash

Structure
splash

### DIFF
--- a/splash/qtrender.py
+++ b/splash/qtrender.py
@@ -152,7 +152,8 @@ class DefaultRenderScript(RenderScript):
 
     def _runjs(self, js_source, js_profile):
         js_output, js_console_output = None, None
-        if js_source:
+
+        if js_source or js_profile:
             if self.console:
                 self.tab._jsconsole_enable()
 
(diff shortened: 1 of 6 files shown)


Youwotma:
@kmike can you have a look?

kmike:
@Youwotma :+1: 

Do you think it is possible to enforce/hardcode the PNG -> HAR -> HTML order in the result?

kmike:
It seems you don't need it, but FYI: inspection data for Splash Notebook autocompletion is here: https://github.com/scrapinghub/splash/blob/master/splash/kernel/inspections/splash-auto.json

Youwotma:
@kmike Nice! didn't know about that. That's better that what is using because some aliases that are implemented in lua are not autocompleted.

And maybe I can add inline docs in the popup like in http://ternjs.net/doc/demo/#simple

I'll update the PR with both changes

Youwotma:
I made the changes, the auto-completion now looks like this: http://i.imgur.com/J6WOLMV.png

kmike:
Awesome, thanks!

PR #366, opened 2016-01-19
Make strings explicitly unicode

Youwotma:
Several Travis python2 test were failing (https://travis-ci.org/scrapinghub/splash/jobs/103174284#L1967) and if using splash with python2 you would get UnicodeDecodeErrors when trying to run JavaScript with Unicode characters.


kmike:
Thanks! Too bad 2.x tests hang on Travis.

Issue #286, opened 2015-08-27
Create a set of reference/example scripts


kmike:
@pablohoffman's ideas (which I agree with):

1. Create a folder with example scripts for common tasks and move many example scripts from the docs there;
2. include these scripts in docs;
3. show these examples in Splash UI.

Another idea is to create a list of Jupyter notebooks; they render nicely in github.

pawelmhm:
we have bunch of Lua scripts in docs already, for starters we could just copy paste them from docs to example folder. If we store scripts in two places (docs and /examples folder) they can get out of sync, someone will update in one place but forget about other, is it ok or should we keep them in sync somehow?

kmike:
Yeah, I think we should put 'canonical' version to 'examples' folder and include them in Sphinx docs. Neither duplicating nor removing examples from docs look good to me.

PR #368, opened 2016-01-20
Add examples UI and some examples (#286)

Youwotma:
I have added some examples and UI to load the examples as part of issue #286.

 Feel free to give feedback about the UI, the examples and suggestions for other examples.

screensoot


kmike:
Awesome!

* What do you think about clearing URL field when an example is selected? For empty URLs UI adds 'http://', I think it should be fixed.
* It may worth it to try reformat the examples because most of them don't fit in the editor window

kmike:
screen shot 2016-01-20 at 3 22 36 pm

kmike:
I think the original idea of #286 was to move examples from http://splash.readthedocs.org/en/stable/scripting-tutorial.html and http://splash.readthedocs.org/en/stable/scripting-ref.html to files, make them available as standalone scripts and then include to docs (without copy-pasting). The problem is that these examples often require HTTP arguments - this is a good style (so it is nice to keep following it in docs), but the UI only allows to pass 'url' as an argument.

There are some examples which still can be extracted:

* http://splash.readthedocs.org/en/latest/scripting-ref.html#splash-call-later
* http://splash.readthedocs.org/en/latest/scripting-ref.html#splash-png - render.png example implementation
* examples from http://splash.readthedocs.org/en/latest/scripting-ref.html#splash-on-request;
* examples from http://splash.readthedocs.org/en/latest/scripting-ref.html#splash-on-response-headers

kmike:
hm, I think the UI for examples won't work if Splash is installed via pip

Youwotma:
I have extracted every example with a main method as a separate file, but I haven't added them to the UI because because a lot of them are not very visual or need extra parameters.

I have also reformatted the ones that are included in the UI so they fit in the editor.

About the pip problem, how do you think we should fix it? if we add the examples folder to `package_data` in `setup.py` is enough? (I don't know much about setuptools)

kmike:
Hey @Youwotma,

As you said, many of these examples are not very visual and require extra parameters, and are not helpful without the context; I think we shouldn't extract all code snippets to files, sorry for not saying it earlier!

I think yes, adding examples to package_data should be enough, but we may have to switch from building file paths using `os.path.join()` to pkg_resources, and figure out how to access examples folder which is not in a Python package.

Youwotma:
@kmike OK, so what should the criteria be for extracting it be? Extract all that work/make sense in the demo UI and show all of them in the demo UI? We could group the examples menu in different categories/submenues if there are too many.

kmike:
@Youwotma yeah, I like your criteria.

If we are to group examples in the UI it could make sense to put them in separate folders under examples, and group based on that.

Youwotma:
@kmike Re-based and extracted only some of the examples.

I have also added the examples to package_data and MANIFEST.in, for the examples to be packaged correctly when doing a binary distribution (setup.py bdist) I had to move them inside splash folder.

I have confirmed that they end up in the expected location (both with sdist and bdist).

kmike:
Too bad we have to move examples/ folder, it was nice to have it top-level. 

`splash-haproxy.conf` is linked in docs (faq.rst), we should update the link. This change probably will break Splash 1.8 docs as well.

Would you mind including ui/ static files to sdist/bdist as well?

Youwotma:
@kmike done! I couldn't find a way of doing it without moving it, but I don't know much about setuptools, maybe there is a way.

kmike:
Awesome, thanks!

PR #375, opened 2016-01-29
[browser_tab] set UA for requests not going via web page

### LINKED ISSUE #290: User agent not set correctly when specifying a base url
Test script:
```
function main(splash)
  local url = 'http://httpbin.org/headers?nocache=' .. math.random()
  splash:set_user_agent('FooBar/1.0')
  assert(splash:go(url, 'http://httpbin.org'))
  assert(splash:wait(0.5))
  return {
    html = splash:html(),
    png = splash:png(),
    har = splash:har(),
  }
end
```
Uses `Mozilla/5.0` instead of `FooBar/1.0`.

Specifying the user agent as a header works.

pawelmhm:
before this commit requests not issued with QWebPage (e.g. from Lua with go and base_url,
or with http_get, http_post or with all endpoints but base url argument) were not getting user agent. After this commit all requests issued by Splash will have UA set, regardless of where they are issued from (either web page or browser tab-http_client).

fixes #290


splash/tests/test_execute.py
```diff
@@ -2504,6 +2519,66 @@ def test_get(self):
...
+        headers = resp.json()
+        self.assertNotEqual(len(headers), 0)
+        self.assertEqual(headers["user-agent"], "CUSTOM UA")
+
+    def test_get_with_custom_ua_in_headers(self):
+        resp = self.request_lua("""
+        function main(splash)
+            response = assert(splash:http_get{splash.args.url, headers={["user-agent"]="Value 1"}})
+            return response.request.headers
+        end
```
kmike: For completeness, could you please add a test with user agent set both with `set_user_agent` and `headers`?

pawelmhm: added tests in recent commit, I noticed one unusual behavior relating to UA, noted in 4f82d19aa694c6fe1b46d5b7df4d3ddd3f995d19

kmike: Hm, I think that's fine for splash:go - if you go to a page using some user-agent you expect to download all related resources using the same user-agent. This is documented [here](http://splash.readthedocs.org/en/stable/api.html#arg-headers) and [here](http://splash.readthedocs.org/en/stable/scripting-ref.html#splash-go) (at the bottom). 

But this behaviour looks incorrect for `splash:http_get`.

pawelmhm: yeah you're right, I'll see if we can fix that. Also maybe it would be good to add some tests for those outcoming requests issued when rendering? dont see any tests checking if UA is set correctly for them.

kmike: +1. If there are such tests then I think we can remove a warning in splash:go docs.

pawelmhm: I added tests for UA for requests used when rendering. Turns out UA was not set when there was 'headers' argument passed because of some minor bug comparing string with bytes. Now only need to differentiate behavior of headers argument regarding go() and http_get

pawelmhm: BTW do we still need to support Python 2? I'm running tests in 3 and it works fine, tests for 2 are marked as "can fail" in tox, so I guess 3 is primary environment now, right?

pawelmhm: > https://github.com/scrapinghub/splash/pull/375/files#r51277538

is fixed by ad8591a4def12ff

splash/browser_tab.py
```diff
@@ -299,6 +301,16 @@ def go(self, url, callback, errback, baseurl=None, http_method='GET',
         if body is not None:
             body = to_bytes(body)
 
+        if headers is not None:
+            # headers can be list or dict, keys can be bytes or str, HTTP specs
+            # say headers are case-insensitive, so they can be uppercase or lowercase
+            if not isinstance(headers, dict):
+                headers = dict(headers)
+            headers = {to_bytes(k.lower()): to_bytes(v) for k, v in headers.items()}
```
kmike: Wouldn't this make all headers lower-case? While HTTP specs says headers are case-insensitive we shouldn't change case of headers if user passes them.

splash/browser_tab.py
```diff
@@ -892,6 +904,12 @@ def _send_request(self, url, callback, method='GET', body=None,
 
         request = self.request_obj(url, headers=headers, body=body)
 
+        if headers and "user-agent" in [k.lower() for k in headers.keys()]:
+            user_agent = headers.get("user-agent", headers.get("User-Agent"))
+        else:
+            user_agent = self.web_page.userAgentForUrl(to_qurl(url))
+        request.setRawHeader(b"user-agent", to_bytes(user_agent))
```
kmike: Content-Type header is set in self.request_obj; is there a reason User-Agent header is not set there as well?

splash/browser_tab.py
```diff
@@ -946,9 +964,8 @@ def _set_request_headers(self, request, headers):
             headers = headers.items()
 
         for name, value in headers or []:
-            request.setRawHeader(to_bytes(name), to_bytes(value))
-            if name.lower() == 'user-agent':
-                self.set_user_agent(value)
+            name, value = to_bytes(name), to_bytes(value)
+            request.setRawHeader(name, value)
```
kmike: `request.setRawHeader(to_bytes(name), to_bytes(value))` was not worse :) It had a tiny advantage of not mutating name and value.

splash/browser_tab.py
pawelmhm: splash go() has two flows when creating new requests. When you pass baseurl request is downloaded by splash http client. If you dont pass baseurl it is downloaded with web page. request_obj is used in both of these scenarios. But User agent is missing **only** if you pass base_url. If user-agent would be set in request_obj it would be set also for request downloads going via webpage which is not necessary. This is why it is set here.

splash/browser_tab.py
pawelmhm: > Wouldn't this make all headers lower-case?

yeah it will make all headers case insensitive. I thought that it would be easier for us to work with headers in consistent case, it is irritating to account for case when fetching headers from dictionary. 

But you're right that perhaps users care about case they pass so we should not change it, so I will update this.

splash/browser_tab.py
pawelmhm: > It had a tiny advantage of not mutating name and value.

why do you think it mutates name & value here?

pawelmhm: ah you mean changing values we iterate on? yeah it's tiny edit and probably not needed so will revert it

splash/browser_tab.py
```diff
@@ -892,6 +907,12 @@ def _send_request(self, url, callback, method='GET', body=None,
 
         request = self.request_obj(url, headers=headers, body=body)
 
+        # setting UA for request that is not downloaded via webpage.mainFrame().load_to_mainframe()
+        ua_from_headers = lowercase_byte_dict(headers).get(b"user-agent")
+        web_page_ua = self.web_page.userAgentForUrl(to_qurl(url))
+        user_agent = ua_from_headers if ua_from_headers else web_page_ua
+        request.setRawHeader(b"user-agent", to_bytes(user_agent))
```
pawelmhm: ammeded previous commit @kmike 

* not converting values to bytes
* not copying headers twice
* removed copy() call from utils function

kmike:
Thanks @pawelmhm!

Issue #390, opened 2016-02-25
Seems setup.py miss a little code.


genius001:
There will be failed to run splash-jupyter when perform "jupyter notebook --no-browser --port=8888 --ip=*"

The error info displayed:
no found treat.lua or json.lua.

If I added line "lua_modules/libs/*.lua" to the file setup.py, and reinstalled splash, the problem will gone. 

So I thought it maybe a little mistake in setup.py.

Youwotma:
@genius001 Thanks for catching that!

PR #392, opened 2016-02-25
Add lua_modules/libs/*.lua to setup.py (#390)

Youwotma:
Thanks @genius001 for reporting this


kmike:
Thanks @Youwotma @genius001! I've created 2.0.1 release; docker images should be updated in about an hour.

Issue #397, opened 2016-02-29
splash segfaults if you pass filters but dont pass "url" argument


pawelmhm:
Really interesting bug, if you want to use filters but dont pass "url" and try to make call to go() splash will segfault, following Python script reproduces the problem

```python
base_url = "http://localhost:8050/execute"
post_url = base_url + "?filters=easylist"
lua_source = "function main(splash) return splash:go('http://splash.readthedocs.com') end"
res = requests.post(post_url, data=json.dumps({"lua_source": lua_source}), headers={"content-type": "application/json"})
print(res.json())
```

This ends with SIGSEGV (Address boundary error)

``` 
> python3 ./bin/splash -v 3 --filters-path=dev/filters/                                                                                                           
2016-02-29 16:01:10.502342 [pool] [140131975982720] SLOT 0 is starting
2016-02-29 16:01:10.502430 [render] [140131975982720] function main(splash) return splash:go('http://splash.readthedocs.com') end
2016-02-29 16:01:10.505161 [render] [140131975982720] [lua_runner] dispatch cmd_id=__START__, args=None
2016-02-29 16:01:10.505255 [render] [140131975982720] [lua_runner] arguments are for command __START__, waiting for result of __START__
2016-02-29 16:01:10.505335 [render] [140131975982720] [lua_runner] send None
2016-02-29 16:01:10.505496 [render] [140131975982720] [lua_runner] got AsyncBrowserCommand(id=None, name='go', kwargs={'http_method': 'GET', 'baseurl': None, 'errback': '<an errback>', 'headers': None, 'url': 'http://splash.readthedocs.com', 'callback': '<a callback>', 'body': None})
2016-02-29 16:01:10.505586 [render] [140131975982720] [lua_runner] instructions used: 62
2016-02-29 16:01:10.505673 [render] [140131975982720] [lua_runner] executing AsyncBrowserCommand(id=0, name='go', kwargs={'http_method': 'GET', 'baseurl': None, 'errback': '<an errback>', 'headers': None, 'url': 'http://splash.readthedocs.com', 'callback': '<a callback>', 'body': None})
2016-02-29 16:01:10.505749 [render] [140131975982720] HAR event: _onStarted
2016-02-29 16:01:10.505872 [render] [140131975982720] callback 0 is connected to loadFinished
2016-02-29 16:01:10.507348 [network] [140131975982720] GET http://splash.readthedocs.com/

[...]

PR #402, opened 2016-03-03
request/response middleware fixes

kmike:
A fix for https://github.com/scrapinghub/splash/issues/397.

### DIFF
--- a/splash/network_manager.py
+++ b/splash/network_manager.py
@@ -332,12 +332,9 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
                 # should we raise errors here?
                 # https://github.com/scrapinghub/splash/issues/161
                 self.log("error in %s callback" % event_name, min_level=1)
-                # Log the traceback. It requires escaping because
-                # self.log uses str.format.
-                tb = traceback.format_exc().replace('{', '{{').replace('}', '}}')
-                self.log(tb, min_level=1)
+                self.log(traceback.format_exc(), min_level=1, format_msg=False)
 
-    def log(self, msg, reply=None, min_level=2):
+    def log(self, msg, reply=None, min_level=2, format_msg=True):
         if self.verbosity < min_level:
             return
 
@@ -348,7 +345,8 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
             if not url:
                 return
 
-        msg = msg.format(url=url)
+        if format_msg:
+            msg = msg.format(url=url)
         log.msg(msg, system='network-manager')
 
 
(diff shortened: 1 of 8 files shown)


pawelmhm:
do you think we should also add sys.excepthook for unhandled QT crashes? I noticed that some QT crashes go completely unnoticed.

Something like this:

```python
def excepthook(etype, evalue, tb):
    from twisted.python import log
    tb = "".join(traceback.format_exception(etype, evalue, tb))
    msg = "\n\nUnhandled QT failure \n{}".format(tb)
    log.msg(msg)
    sys.__excepthook__(etype, evalue, tb)
```

in main() as 

```
def main():
    sys.excepthook = excepthook
```
will produce something like this if you return None from NAT:

```
	Unhandled QT crash 
	TypeError: invalid result from SplashQNetworkAccessManager.createRequest(), NoneType cannot be converted to PyQt5.QtNetwork.QNetworkReply in this context
	
```
i was also trying to figure out if there is some way to catch SIGSEGV-s in Python and perhaps get some hint where they occur to make debugging easier, but cannot find any good way

splash/network_manager.py
```diff
@@ -372,14 +370,23 @@ def run_response_middlewares(self):
...
-            for middleware in self.response_middlewares:
-                middleware.process(reply, render_options)
+            try:
+                for middleware in self.response_middlewares:
+                    middleware.process(reply, render_options)
+            except:
+                self.log("internal error in response middleware", min_level=1)
+                self.log(traceback.format_exc(), min_level=1, format_msg=False)
 
     def createRequest(self, operation, request, outgoingData=None):
```
pawelmhm: perhaps we should add docstring saying that you cannot return anything else but QNetworkReply from here, if anything else is returned Splash will always crash. I tested this with returning None and it died. It is easy to forget about this if you're used to Python. Usually there is no harm in doing return None somewhere but here it is very harmfful.

splash/network_manager.py
```diff
@@ -332,12 +332,9 @@ def _run_webpage_callbacks(self, request, event_name, *args):
                 # should we raise errors here?
                 # https://github.com/scrapinghub/splash/issues/161
                 self.log("error in %s callback" % event_name, min_level=1)
-                # Log the traceback. It requires escaping because
-                # self.log uses str.format.
-                tb = traceback.format_exc().replace('{', '{{').replace('}', '}}')
-                self.log(tb, min_level=1)
+                self.log(traceback.format_exc(), min_level=1, format_msg=False)
```
pawelmhm: does it make sense to log exception as json somehow? so that log parsers could easily catch them? I think currently they can easily go unnoticed in production setup

splash/network_manager.py
kmike: Yeah, makes sense.

kmike:
I'm not sure excepthook helps with segfaults; [faulthandler](https://docs.python.org/3.4/library/faulthandler.html) looks more promising.

splash/network_manager.py
kmike: Tracebacks will be much less readable as json; you can make log parsers parse Python stacktraces, but you can't make json stacktraces readable for humans.

kmike:
But if enabling exception hooks can help with qt exceptions then +1 to do that.

pawelmhm:
> I'm not sure excepthook helps with segfaults; faulthandler looks more promising.

ah nice suggestion from initial tests I see it works pretty nice, will create PR with this enabled

kmike:
I haven't added a test for the original issue (filters + /execute) because it is not documented that 'filters'  argument has a special meaning for `/execute`.

PR #405, opened 2016-03-05
Implement render region param

### LINKED ISSUE #213: allow to get PNG snapshots of arbitrary page areas
We should provide a way to render only a specified rectangular part of a page.

immerrr:
This PR should ultimately fix #213.

It adds support for `splash:png{region={left, top, right, bottom}}` (and `splash:jpeg`, too) where the coordinates are specified relative to the viewport (so you must still use `render_all` to be able to specify absolute coordinates within a page). The rect includes `[left, right)` and `[top, bottom)` pixel ranges. Scaling & trimming should apply to the region too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
 
 script:
     - py.test --version
-    - py.test --cov=splash --doctest-modules -n $SPLASH_BUILD_PARALLEL_JOBS --duration=50 splash
+    - py.test --cov=splash --doctest-modules --duration=50 splash
 
 after_success:
     - codecov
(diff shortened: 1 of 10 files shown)


immerrr:
As of now, it does lack a few tests and mentions nothing in the documentation. I don't have much more time to devote to it (unfortunately I almost ran out of time trying to get it to run on my 12.04) so I hope @kmike you can help me with that.

kmike:
Awesome, thanks @immerrr!

immerrr:
Great stuff, @kmike !

PR #416, opened 2016-03-14
Fix encoding issue in download button (Fix  #412)

### LINKED ISSUE #412: HTML download is incorrect in Splash HTTP UI
1. Render this URL http://forums.pdfforge.org/categories/help using default script in Splash UI;
2. Click 'download' button near 'html'.

For me it returns a 4K document, while it should be 60K+. HTML code in the textarea is correct.

Firefox 44.0.2, OS X.

Youwotma:


kmike:
Thanks @Youwotma!

Issue #380, opened 2016-02-08
add splash:select() method


pawelmhm:
qt web frame has nice selector interface, it could be used to select items in html page source from Lua script, web frame select method takes css selector as argument and returns web QWebElement: http://doc.qt.io/qt-5/qwebframe.html#findFirstElement this could be good alternative to running javascript scripts within Splash

mike1808:
hi @pawelmhm 
I want to implement this functionality but I see the main problem here. What methods should the returned element have? 
Let assume we got our element ```element = splash:select('.container')``` What's next? What we'd be able to do with it? In #213 you've suggested to call `#png` method on it. What other methods do you think will be good to have?

pawelmhm:
hey @mike1808 this issue has become part of Google Summer of Code proposal http://gsoc2016.scrapinghub.com/ideas/#splash-ideas the idea is to have helpers similar to the ones provided by Scrapy, Selenium, PhantomJS/CasperJS, I think the call for applications is still open there. 

>  I assume that it should have some basic utils methods like click, scrollTo, getHtml, etc.

yes I was thinking about something like this, click(), hover(), scroll() would be all very useful things to have IMO.

Implementation could either use QT API calls or it could rely on JavaScript executed in QT. Second option may be better as it doesn't tie us to any specific QT api (we may change engine implementation #349 so api calls could change too). 

Currently there is a way to define your own helpers in javascript and convert them to lua callable via splash jsfunc, so those helpers could perhaps work in some similar way.

genius001:
This link was broken

[http://doc.qt.io/qt-5/qwebframe.html#findFirstElement](url)

mike1808:
@genius001 QWebFrame was removed from Qt5, the working link is [QWebFrame::findFirstElement](http://doc.qt.io/qt-4.8/qwebframe.html#findFirstElement)

genius001:
@mike1808 

Have you developing functionality splash:select()?

I also need a feature similar splash:select() .
Referer https://github.com/scrapinghub/splash/issues/433

mike1808:
@genius001 I've only started to work on it.

genius001:
Excellent. @mike1808 

I will try to implement some functionality e.g. splash:mouse_move splash:move_click splash:move_hover.

kmike:
Fixed in #490, merged in #531.

PR #422, opened 2016-03-20
feat(scripting): add Element class with `offset` method

mike1808:
I've tried to implement _Element_ class suggested in #380 
HTML element can be retrieved using _#select_ method on the _splash_ object. 

```lua
local element = splash:select('.container .title')
```

In this PR I've implemented only 1 method for this class: _#offset_. It returns a dictionary with top and left offset value like in jQuery#offset method.

There is no tests and docs update. Only one script example. I'll write them if my assumptions what'd be done is right.

Is this what #380 have suggested or I'm going to the wrong direction?  

P.S. This PR was creating for my proposal for GSoC 2016.

### DIFF
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -833,6 +833,17 @@ class BrowserTab(QObject):
 
         return res
 
+    def select(self, selector):
+        """ Find element by its selector """
+        self.logger.log("finding element using `%s` selector" % selector, min_level=3)
+        element = self.web_page.mainFrame().findFirstElement(selector)
+
+        # TODO: change error handling, it will raise 400 error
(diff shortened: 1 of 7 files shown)


kmike:
Hey @mike1808,

The PR looks fine for what it does, but I'm reclucant to merge it without a clear vision on API we're going to provide. You're asking right questions in https://github.com/scrapinghub/splash/issues/380#issuecomment-198931640, but there are more of them, e.g.:

* what to do in cases where a selector returns multiple elements?
* what if an user has a DOM element selected in a JS function - how to convert it to Element?
* is it possible to pass Element back to JavaScript code?

PR #424, opened 2016-03-22
Add FAQ entry on how to disable private mode

redapple:
Triggered by this StackOverflow question: [Can Scrapy be used with the Chrome Browser?](https://stackoverflow.com/questions/36145443/can-scrapy-be-used-with-the-chrome-browser)

> (...) but I was really hoping to use ScrapyJS/Splash for this project. However, it looks like the Scrapy/Splash combination suffers from the website's private browsing wall.

I believe an entry in the FAQ would help getting this option known.


kmike:
+1 to add it to FAQ. I think it also makes sense to document consequences: cookies may become shared, localStorage starts working.

PR #426, opened 2016-03-24
Update browser_tab.py

genius001:
Line:915
Change "user-agent" to "User-Agent"


Youwotma:
Hi @genius001 thanks for the pull request! There are some test failing because they have "user-agent" instead of "User-Agent" hardcoded in them, would you mind fixing that as well?

genius001:
Hi, @Youwotma 

I noticed if change all 'user-agent' to 'User-Agent' in tests/test_execute.py that will cause another problem, And test will failing.

The error log [https://travis-ci.org/scrapinghub/splash/jobs/118187194#L2146](url)

I could not figure out how to resolve it.
If I fix all these headers key(cache-control, accept-encoding,accept-language) to case sensitive.
That could be affect other test file?

But this time I thought whether is it necessary to do this or keep original?

If it is necessary, workaround?

PR #427, opened 2016-03-24
Update HARViewer, add favicon

Youwotma:
Unfortunately if the old harviewer.js file is cached by the browser if will now fail to load.


redapple:
@Youwotma , one can add a note in the release notes about this and how to remove the cached one

Youwotma:
@redapple I couldn't figure how to remove it from cache without whipping the whole cache, `Ctrl+F5` didn't work, and disabling cache in devtools loaded the new version but didn't update the cache and didn't work again when re-enabling caching (tested in firefox only)

kmike:
@Youwotma I think we can change the URL to prevent caching; maybe even adding ?version=2 will work. If not (e.g. because harviewer loads js files dynamically) we can change /_ui/harviewer path to something else.

Youwotma:
@kmike I could not add an url parameter because the files are loaded dynamically by require.js, I have appended the version to the harviewer path

Issue #432, opened 2016-03-29
baseurl documentation unclear


joorei:
I don’t quite understand how the baseurl argument should be used.

> The base url to render the page with.
> If given, base HTML content will be feched from the URL given in the url argument, and render using this as the base url.

My impression was it will render all relative links on the page as absolute links based on the value given for baseurl. But if I test it using the following request (the target page is just a random example) links like href="impressum.html" will still be relative and thus pointing (incorrectly) to http://localhost:8050/impressum.html if the response is shown in a browser.
`http://localhost:8050/render.html?url=http%3A%2F%2Fwww.gesetze-im-internet.de&baseurl=http%3A%2F%2Fwww.gesetze-im-internet.de`
Using a simple javascript I can add a base element to the page and fix this, but what exactly is baseurl for? Am I using it wrong or did I misunderstood its purpose?

kmike:
@joorei with `baseurl` argument you can render a page located at one URL as if it is located at another URL. For example, you can host a copy of page HTML on your server, but use baseurl of the original page. This way Splash will resolve relative URLs as relative to original page URL, so that you can get e.g. a proper screenshot or execute proper JavaScript code.

Note that by passing `baseurl` you're instructing *Splash* to use it, not *your* browser. It doesn't change relative links to absolute in DOM, it makes *Splash* to treat them as relative to baseurl when rendering. 

Changing links to absolute in DOM tree is not what browsers do when base url is applied - e.g. if you check href attribute using JS code it will still contain relative value even if `<base>` tag is used.  `render.html` returns DOM snapshot, so the links are not changed. When you load `render.html` result in a browser it is your browser who resolves relative links, not Splash, so they are resolved incorrectly.

joorei:
@kmike Thank you for your detailed answer, this certainly clears things up. :thumbsup:

kmike:
@joorei in addition to your baseurl docs improvement there is now also a FAQ entry (http://splash.readthedocs.org/en/latest/faq.html#render-html-result-looks-broken-in-a-browser), so I'm closing this. Thanks!

PR #434, opened 2016-03-31
Make baseurl description in api.rst more verbose

joorei:
Following #432 I tried to make the description for baseurl a bit more precise.


kmike:
Thanks @joorei!

PR #438, opened 2016-04-05
adds splash:click() method

pawelmhm:
~~for now it gets css selector and does click on this element~~ updated to work with x, y coordinates

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -2102,3 +2102,141 @@ Example:
              error("Splash 1.8 or newer required")
          end
      end
+
+.. _splash-mouse-click:
+
+splash:mouse_click
+------------------
+
+Trigger mouse click event in web page.
+
+**Signature:** ``splash:mouse_click(x, y)``
+
+**Parameters:**
+
+* x - integer or float with x position of element to be clicked
+* y - integer or float with y position of element to be clicked
+
+**Returns:** nil
+
+**Async:**: no.
+
+
+.. note::
+
+    Coordinates for mouse events must be relative to web document. Element on which action is performed must be inside viewport (must be visible
+    to the user). If element is outside viewport and user needs to scroll to see it, you must either scroll
+    down with splash or simply set viewport to full with :ref:`splash-set-viewport-full`.
+
+.. note::
+
+    Mouse events are not propagated immediately, to see consequences of click  reflected in page source you must call :ref:`splash-wait`
+
+Example:
+
+.. code-block:: lua
+
+    -- get button element dimensions with javascript and perform mouse click
+
+    function main(splash)
+        assert(splash:go(splash.args.url))
+        get_dimensions = splash:jsfunc([[
+            function () {
+                rect = document.getElementById('button').getBoundingClientRect();
+                return {"x":rect.left, "y": rect.top}
+            }
+        ]])
+        dimensions = get_dimensions()
+        splash:mouse_click(dimensions.x, dimensions.y)
+        -- wait split second to allow event to propagate
+        splash:wait(0.1)
+        return splash:html()
+    end
+
+Example 2:
+
+.. code-block:: lua
+
+    -- click on element outside viewport
+
+    function main(splash)
+        get_dimensions = splash:jsfunc([[
+            function () {
+                rect = document.getElementById('must_scroll_to_see').getBoundingClientRect();
+                return {"x":rect.left, "y": rect.top}
+            }
+        ]])
+        assert(splash:go(splash.args.url))
+        splash:set_viewport_full()
+        dimensions = get_dimensions()
+        splash:mouse_click(dimensions.x, dimensions.y)
+        splash:wait(0.1)
+        return splash:html()
+    end
+
+
+Under the hood ``splash:mouse_click`` simply performs :ref:`splash-mouse-press` followed by :ref:`splash-mouse-release`.
+
+At the moment only left click is supported by Splash. Click with right mouse button or double click is not supported.
+
+
+.. _splash-mouse-hover:
+
+splash:mouse_hover
+------------------
+
+Trigger mouse hover (JavaScript mouseover) event in web page.
+
+**Signature:** ``splash:mouse_hover(x, y)``
+
+**Parameters:**
+
+* x - integer or float with x position of element to be hovered on
+* y - integer or float with y position of element to be hovered on
+
+**Returns:** nil
+
+**Async:**: no.
+
+See notes about mouse events in :ref:`splash-mouse-click`
+
+
+.. _splash-mouse-press:
+
+splash:mouse_press
+------------------
+
+Trigger mouse press event in web page.
+
+**Signature:** ``splash:mouse_press(x, y)``
+
+**Parameters:**
+
+* x - integer or float with x position of element over which mouse button is pressed
+* y - integer or float with y position of element over which mouse button is pressed
+
+**Returns:** nil
+
+**Async:**: no.
+
+See notes about mouse events in :ref:`splash-mouse-click`
+
+.. _splash-mouse-release:
+
+splash:mouse_release
+--------------------
+
+Trigger mouse release event in web page.
+
+**Signature:** ``splash:mouse_release(x, y)``
+
+**Parameters:**
+
+* x - integer or float with x position of element over which mouse button is released
+* y - integer or float with y position of element over which mouse button is released
+
+**Returns:** nil
+
+**Async:**: no.
+
+See notes about mouse events in :ref:`splash-mouse-click`
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -2114,8 +2114,8 @@ Trigger mouse click event in web page.
 
 **Parameters:**
 
-* x - integer or float with x position of element to be clicked
-* y - integer or float with y position of element to be clicked
+* x - number with x position of element to be clicked (distance from the left, relative to viewport)
+* y - number with y position of element to be clicked (distance fromt the top, relative to viewport)
 
 **Returns:** nil
 
@@ -2124,19 +2124,28 @@ Trigger mouse click event in web page.
 
 .. note::
 
-    Coordinates for mouse events must be relative to web document. Element on which action is performed must be inside viewport (must be visible
+    Coordinates for mouse events must be relative to viewport. Element on which action is performed must be inside viewport (must be visible
     to the user). If element is outside viewport and user needs to scroll to see it, you must either scroll
     down with splash or simply set viewport to full with :ref:`splash-set-viewport-full`.
 
 .. note::
 
-    Mouse events are not propagated immediately, to see consequences of click  reflected in page source you must call :ref:`splash-wait`
+    Mouse events are not propagated immediately, to see consequences of click reflected in page source you must call :ref:`splash-wait`
+
+.. hint::
+
+    If you want to get coordinates of html element use JavaScript getBoundingClientRect_, x must be distance from the
+    left, and y distance from the top.
+
+    .. _getBoundingClientRect: https://developer.mozilla.org/en/docs/Web/API/Element/getBoundingClientRect
+
+
 
 Example:
 
 .. code-block:: lua
 
-    -- get button element dimensions with javascript and perform mouse click
+    -- Get button element dimensions with javascript and perform mouse click.
 
     function main(splash)
         assert(splash:go(splash.args.url))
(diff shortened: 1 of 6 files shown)


genius001:
@pawelmhm Very great.

There is something I Encounter, In you code could not process element <a> with target=_blank property.

splash:click only valid for element in viewport, this behavior like splash:png if want get a region but not in a viewport, in this case, splash:png will failed.

pawelmhm:
>  In you code could not process element with target=_blank property.

hmm I'm not sure, why target=blank should be handled differently when clicked? Clicking is done on any elements with some coordinates in viewport so I think it should work regardless of node attributes? Or am I missing something?

> splash:click only valid for element in viewport

are there any use cases for clicking outside viewport? Usually all html elements user wants to click are inside viewport. We could add checks for clicks outside viewports with current code they will just be ignored, we could perhaps raise some error for them but it would require validation logic so would add complexity without clear benefits.

> if want get a region but not in a viewport, in this case, splash:png will failed.

render.png supports screenshots of page regions now, so you can get screenshot of arbitrary page area.

kmike:
Hey @pawelmhm,

Coordinates are the way to go! But what do you think about separating mouse events into their own methods? Users may want to do a right click, or to insert a delay between press and release. We can still provide splash:mouse_click implemented on top of splash:mouse_press and splash:mouse_release.

The use case for clicking elements not in a viewport: you've got coordinates of an element by its CSS selector - in this case you can't guarantee element is in a viewport. We have the same issue with region rendering; in docs it is suggested to use `splash:set_viewport_full()`:

> To render an arbitrary part of a page use region parameter. It should be a table with {left, top, right, bottom} coordinates. Coordinates are relative to current scroll position. Currently you can’t take anything which is not in a viewport; to make sure part of a page can be rendered call splash:set_viewport_full before using splash:png with region. This may be fixed in future Splash versions.

`target=_blank` may be different because Qt may try to open a new window as a result of clicking; I don't know what happens in this case.

genius001:
>
hmm I'm not sure, why target=blank should be handled differently when clicked? Clicking is done on any elements with some coordinates in viewport so I think it should work regardless of node attributes? Or am I missing something?
>

the reason as @kmike said,
>
target=_blank may be different because Qt may try to open a new window as a result of clicking; I don't know what happens in this case.
>

I solved it reference this http://doc.qt.io/qt-4.8/qwebpage.html#acceptNavigationRequest

```
def acceptNavigationRequest(self, webFrame, networkRequest, navigationType):
        if(navigationType == QWebPage.NavigationTypeLinkClicked):
            if(webFrame == self.mainFrame()):
                self.mainFrame().load(networkRequest.url())
            else:
                # TODO will popup 
                #self.mainFrame().load(networkRequest.url())
                createWindow or load it to other frame.

            return False
```

In code snippet, I just load it to self.mainFrame, this behavior same as splash:go,  
Then, we can using splash:png() splash:html() etc. 
If we create new frame/window/tab for target=_blank, that will complicate the code.

pawelmhm:
@genius001 

> In code snippet, I just using self.mainFrame().load(networkRequest.url()),

so this is not about clicking but about navigation, am I right? clicking target blank will cause navigation request  and thus make request to other url and this behavior is not handled by Splash in the way you expect. This means that click() event is fired ok but navigation request is unhandled by splash or maybe it is handled in a way that is not matching your needs. In this case maybe we can consider opening new PR or issue and discuss what expected behavior should be?

@kmike 

> But what do you think about separating mouse events into their own methods? Users may want to do a right click, or to insert a delay between press and release. 

sounds like good idea, will add this. 

> We can still provide splash:mouse_click implemented on top of splash:mouse_press and splash:mouse_release.

current method name is simply click() and not mouse_click(), are we ok with click() name? it is nice and short IMO. We can have mouse_press() and mouse_release() for press and release since pressing and releasing can be done with and without mouse. 

Are we ok with adding only left click in this PR? Adding right click or dbclick is not difficult, but will require extra code and tests, perhaps we can add left click first and add other things later?

kmike:
@pawelmhm I like `mouse_click(x, y)` more; `click` name can be reserved for a method with more high-level API, e.g. click an element by its CSS selector (maybe with a delay between press and release as a parameter), or click on an Element instance (if we add this abstraction).

pawelmhm:
changes:

* refactored click into press/release
* method name mouse_click()
* tests for press/release

splash/tests/test_execute.py
```diff
@@ -3456,3 +3453,145 @@ def test_enable_and_disable_private_mode(self):
...
+        resp = self.request_lua("""
+             function main(splash)
+                assert(splash:go(splash.args.url))
+                splash:mouse_click(nil, nil)
+                splash:wait(0.1)
+                return splash:html()
+            end
+            """, {"url": self.mockurl("jsevent?event_type=click")})
+        self.assertScriptError(resp, ScriptError.LUA_ERROR,
+                               'AssertionError()')
```
kmike: Could you please make errors messages a bit more detailed? See e.g. how are errors raised in splash:go.

kmike:
@pawelmhm looks good! 
Could you please add docs for these methods? 
A test with out-of-screen clicking (i.e. clicking on an element outside a current viewport, or clicking somethign when window is scrolled) would be also nice.

pawelmhm:
- [x] test for click outside viewport
- [ ] docs
- [ ] better error messages

pawelmhm:
@kmike so all items on TODO list are checked now (tests, error messages, docs)

docs/scripting-ref.rst
```diff
@@ -2102,3 +2102,141 @@ Example:
...
+------------------
+
+Trigger mouse click event in web page.
+
+**Signature:** ``splash:mouse_click(x, y)``
+
+**Parameters:**
+
+* x - integer or float with x position of element to be clicked
+* y - integer or float with y position of element to be clicked
```
kmike: in Lua 5.2 there is only a single numeric type, there are no ints or floats

docs/scripting-ref.rst
```diff
@@ -2102,3 +2102,141 @@ Example:
...
+
+    function main(splash)
+        assert(splash:go(splash.args.url))
+        get_dimensions = splash:jsfunc([[
+            function () {
+                rect = document.getElementById('button').getBoundingClientRect();
+                return {"x":rect.left, "y": rect.top}
+            }
+        ]])
+        dimensions = get_dimensions()
```
kmike: this example misses `splash:set_viewport_full`

kmike: or a note that it won't work reliably for all pages :)

docs/scripting-ref.rst
```diff
@@ -2102,3 +2102,141 @@ Example:
...
+        -- wait split second to allow event to propagate
+        splash:wait(0.1)
+        return splash:html()
+    end
+
+Example 2:
+
+.. code-block:: lua
+
+    -- click on element outside viewport
```
kmike: this also works for clicking on element inside a viewport - what about keeping only a single example?

splash/qtrender_lua.py
```diff
@@ -758,6 +758,36 @@ def callback(reply):
     def autoload_reset(self):
         self.tab.autoload_reset()
 
+    def _check_mouse_coordinantes(self, x, y):
+        try:
+            assert isinstance(x, int) or isinstance(x, float)
+            assert isinstance(y, int) or isinstance(y, float)
+        except AssertionError:
+            raise ScriptError({
+                "argument": "x, y",
```
kmike: Could you please split this error in two separate errors for each argument? In all other cases "argument" contains a single argument which fails

splash/qtrender_lua.py
```diff
@@ -758,6 +758,36 @@ def callback(reply):
...
+                "message": "splash mouse event x, y coordinates must be int or float "
```
kmike: there are no ints and floats in Lua; what about changing it to "splash mouse even x coordinate is not a number"?

kmike:
@pawelmhm could you please also add a test with a large page and window.scrollTo? Currently it untested and undocumented if coordinates are absolute or relative to the current viewport or a current scroll position.

docs/scripting-ref.rst
```diff
@@ -2102,3 +2102,141 @@ Example:
...
+* y - integer or float with y position of element to be clicked
+
+**Returns:** nil
+
+**Async:**: no.
+
+
+.. note::
+
+    Coordinates for mouse events must be relative to web document. Element on which action is performed must be inside viewport (must be visible
```
kmike: What does "relative to web document" mean?

docs/scripting-ref.rst
pawelmhm: :+1: good point will fix that

docs/scripting-ref.rst
pawelmhm: sounds like good idea will do

splash/qtrender_lua.py
pawelmhm: :+1:

pawelmhm:
> could you please also add a test with a large page and window.scrollTo? Currently it untested and undocumented if coordinates are absolute or relative to the current viewport or a current scroll position.

in our test resources I added one test resource with [button element that requires scroll](https://github.com/scrapinghub/splash/pull/438/files#diff-3cb74b2bdc3baec8d0368069cdcab2fdR708) this [is tested here](https://github.com/scrapinghub/splash/pull/438/files#diff-98e31963947820b97ea56477e27b6294R3493)

Current test sets viewport full but we can add test with scroll no problem. will add it

docs/scripting-ref.rst
pawelmhm: in JavaScript you have browser window and document. For me this evaluates to

```javascript
window.innerWidth: 1855
window.innerHeight: 925
```

and remains constant no matter what you do (it doesn't change when you scroll). Window is just frame through which you view document. 

Document is what is inside web page, you can get its size by doing:
```javascript
> $(document).height()
9582
> $(document).width()
1840
```
 I was always thinking that by "viewport" we mean browser window, but it seems like this is not obvious and clear. Is my assumption correct? Coordinates of click should be coordinates inside web page, so they should be inside $(document), e.g. for this github page ATM you can click on elements with maximum y 9000.

Is this clear or am i missing something here?

kmike:
@pawelmhm my main question about these viewports, documents and windows is whether coordinates change when we scroll the window. E.g. if the element has (100px, 500px) coordinates relative to the whole document and window is scrolled down 200px then should we click on (101px, 501px) or (101px, 301px) - it is not tested currently and not documented explicitly.

pawelmhm:
> my main question about these viewports, documents and windows is whether coordinates change when we scroll the window. E.g. if the element has (100px, 500px) coordinates relative to the whole document and window is scrolled down 200px then should we click on (101px, 501px) or (101px, 301px) - it is not tested currently and not documented explicitly.

no coordinates dont change when we scroll, they should describe element position inside web page regardless of where user is looking at, I'll add this explanation to docs

pawelmhm:
ok @kmike sorry turns out I was wrong. 

Coordinates are relative to viewport, I missed this. I realized today that I'be been using JS getBoundingClientRect() all along in all tests and this function returns [element coordinates relative to viewport](https://developer.mozilla.org/en/docs/Web/API/Element/getBoundingClientRect). Also when adding test for scroll I realized that coordinates DO indeed change when you scroll.

So I updated docs to reflect that and added test for scrolling

splash/qtrender_lua.py
```diff
@@ -758,6 +758,36 @@ def callback(reply):
     def autoload_reset(self):
         self.tab.autoload_reset()
 
+    def _check_mouse_coordinantes(self, x, y):
+        try:
+            assert isinstance(x, int) or isinstance(x, float), "x"
+            assert isinstance(y, int) or isinstance(y, float), "y"
+        except AssertionError as e:
```
kmike: A nitpick: this exception won't be fired if Splash is started with `python -O` or with `PYTHONOPTIMIZE` env variable set; it is not too important, but could you please rewrite it without assert?

pawelmhm: :+1: fixed in https://github.com/scrapinghub/splash/pull/438/commits/3cf182eafc52643205d986dbe197d2a5207056f8

kmike:
Thanks @pawelmhm!

wusy1209:
hi @pawelmhm , that is useful , Can you add a method of `mouse_move `click ?

pawelmhm:
@wusy1209  there is `moue_hover` added in this PR, it moves mouse to some location http://splash.readthedocs.org/en/stable/scripting-ref.html#splash-mouse-hover

wusy1209:
Thanks @pawelmhm  ！

kmike:
@pawelmhm do you recall why is the method named `mouse_hover`, not `mouse_move`?

pawelmhm:
> @pawelmhm do you recall why is the method named mouse_hover, not mouse_move?

hmm good question, we didn't discuss this. I thought mouse_hover is intuitive it refers to JS event name (which is mouse over or mouse hover), do you think mouse move is more natural to users? Not sure if we can change method name after release.

kmike:
@pawelmhm there are also [mousemove](https://developer.mozilla.org/en/docs/Web/Events/mousemove) JS event

pawelmhm:
our method fires mouseover event (mouse hover/mouse over) https://developer.mozilla.org/en-US/docs/Web/Events/mouseover see test here: https://github.com/scrapinghub/splash/blob/master/splash/tests/test_execute.py#L3574 so the reason why I named it hover is that it fires this event

PR #447, opened 2016-04-21
Add splash:send_keys() method

eskerda:
This set of commits adds support for native send_keys method.


eskerda:
Some questions:

- Should we support Key Modifiers to go along with key presses? (like Ctrl + Shift) Can't think of a valid usecase except for automatic usage of webapps that might have keyboard shortcuts.
- This isn't handling a usecase with a selector. Assumedly, it would work like:
    - Find selector element, click
    - send keys
    - element blur? Maybe not necessary

eskerda:
- Added support for `click` command with selector (https://github.com/scrapinghub/splash/pull/438#issuecomment-206790014)
- Added support for selector on send_keys command

splash/qtrender_lua.py
```diff
@@ -771,24 +771,41 @@ def _assert_number(value, name):
...
+    @command()
+    def click(self, selector):
+        element = self.tab.find_element(selector)
+        if element.isNull():
+            raise ScriptError({
+                "argument": selector,
+                "message": "element at {!r} not found".format(selector)
+            })
+        x, y = element.geometry().x(), element.geometry().y()
+        return self.mouse_click(x, y)
```
eskerda: I can't seem to make this exception reach up if click is called from another command. You can see one test failing checking for that. Ideas?

eskerda:
Maybe we should discuss the signature for `send_keys` function

It currently accepts `text`, `selector (def: None)` and `key_type (def: None)`

Some users may want to send raw send_keys like
```
splash:send_keys('', nil, 'Tab')
```
but expecting a nil to call it maybe is ugly.

Reference for key types can be found at http://doc.qt.io/qt-5/qt.html#Key-enum

splash/browser_tab.py
```diff
@@ -870,6 +870,23 @@ def _post_mouse_event(self, type, button, x, y):
...
+        """
+        key_type = key_type or 'unknown'
+        # http://doc.qt.io/qt-5/qt.html#Key-enum
+        key_type = getattr(Qt, 'Key_%s' % key_type, Qt.Key_unknown)
+        self._send_keys(text, key_type)
+
+    def _send_keys(self, text, key_type):
+        modifiers = QApplication.keyboardModifiers()
+        event = QKeyEvent(QEvent.KeyPress, key_type, modifiers, text)
+        QApplication.postEvent(self.web_page, event)
```
kmike: Hm, I wonder if it is better to do that in a loop. If I understand the code properly, currently it sends a single keyboard event which results in `text` being entered. It is like an user who presses a single key on a keyboard, and this key generating some text - usually it is a single character which depends on user's currently chosen language, but here it is complete text.

splash/browser_tab.py
```diff
@@ -870,6 +870,23 @@ def _post_mouse_event(self, type, button, x, y):
...
+        :return: None
+        """
+        key_type = key_type or 'unknown'
+        # http://doc.qt.io/qt-5/qt.html#Key-enum
+        key_type = getattr(Qt, 'Key_%s' % key_type, Qt.Key_unknown)
+        self._send_keys(text, key_type)
+
+    def _send_keys(self, text, key_type):
+        modifiers = QApplication.keyboardModifiers()
+        event = QKeyEvent(QEvent.KeyPress, key_type, modifiers, text)
```
kmike: Should we also send KeyRelease events after KeyPress?

splash/tests/test_execute.py
```diff
@@ -3653,3 +3653,34 @@ def test_release_with_bad_arguments(self):
...
+                    function () {
+                        var inputs = document.getElementsByTagName('input');
+                        var values = [];
+                        for (var i = 0; i < inputs.length; i++) {
+                            values.push(inputs[i].value);
+                        }
+                        return values.join('|');
+                    }
+                ]])
+                splash:send_keys('', 'Tab')
```
kmike: Hm, this API looks strange. Any ideas on how to make it cooler?

splash/browser_tab.py
eskerda: We could of course send each letter on a loop, but that might not be straightforward. What about special characters or acutes. It would be quite similar as: "an unknown key that when pressed inputs `â`".

IMHO `Key_unknown` as a single "magic" key works. We can improve it later if there's the use case of actually needing an accurate keyboard simulation.

splash/browser_tab.py
eskerda: I saw no apparent difference when sending a KeyRelease after KeyPress, so decided to go without it.

I see in this QT example they send an empty KeyRelease event, though it's for single keys.
```C++
void MyInputPanelContext::sendCharacter(QChar character)
{
    QPointer<QWidget> w = focusWidget();

    if (!w)
        return;

    QKeyEvent keyPress(QEvent::KeyPress, character.unicode(), Qt::NoModifier, QString(character));
    QApplication::sendEvent(w, &keyPress);

    if (!w)
        return;

    QKeyEvent keyRelease(QEvent::KeyPress, character.unicode(), Qt::NoModifier, QString());
    QApplication::sendEvent(w, &keyRelease);
}

```
http://doc.qt.io/qt-4.8/qt-tools-inputpanel-example.html

To note on this example, they send as key code the unicode value of the character. Might work, if we are to implement keyboard input on a loop.

splash/browser_tab.py
eskerda: As an extra, this is the documentation on QKeyEvent

```
int QKeyEvent::key() const

Returns the code of the key that was pressed or released.

See Qt::Key for the list of keyboard codes. These codes are independent of the
underlying window system. Note that this function does not distinguish between
capital and non-capital letters, use the text() function (returning the Unicode
text the key generated) for this purpose.

A value of either 0 or Qt::Key_unknown means that the event is not the result of
a known key; for example, it may be the result of a compose sequence, a keyboard
macro, or due to key event compression.

See also Qt::WA_KeyCompression.
```
http://doc.qt.io/qt-5/qkeyevent.html#key

splash/tests/test_execute.py
eskerda: Hmm, yes, it makes no sense to have the same signature for a succession of keys (aka, full input) and a single key.

Maybe:
```
splash:send_key('Tab')
splash:send_keys('Hello World')
splash:send_key('Enter')
```

splash/browser_tab.py
kmike: Yeah, it may go out of control pretty soon: if we want to closely emulate key presses we'd have to add keyboard layouts support to match keys pressed and unicode values produced. That'd be a nice feature, but that's too much for now.

splash/browser_tab.py
kmike: It should be possible to detect using JavaScript: there are keyup, keydown and keypress JS events, and likely they'll behave differently with just keyPress vs keyPress+keyRelease.

splash/tests/test_execute.py
eskerda: @kmike So, is the proposed API better than before? `send_key` + `send_keys` might be confusing.

immerrr: In Emacs, for example, there's a "keyboard macro" format which uses spaces as non-obligatory separators between keys which is good for describing both single and multiple keystrokes and which I find quite easy on the eyes.
```
'f' => 'f'
'foobar' => "foobar"
'f o o b a r' => "foobar"
'SPC' => " "
'RET' => "\n"
'Hello SPC world' => "Hello world"
'2*Hello 3*SPC world' => 'HelloHello   world'
'<f1>' => F1 (the key)
'< f 1 >' => '<f1>' (string)
```

We can borrow the idea of using spaces as separators at the price of requiring "SPC" to be input instead of " ".

The complete format is described as follows (for more inspiration):
```
Format of keyboard macros during editing:

Text is divided into "words" separated by whitespace.  Except for
the words described below, the characters of each word go directly
as characters of the macro.  The whitespace that separates words
is ignored.  Whitespace in the macro must be written explicitly,
as in "foo SPC bar RET".

 * The special words RET, SPC, TAB, DEL, LFD, ESC, and NUL represent
   special control characters.  The words must be written in uppercase.

 * A word in angle brackets, e.g., <return>, <down>, or <f1>, represents
   a function key.  (Note that in the standard configuration, the
   function key <return> and the control key RET are synonymous.)
   You can use angle brackets on the words RET, SPC, etc., but they
   are not required there.

 * Keys can be written by their ASCII code, using a backslash followed
   by up to six octal digits.  This is the only way to represent keys
   with codes above \377.

 * One or more prefixes M- (meta), C- (control), S- (shift), A- (alt),
   H- (hyper), and s- (super) may precede a character or key notation.
   For function keys, the prefixes may go inside or outside of the
   brackets:  C-<down> = <C-down>.  The prefixes may be written in
   any order:  M-C-x = C-M-x.

   Prefixes are not allowed on multi-key words, e.g., C-abc, except
   that the Meta prefix is allowed on a sequence of digits and optional
   minus sign:  M--123 = M-- M-1 M-2 M-3.

 * The ‘^’ notation for control characters also works:  ^M = C-m.

 * Double angle brackets enclose command names:  <<next-line>> is
   shorthand for M-x next-line RET.

 * Finally, REM or ;; causes the rest of the line to be ignored as a
   comment.

Any word may be prefixed by a multiplier in the form of a decimal
number and ‘*’:  3*<right> = <right> <right> <right>, and
10*foo = foofoofoofoofoofoofoofoofoofoo.

Multiple text keys can normally be strung together to form a word,
but you may need to add whitespace if the word would look like one
of the above notations:  ‘; ; ;’ is a keyboard macro with three
semicolons, but ‘;;;’ is a comment.  Likewise, ‘\ 1 2 3’ is four
keys but ‘\123’ is a single key written in octal, and ‘< right >’
is seven keys but ‘<right>’ is a single function key.  When in
doubt, use whitespace.
```

eskerda:
> In Emacs, for example, there's a "keyboard macro" format which uses spaces as non-obligatory separators between keys which is good for describing both single and multiple keystrokes and which I find quite easy on the eyes.

I like that format too, but implementation won't be straightforward. Current implementation is quite easy because we directly support http://doc.qt.io/qt-5/qt.html#Key-enum key names, input `unknown` key by default and do not support modifiers. I worry this would take much, and the issue I'm trying to fix is starting to be a bit urgent (about that, the issue can be fixed with just current `send_keys`, no key types nor modifiers needed).

Maybe we can think on the simplest for now and leave a better syntax for the future (with selectors and so on) like with mouse_click?

If that was the case I could simply remove support for key types. Mouse clicking to coordinates and direct input of text can cover at least 80% of the cases.

eskerda:
With @immerrr contribution https://github.com/scrapinghub/splash/pull/456, we've got now support for a limited subset of the syntax for keyboard macros on emacs. That goes under `send_keys`.

Added a shortcut that does not need the syntax, to avoid using `SPC` too much (specially when handling input from outside of splash, where it might come with spaces or not). That goes under `send_text`.

@kmike Does this API look cooler?

```
splash:send_keys('TAB Hello SPC World RET`)
splash:send_keys('<Tab>')
splash:send_text('Hello World')
splash:send_keys('<Return>')
```

splash/browser_tab.py
```diff
@@ -889,19 +889,28 @@ def send_keys(self, text):
...
+        # 'LFD': 0, ## QChar.LineFeed?
+        # 'NUL': 0, ## QChar.Null?
+    }
+
+    QT_KEY_INPUTS = {
+        # TODO: More needed?
+        Qt.Key_Return: '\n',
+        Qt.Key_Enter: '\n',
+        Qt.Key_Space: ' ',
+        Qt.Key_Tab: '\t',
```
eskerda: It turns out that for the Return event to work it needs to input `\n`, so I added this mapping here. Actually, if we wanted to limit the number of QtKeys supported, we could do so here.

splash/browser_tab.py
```diff
@@ -870,6 +871,72 @@ def _post_mouse_event(self, type, button, x, y):
...
+    }
+
+    def _send_word(self, word):
```
kmike: Why is it called `_send_word`? 
I think it should work better as a separate function in splash.qtutils.

kmike:
The new API is very cool 👍 

Could you please add a couple more tests? 

* A test for non-ascii characters (both for send_keys and send_text);
* a test for entering text which contains \n;
* a test for send_text which checks that text like RET can be sent;
* tests for DEL and ESC keys.

It also needs docs.

splash/qtutils.py
```diff
@@ -299,3 +321,47 @@ def qt_header_items(request_or_reply):
...
+    keyboard macro."
+    """
+    modifiers = QApplication.keyboardModifiers()
+    text = list(text) or ['']
+    for x in text:
+        event = QKeyEvent(QEvent.KeyPress, key_type, modifiers, x)
+        QApplication.postEvent(target, event)
+        # Key release does not generate any input
+        event = QKeyEvent(QEvent.KeyRelease, key_type, modifiers, '')
+        QApplication.postEvent(target, event)
```
eskerda: @kmike: Turns out, regardless of the key event sent, browser understands a keydown + keyup as a correct keypress. Assuming that and being the implementation straightforward, it makes sense to send an event per key, so I added it.

eskerda:
Added a big bunch of tests + a key event per character.

Just missing docs

splash/tests/test_execute.py
```diff
@@ -3719,3 +3719,262 @@ def test_send_keys_enter_event(self):
...
+                        return document.getElementById('text').value
+                    }
+                ]])
+                splash:send_text('Поехали!')
+                assert(splash:wait(0.5))
+                return get_input()
+            end
+            """, {"url": self.mockurl("focused-input")})
+        self.assertStatusCode(resp, 200)
+        self.assertEqual('Поехали!', resp.text)
```
kmike: u'Поехали' is better in Python 2

splash/tests/test_execute.py
```diff
@@ -3719,3 +3719,262 @@ def test_send_keys_enter_event(self):
...
+            function main(splash)
+                assert(splash:go(splash.args.url))
+                assert(splash:wait(0.5))
+                get_input = splash:jsfunc([[
+                    function () {
+                        return document.getElementById('text').value
+                    }
+                ]])
+                splash:send_text('RET ESC TAB SPC')
+                assert(splash:wait(0.5))
```
kmike: can we change all these wait to 0.0 to make tests run faster?

eskerda:
- Added docs
- Made all waits on tests 0
- Are there too many tests? Maybe I can cut them down.
- Anything missing?

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,128 @@ Trigger mouse release event in web page.
...
+
+Key sequences are specified by using a small subset of emacs edmacro syntax:
+
+* whitespace is ignored and only used to separate the different keys
+* characters are literally represented
+* ``RET``, ``SPC``, ``TAB``, ``DEL``, ``ESC`` are reserved and represent
+  special control characters.
+* words within brackets represent function keys, like ``<Return>``, ``<Left>``
+  or ``<Home>``. See `Qt docs`__ for a full list of function keys. ``<Foo>``
+  will try to match ``Qt::Key_Foo``.
```
kmike: What's the difference between `RET` and `<Return>`?

splash/qtutils.py
```diff
@@ -299,3 +321,47 @@ def qt_header_items(request_or_reply):
...
+    key_type = EDMACRO_SHORTCUTS.get(key)
+    if key_type:
+        text = QT_KEY_INPUTS.get(key_type, '')
+        return qt_send_text(text, target, key_type)
+
+    # Try to match "<Key_Name>"
+    key_match = re.match(r'^\<(.+)\>$', key)
+    if key_match:
+        key_type = getattr(Qt, 'Key_%s' % key_match.group(1),
+                           Qt.Key_unknown)
```
kmike: I think it is better to raise an error for unsupported keys, it should make debugging easier

docs/scripting-ref.rst
eskerda: None, from https://github.com/emacs-mirror/emacs/blob/master/lisp/edmacro.el

```

 * A word in angle brackets, e.g., <return>, <down>, or <f1>, represents
   a function key.  (Note that in the standard configuration, the
   function key <return> and the control key RET are synonymous.)
   You can use angle brackets on the words RET, SPC, etc., but they
   are not required there.
```

Maybe we prefer non-capitalized words for function keys? <return> and <down> instead of <Return> and <Down>?

immerrr: Then you'll have to title-case them before trying to do `getattr(Qt, 'Key_%s' % key)`

eskerda: Yes, plus it will accept 'Return', 'ReTuRn' and 'return' as 'Key_Return', which should be nice?

kmike: Heh. I'm on fence about that: on one hand, 3-letter shortcuts provide a second way to do the same as `<Shortcuts>`, this is not good. I don't see what're the advantages: `<Space>` is more readable than `SPC`. On another hand, this way the format becomes compatible with the existing format (edmacro), and this is awesome. I'd prefer not to invent a format if we can. But only a small subset of edmacro is implemented.

Is there a reason why 'special function keys' are separated from regular 'function keys' in edmacro format?

eskerda: I am not sure and probably @immerrr can shed more and better light into it, but I guess it's some kind of compatibility with ASCII control characters

immerrr: edmacro format *is* repetitive, there are more ways to do smth, with Return key being one of the richest having `RET`, `<return>`, `^M`, `C-m` and `\015` as alternative representations. I suppose those shorthand notations were selected as one of the most frequent ones to be used.

kmike: My vote is to remove support for RET, DEL, etc. It should make validation easier, we'll be able to detect typos and raise errors. Alternatively, we could support full edmacro format, I'm also fine with that.

immerrr: Ok, for now let's leave `<Key>` syntax only.

eskerda:
@kmike @immerrr All should be good now. Anything missing (more tests, remove tests, etc). Also worth checking if documentation is good: https://github.com/scrapinghub/splash/blob/89e22570acbc005d5c44378bbb3693407a3fceb8/docs/scripting-ref.rst#splashsend_keys

Also, I can squash all commits into one if that's better

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+
+.. code-block:: lua
+
+    function main(splash)
+        assert(splash:go(splash.args.url))
+        assert(splash:wait(0.5))
+        splash:send_keys("<Tab>")
+        splash:send_text("zerocool")
+        splash:send_keys("<Tab>")
+        splash:send_text("hunter2")
```
immerrr: Let's maybe add spaces to `send_text` examples here and to commented-out `send_keys` example below in case someone misses the table in `send_keys` doc.

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+Following table shows some examples of macros and what they would generate on
+an input:
+
+======================    ===============
+Macro                     Result
+======================    ===============
+``Hello World``           ``HelloWorld``
+``Hello <Space> World``   ``Hello World``
+``< S p a c e >``         ``<Space>``
+``Hello <Home> DEL``      ``ello``
```
immerrr: `DEL` is no more

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+splash:send_text
+----------------
+
+Send text as keyboard events to page context.
+
+**Signature:** ``splash:send_text(text)``
+
+**Parameters:**
+
+* text - string to be sent as keyboard events.
```
immerrr: In `send_text` doc I'd probably avoid mentioning keyboard events altogether and just say that the text is sent as inputu literally, character-for-character, and that's all. And refer to `send_keys` as something that works closer to events and function/special keys.

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+
+.. code-block:: lua
+
+    function main(splash)
+        local get_elem_pos = splash:jsfunc([[
+            function (selector) {
+                var elem = document.querySelector(selector);
+                var rect = elem.getBoundingClientRect();
+                return {"x": rect.left, "y": rect.top}
+            }
```
immerrr: Can we just use `document.querySelector(selector).focus()` here? Do we really need to click it?

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+        -- note how this could be translated to
+        -- splash:send_keys("<Tab> zerocool <Tab> hunter2 <Return>")
+        assert(splash:wait(0))
+        -- ...
+    end
+
+We can't always assume that a TAB will focus the input we want or an Enter will
+submit a form. By using :ref:`splash-mouse-click` we can focus an input. The
+following example will click an input, fill in a form and submit it. It assumes
+there are two arguments passed to splash, `username` and `password`.
```
immerrr: :+1: nice remark

splash/tests/test_execute.py
```diff
@@ -3653,3 +3653,309 @@ def test_release_with_bad_arguments(self):
...
+                splash:send_keys('<Delete>')
+                splash:send_keys('<End>')
+                splash:send_keys('<Left>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
+                splash:send_keys('<Backspace>')
```
immerrr: Maybe loops would be better here?

splash/tests/test_execute.py
```diff
@@ -3653,3 +3653,309 @@ def test_release_with_bad_arguments(self):
...
+        self.assertEqual(expected, result)
+
+    def test_key_error(self):
+        resp = self.request_lua("""
+            function main(splash)
+                splash:send_keys('<Foobar>')
+            end
+        """)
+        self.assertScriptError(resp, ScriptError.SPLASH_LUA_ERROR,
+                               message="Unknown key")
```
immerrr: :+1:

immerrr:
> Also, I can squash all commits into one if that's better

I'd say it is. @kmike ?

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,126 @@ Trigger mouse release event in web page.
...
+
+.. code-block:: lua
+
+    function main(splash)
+        local get_elem_pos = splash:jsfunc([[
+            function (selector) {
+                var elem = document.querySelector(selector);
+                var rect = elem.getBoundingClientRect();
```
kmike: it may be not the best way, see https://github.com/scrapinghub/splash/pull/472

docs/scripting-ref.rst
eskerda: You are right. For trivial inputs that will work, but sometimes forms are generated via js and only way to focus them is clicking on a transparent div that sits on top of the input.

I think mouse click + send keys / text is a pretty complete example of "all things form input". But limiting example to simpler case also sounds like a good idea.

eskerda: Was just thinking, what about pressing buttons (without resorting to form.submit)?

eskerda: Just found out another thing about using focus (not really important).

Since focus happens on JS and not on the event loop, a wait must be issued after each send keys, like

```
-- This will input barbaz on assumed input bar
focus('foo')
send_text('bar')
focus('bar')
send_text('baz')
assert(splash:wait(0))
```

```
-- This does what we want
focus('foo')
send_text('bar')
assert(splash:wait(0))
focus('bar')
send_text('baz')
assert(splash:wait(0))
```

Anyway, I have updated the example to include both a focus solution and a click solution, at the same time (focus for selecting an input and click for pressing the send button).

docs/scripting-ref.rst
eskerda: right. I updated the example to reflect that

eskerda:
- squashed commits
- rebased master
- updated docs

splash/qtutils.py
```diff
@@ -299,3 +310,42 @@ def qt_header_items(request_or_reply):
...
+        - <FooBar> -> Qt.Key_FooBar
+    See: http://doc.qt.io/qt-5/qt.html#Key-enum for valid key types
+    """
+
+    # Try to match "<Key_Name>"
+    key_match = re.match(r'^<(\w+)\>$', key)
+    if key_match:
+        key_type = getattr(Qt, 'Key_%s' % key_match.group(1), None)
+        if not key_type:
+            raise TypeError('Unknown key: %s' % key_match.group(1))
```
kmike: TypeError is used when type of Python variable is incorrect, e.g. when int is passed instead of string. I think it should be ValueError.

kmike:
Awesome work, thanks @eskerda and @immerrr!

PR #448, opened 2016-04-21
Update Install Documentation for Mac

lsenta:
Default way to run docker on a Mac is through docker-machine now.

Please have someone else test before merging.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -21,7 +21,13 @@ Linux + Docker
 OS X + Docker
 -------------
 
-1. Install Docker_ (via Boot2Docker_).
+1. Install Docker_ (via the Toolbox Instructions_).
+2. Run the Docker Machine & load its environment
+
+       $ docker-machine start default
+       
+       $ eval "$(docker-machine env default)"                                                            ☺
+
 2. Pull the image::
 
        $ docker pull scrapinghub/splash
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -21,7 +21,13 @@ Linux + Docker
 OS X + Docker
 -------------
 
-1. Install Docker_ (via Boot2Docker_).
+1. Install Docker_ (via the Toolbox Instructions_).
+2. Run the Docker Machine & load its environment
+
+       $ docker-machine start default
+       
+       $ eval "$(docker-machine env default)"                                                            ☺
```
kmike: what is `☺` for? :)

docs/install.rst
```diff
@@ -21,7 +21,13 @@ Linux + Docker
 OS X + Docker
 -------------
 
-1. Install Docker_ (via Boot2Docker_).
+1. Install Docker_ (via the Toolbox Instructions_).
+2. Run the Docker Machine & load its environment
+
+       $ docker-machine start default
```
kmike: There are no machines available by default; AFAIK there is nothing special about `default` name

docs/install.rst
lsenta: Whoops, thanks

docs/install.rst
lsenta: I added the line to create the machine.
It'd be better if someone can confirm this on a fresh install.

Issue #208, opened 2015-04-27
splash:on_request error handling


kmike:
Currently errors in splash:on_request callbacks are logged, but the errors are then discarded. It may be OK, but there should be an example in docs on how to return these errors e.g. using pcall. 

Another option is to raise an error and stop the script if there is an error in on_request callback.

PR #451, opened 2016-04-28
splash:set_custom_headers() should not cause segfaults

pawelmhm:
invalid header values (e.g. integer values) passed to splash/utils/to_bytes() in _handle_custom_headers() were causing SEGFAULT

Commit adds try except around createRequest() to avoid bugs like this in the future.
It also adds validation of custom_headers.

### DIFF
--- a/splash/network_manager.py
+++ b/splash/network_manager.py
@@ -105,7 +105,21 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
     def _on_finished(self, reply):
         reply.deleteLater()
 
+
     def createRequest(self, operation, request, outgoingData=None):
+        try:
+            return self._createRequest(operation, request, outgoingData=outgoingData)
+        except:
+            # There are too many errors that can happen in NAT, in case something happens
+            # splash will segfault. To avoid this in case of Splash NAT error we will just
+            # delegate to superclass (no splash intervention).
+            # WARNING: this can hide error from user (no more segfault on error, and response will be 200,
+            # but without intended splash behavior). This is not perfect but is better than crash.
+            self.log("internal error in _createRequest middleware", min_level=1)
+            self.log(traceback.format_exc(), min_level=1, format_msg=False)
+            return super(ProxiedQNetworkAccessManager, self).createRequest(operation, request, outgoingData)
+
+    def _createRequest(self, operation, request, outgoingData=None):
         """
         This method is called when a new request is sent;
         it must return a reply object to work with.
@@ -226,7 +240,12 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
             headers = headers.items()
 
         for name, value in headers or []:
-            request.setRawHeader(to_bytes(name), to_bytes(value))
+            try:
+                request.setRawHeader(to_bytes(name), to_bytes(value))
+            except TypeError:
+                msg = "invalid header {}: {}. Header keys and values must be string or bytes"
+                self.log(msg.format(name, value), min_level=1, format_msg=False)
+                continue
 
     def _handle_request_cookies(self, request):
         self.cookiejar.update_cookie_header(request)
--- a/splash/qtrender_lua.py
+++ b/splash/qtrender_lua.py
@@ -903,7 +903,20 @@ class Splash(BaseExposedObject):
 
     @command(table_argument=True)
     def set_custom_headers(self, headers):
+        self.validate_headers(headers)
         self.tab.set_custom_headers(self.lua.lua2python(headers, max_depth=3))
+
+    def validate_headers(self, headers):
+        for key, value in headers.items():
+            try:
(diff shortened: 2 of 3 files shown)


splash/network_manager.py
```diff
@@ -105,7 +105,21 @@ def _on_ssl_errors(self, reply, errors):
...
+            return self._createRequest(operation, request, outgoingData=outgoingData)
+        except:
+            # There are too many errors that can happen in NAT, in case something happens
+            # splash will segfault. To avoid this in case of Splash NAT error we will just
+            # delegate to superclass (no splash intervention).
+            # WARNING: this can hide error from user (no more segfault on error, and response will be 200,
+            # but without intended splash behavior). This is not perfect but is better than crash.
+            self.log("internal error in _createRequest middleware", min_level=1)
+            self.log(traceback.format_exc(), min_level=1, format_msg=False)
+            return super(ProxiedQNetworkAccessManager, self).createRequest(operation, request, outgoingData)
```
kmike: :+1:

splash/network_manager.py
```diff
@@ -226,7 +240,12 @@ def _handle_custom_headers(self, request):
             headers = headers.items()
 
         for name, value in headers or []:
-            request.setRawHeader(to_bytes(name), to_bytes(value))
+            try:
+                request.setRawHeader(to_bytes(name), to_bytes(value))
+            except TypeError:
+                msg = "invalid header {}: {}. Header keys and values must be string or bytes"
+                self.log(msg.format(name, value), min_level=1, format_msg=False)
+                continue
```
kmike: I think that's fine to accept numbers as header values without raising an error; it looks like a common pain point.

pawelmhm: My thinking here is that we'll have to do str() on those values. str() can throw some error again (e.g. someone can pass Lua function by accident or pass something else weird), so to be 100% sure no exception happens we'll have to wrap this call to str() in another try-except which may decrease code readability.

HTTP headers are strings, and users can easily fix errors on their side - just pass proper value, string not number.

pawelmhm:
I was thinking that some future TODO (not sure if in this PR) could involve keeping track of those NAT errors and returning 500 response to client, so that we don't loose track of those errors in createRequest. In current master we have segfaults if exception happens in createRequest. On this branch we'll have no segfaults but 200 with error traceback in logs.

 Probably ideal situation would be having 500 returned to user, traceback in logs and no segfault

splash/network_manager.py
kmike: Yeah, but these values are usually set in splash:on_request, and it is hard to get an error from there, so it is harder for user to detect this error in script. Errors are raised so that users can fix their scripts and make them more robust, but here error doesn't serve this need because it is not obvious how to fix a script after this error (it is only in the logs). 

You're right that we should make it robust (and so we may have to catch str() errors), but I think it worths it; the alternative is worse.

kmike:
@pawelmhm from logs you've sent to me the other day it looked like errors were raised for network requests which QT tries to process after we already returned result to user; in this case it is impossible to return HTTP 500 error to the client.

pawelmhm:
> pawelmhm from logs you've sent to me the other day it looked like errors were raised for network requests which QT tries to process after we already returned result to user; in this case it is impossible to return HTTP 500 error to the client.

yes this is correct. But here I'm not actually fixing this past error, just one with invalid headers set by set_custom_header(). 

You are right that some errors in createRequest will arise after response will be returned to client and for them it does not make sense to communicate 500. But not ALL of errors will be like this. For example IF some yet unknown error will happen when passing values from Lua to Python, e.g. someone makes weird call, for example passing function to go()

```lua
function main (splash)
    function alfa () 
        -- passing this to splash go, what will happen? will there be some QT crash?
   end
   splash:go(alfa)
end
```

if something like this would cause error in  NAT.createRequest(), we should probably return 500 to client. With this branch code if there is exception we will return 200 and catch exception in logs.

But returning error codes on errors in Lua manipulation is not that easy, we have tickets for that here: https://github.com/scrapinghub/splash/issues/161 and https://github.com/scrapinghub/splash/issues/208 so I think doing this properly would require us to think how to properly communicate errors in NAT to splash client.

splash/network_manager.py
pawelmhm: yeah sounds convincing will add support for integers (do you think this  should be documented too? I mean the fact that you can pass integers to header values?)

kmike: I think that's fine to leave it as an undocumented edge case to encourage users to use strings :)

pawelmhm:
I found out that this error happens not only in set_custom_headers() but in go() and other methods as well (basically any time you pass headers to splash request). 

So later this week I'll push some more commits fixing it for all situations not just in splash:set_custom_headers()

kmike:
Closed by #644. Thanks @pawelmhm!

PR #452, opened 2016-04-28
Added flash support, Added HAR contents, Created a test for Flash support

shaylevi2:
Further:

* DOCKERFILE now contains flash installation (using provision.sh) and is not calling the rm -rf command in provision.sh that deletes necessary perl modules.
* Plugins (Flash) are now enabled by default, maybe we should export this as an optional parameter
* HAR file contents are now saved by default (because that is what you expect of an HAR file...) - We can/should also provide that as an optional parameter
* I removed the useless param of "binary_content" that wasn't used and anyway did not work.
* Changed default view port size to 1920x1080, websites behave differently and this seems like a more standard size for today.

Hope the tests will go smoothly, if they won't I'll repair and recreate the request.


kmike:
Thanks @shaylevi2, looking forward to it!

There were past attempts to make flash and HAR content work.

For flash the problem was with rendering: Flash plugin rendered its contents without using QT utilities, so it was harder to take a screenshot, existing method didn't work. It was worked around in a branch (https://github.com/andresp99999/splash/tree/flash_support), but the solution had other issues.

For HAR content the problem was that the content of response is streaming, and it can be only read once. If you read it in QNetworkManager then it is no longer available for browser, and vice versa. A custom QNetworkReply subclass is used by other projects to overcome that.

shaylevi2:
Hi kimke, 

I've already solved the HAR content problem without proxying the class, which is a lot better as it avoids cached responses issues. 

Regarding the flash, it seems to work fine with screenshots as well :) 

The only problem I'm failing on now is the non-utf non-ascii response which crashes the server. I'm gonna work on it next week and I'll reopen the pull request once it passes. 

It's going to be awesome. 

(I solved the HAR issue using "peek" instead of "read" to get a copy of the content while keeping it available for the web page)

kmike:
Cool, great to hear that! Maybe qt5 helps with some of these issues. 

I recall @AlexIzydorczyk said peek didn't work for him, I wonder why. It should be a better solution indeed if it works.

shaylevi2:
Opened a new pull request.

genius001:
If peek work, a lot of things will have better solution.
That will be excellent!

shaylevi2:
@genius001 It's already working. See my new pull request:

https://github.com/scrapinghub/splash/pull/454

PR #454, opened 2016-05-02
Added flash support, Added HAR contents

shaylevi2:
### Further:
* DOCKERFILE now contains flash installation (using provision.sh) and is not calling the rm -rf command in provision.sh that deletes necessary perl modules.
* Plugins (Flash) are now enabled by default, maybe we should export this as an optional parameter
* HAR file contents are now saved by default (because that is what you expect of an HAR file...) - We can/should also provide that as an optional parameter
* I removed the useless param of "binary_content" that wasn't used and anyway did not work.
* Changed default view port size to 1920x1080, websites behave differently and this seems like a more standard size for today.
* Fixed python2.7 tests and now forcing them to pass

### Missing:
* Flash test - I tried to use the one from decades ago but I couldn't make it start.
* HAR with content tests - Current HAR tests work, maybe we should add more tests for HAR with content.

### Issues:
* The test with the Cyrillic encoding fails, as it wasn't doing anything before I allowed myself to comment the problematic line out - I tried to fix it multiple times but failed, would appreciate help on that.
* _test_meta_redirect_slowload2_wait_more_ test fails, after trying endlessly to solve it, I eventually gave up. It doesn't fail, it simply freezes. I would appreciate help on that too.

That's it, hope to see the pull request approved :)


shaylevi2:
@kmike

splash/har/qt.py
```diff
@@ -129,16 +133,17 @@ def reply2har(reply, include_content=False, binary_content=False):
...
-            res["content"]["text"] = base64.b64encode(data)
-            res["content"]["size"] = len(data)
-
+        content = getattr(reply, 'content', None)
+        if content is not None:
+            res["content"]["size"] = content.size()
+            bytes_array = bytes(content)
+
+            if not is_binary_string(bytes_array):
+                res["content"]["text"] = bytes_array.decode('utf8', 'replace')
```
kmike: I think we shouldn't try to detect if data is binary ourselves. `.decode('utf8', 'replace')` can lead to data loss if information doesn't look like binary, but can't be decoded using utf8. It also means that client may occasionally get a result which is not base64-encoded (if value which is usually binary only contain ascii letters) - it means client will have to always handle both base64 and raw cases. Currently it is enough to handle base64 case.

splash/tests/test_har.py
```diff
@@ -138,13 +138,13 @@ def test_meta_redirect_delay_wait_enough(self):
...
-            (self.mockurl('slow.gif?n=2'), 200),
-            (self.mockurl('meta-redirect-target/'), 200),
-        ])
+    # def test_meta_redirect_slowload2_wait_more(self):
+    #     data = self.assertValidHar(self.mockurl('meta-redirect-slowload2'), wait=0.3)
+    #     self.assertRequestedUrlsStatuses(data, [
+    #         (self.mockurl('meta-redirect-slowload2'), 200),
+    #         (self.mockurl('slow.gif?n=2'), 200),
+    #         (self.mockurl('meta-redirect-target/'), 200),
+    #     ])
```
kmike: Yeah, these tests fail sometimes, but usually restarting them helps. Could you please bring it back?

splash/defaults.py
```diff
@@ -9,7 +9,7 @@
 
 # Default size of browser window.  As there're no decorations, this affects
 # both "window.inner*" and "window.outer*" values.
-VIEWPORT_SIZE = '1024x768'
+VIEWPORT_SIZE = '1920x1080'
```
kmike: The advantage of a smaller value is that you'll get a smaller screenshot by default. Also, 1920x1080 is not the most popular size - 1366x768 looks more popular. What do you think about changing it to 1366x768? It should be done in another PR: this is a backwards-incompatible change, and it'd be good to hear other's opinions.

splash/tests/test_har.py
shaylevi2: It won't work. I ran it a lot of times, it simply doesn't work, if you can figure out how to fix it, then please do. If I'll uncomment it, the tests will fail.

kmike:
Hey @shaylevi2 - great job, it looks like you've found a way to preserve the contents, that's a wanted feature! 

I left a few comments to the PR. 

Another issue is that after this change we'll be keeping all data in memory twice. This is 100% fine, but I think it should be optional - there should be a way to turn content storage off:

* endpoints which don't use HAR shouldn't store the content, for efficiency - this includes render.html, render.png, render.jpeg and render.json without 'HAR' key. 
* it should be possible to turn it off per-request in on_request or maybe on_requst_headers callback - I think we should add a method or attribute [Request](http://splash.readthedocs.io/en/stable/scripting-request-object.html#methods) - something like `request.record_response_body=true` (ideas for a better name?), similar to `request:abort()` - this way it'd be possible to e.g. keep only images in memory twice, not js or css files;
* for existing users this change will cause slowdowns and increased memory usage for all /execute scripts - existing users don't use this feature, so they'll just get an overhead. What do you think about making it opt-in? In addition to `request.record_response_body` we can add something like `splash.response_body_enabled` which is false by default, but can be turned ON in a script.
* for render.har and render.json there should be an option to disable content storage;
* this feature needs tests.

Flash:

* I think PLUGINS_ENABLED should be customizable at runtime, not only at start time - we should provide an option for HTTP API and a `splash.plugins_enabled` property.
* it'd be nice to have a test for Flash - for example, check that screenshot contains flash contents.

splash/har/qt.py
shaylevi2: Giving a HAR file filled with only base64 is not the result you'd expect from a HAR file.
I think we should try to imitate chrome's behaviour with HAR files.

If you have a better way to differentiate between text and non-text, please do.
I found the above giving the most satisfying results.

kmike:
I realize this is a lot more work. Some of these features are nice to have but not critical though. I think the following is required:

* tests for both features; 
* encodings fix; 
* content storage should be disabled by default for all endpoints, but there should be an option to turn it ON at runtime in /execute endpoint via splash.response_body_enabled (a better name ideas?).

splash/har/qt.py
kmike: According to specs using base64 everywhere is fine. There are 2 issues with this auto-detection: 1) data loss and 2) more code is required at client. Base64 encoding solves both, even if it is not as nice. I think Chrome may be decoding data based on Content-Type headers, but I'm not sure. We should really get this right or leave base64 as-is; data loss is not acceptable. 

What's your use case? If it is machine which processes HAR files then using base64 everywhere simplifies code - you don't have to handle a case with non-base64 encoding.

shaylevi2:
1) Unfortunately, the organization I'm working in decided not to give me any more time to work on the PR, as we're in a hurry to get a product out, writing further tests is being postponed for now (in our fork and generally).

2) For now, the decoding works perfect for us, it's the only decoding we found that works as expected (i.e giving the same results as Chrome Developer Tools would). Other decoding methods we tried, either left too much as base64, or did not decode the text properly.

3) I agree it would be better to use it as a parameter, unfortunately as said previously, I'm not given any time to work on it.

If anyone wants to advance this PR further, do feel free to do so.

kmike:
Ok, thanks @shaylevi2 for your time!

shaylevi2:
@kmike 

Small update: I fixed the bug with the slow_wait, now all tests pass.

shaylevi2:
Small update: added "move_to_end" monkey patch for python2.7 and now forcing 2.7 tests to pass.

coeusdata:
Is this going to be merged?  Would love HAR content support.

kmike:
Thanks @shaylevi2! Merged in #494.

PR #456, opened 2016-05-03
splash:send_keys: support basic edmacro format

immerrr:
This is an attempt to support some basic subset of edmacro format in `splash:send_keys`.


eskerda:
Sounds good to me. Thanks!

PR #465, opened 2016-06-07
Add `with_timeout` API to Lua scripting

mike1808:


splash/lua_runner.py
```diff
@@ -66,6 +67,9 @@ def start(self, coro_func, coro_args=None):
         self._waiting_for_result_id = self._START_CMD
         self.dispatch(self._waiting_for_result_id)
 
+    def stop(self):
+        self._is_stopped = True
```
kmike: looks good 👍

splash/lua_runner.py
```diff
@@ -102,6 +106,9 @@ def dispatch(self, cmd_id, *args):
             try:
                 args = args or None
 
+                if self._is_stopped:
+                    raise StopIteration
```
kmike: I'm not 100% sure this works correctly, or at least the code is not clear: the exception is caught later, and BaseScriptRunner tries to return the result by calling `self.on_result(res)`, where `res` is `lua2python(self.result)`. This looks unnecessary, and it is kind of hard to grasp what happens if `on_result` is called earlier, or if `self.result` is guaranteed to be None or if it can contain something else.

What do you think about raising another exception, and handling it explicitly? Or maybe just add a bit of comments to the source code which explain why does it work.

splash/qtrender_lua.py
```diff
@@ -1095,8 +1168,11 @@ def get_wrapped(self):
 
     def run_async_command(self, cmd):
         """ Execute _AsyncBrowserCommand """
```
kmike: the docstring is outdated

splash/qtrender_lua.py
```diff
@@ -1095,8 +1168,11 @@ def get_wrapped(self):
...
-        meth = getattr(self.tab, cmd.name)
-        return meth(**cmd.kwargs)
+        if isinstance(cmd, AsyncBrowserCommand):
+            meth = getattr(self.tab, cmd.name)
+            return meth(**cmd.kwargs)
+        elif isinstance(cmd, AsyncLuaCommand):
+            cmd.kwargs["callback"]()
```
kmike: What do you think about renaming it to AsyncCallbackCommand? For AsyncLuaCommand I'd expect callback to be a Lua callback, but here it is an arbitrary Python function

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,30 @@ Trigger mouse release event in web page.
...
+**Signature:** ``ok, result = splash:with_timeout(callback, timeout)``
+
+**Parameters:**
+
+* callback - the function to run
+* timeout - timeout, in seconds
+
+**Returns:** ``ok, result`` pair. If ``ok`` is nil then error happened during
+the function call or the timeout has passed; ``result`` provides an information
+about error type. Otherwise, if ``ok`` is not nil then ``result`` contains the
```
kmike: Why is it `"if ok is not nil"` and not `"if ok is true"`?

splash/qtrender_lua.py
```diff
@@ -487,6 +491,75 @@ def error(error_info):
         ))
         return cmd
 
+    @command(async=True, sets_callback=True, decode_arguments=False)
```
kmike: hm, does sets_callback really work here? I recall it requires the first argument to be a callback

mike1808: You are right. However, I've tried to set it False and nothing has changed (or maybe I don't checked some edge case).

docs/scripting-ref.rst
mike1808: I looked how the documentation of `splash:go` was written and thought that there are some cases where `ok` can be `nil`. However, I've checked the code, ok can be only `true` or `false`.

mike1808: Also, I don't understand why in some places the `ok` is `nil` (`None` in Python) and in some places `false`. Is it a way to distinguish between exceptions and "failed" operations?

splash/qtrender_lua.py
```diff
@@ -487,6 +491,75 @@ def error(error_info):
...
+
+            # TODO: make error more verbose (e.g. add line number)
+            cmd.return_result(False, str(ex.args[0]))
+
+        run_coro = self.get_coroutine_run_func(
+            "splash:with_timeout", callback, coro_success, coro_error)
+
+        def start():
+            qtimer.start(timeout)
+            run_coro.runner = run_coro()
```
kmike: Passing a variable as a new attribute looks a bit odd; there can be other ways to write this code (e.g. define a callback inside this function and pass variable in a closure), but I'm not sure they are better though. But if you have ideas on how to clean it up then please do it :)

docs/scripting-ref.rst
kmike: > I don't understand why in some places the ok is nil

hm, I'm not sure this is intentional

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,30 @@ Trigger mouse release event in web page.
...
+result of the executed function.
+
+**Async:** yes.
```
kmike: Is it async (i.e. does it return control to the event loop) if a callback is not async?

mike1808: Yes, it does. That's why I've created `AsyncLuaCommand` and [this](https://github.com/scrapinghub/splash/pull/465/files#diff-f165607d210fc68f95a1e92d9aa15a3bR1175)

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,30 @@ Trigger mouse release event in web page.
...
+
+
+.. _splash-with-timeout:
+
+splash:with_timeout
+-------------------
+
+Run the callback with the allowed timeout
+
+**Signature:** ``ok, result = splash:with_timeout(callback, timeout)``
```
kmike: signture doesn't match the code example

splash/lua_runner.py
mike1808: self.result cannot be anything except an empty string, so calling it is harmless

splash/qtrender_lua.py
mike1808: I see only the one solution - create a class and return an instance when `get_coroutine_run_func` is called.

docs/scripting-ref.rst
mike1808: Now it matches.

splash/qtrender_lua.py
```diff
@@ -487,6 +491,76 @@ def error(error_info):
...
+
+        def coro_error(ex):
+            if not qtimer.isActive():
+                return
+
+            qtimer.stop()
+
+            info = str(ex.args[0]["error"])
+
+            cmd.return_result(None, info)
```
kmike: a nitpick: could you please remove blank lines between two previous commands? PEP8 [says](https://www.python.org/dev/peps/pep-0008/#blank-lines) "Use blank lines in functions, *sparingly*, to indicate logical sections. "

mike1808: Thanks for the note.

kmike:
The tests look good 👍, but some cases are missing: 

* what happnens when function returns a value? 
* what if function returns several values?
* does `with_timeout` exit earlier if a callback finishes faster than timeout, or does it wait for the whole timeout? You can test it by using a small global script timeout ([timeout](http://splash.readthedocs.io/en/stable/api.html#arg-timeout) argument) and a large with_timeout timeout.
* there are some missing lines here: https://codecov.io/gh/scrapinghub/splash/compare/ad656e3ee22f7b3161279777cb7cc0d12a47d786...40af8ff826af5d4f0eca6ea5246325ed620dbb4f. Timeout validation looks easy to test, and it can make codecov happy. But I'm more concerned about if `not qtimer.isActive(): return` - when is it executed?

mike1808:
> But I'm more concerned about if not qtimer.isActive(): return - when is it executed?
It happens when the timeout is elapsed but the callback didn't finished its job. In this case, `StopIterration` exceptions is raised which call `send_results` passing an empty string, after one of those callbacks are called. In that case `qtimer.isActive()` will return `False`.

kmike:
@mike1808 aha, makes sense. Could you please add a test for this?

splash/lua_modules/wraputils.lua
```diff
@@ -74,6 +74,40 @@ local function raises_async(func)
   end
 end
 
+function print_r ( t )
```
kmike: there is inspect.lua bundled (see https://github.com/scrapinghub/splash/blob/master/splash/lua_modules/vendor/inspect.lua), and a wrapper for it: https://github.com/scrapinghub/splash/blob/master/splash/lua_modules/repr.lua, maybe they can help

docs/scripting-ref.rst
immerrr: It's better for the eyes to have timeout first, so that you don't have to squint looking for the number somewhere after the callback definition. It would also match the Lua de-facto style convention neatly with 
```
with_callback(0.1, function()
   do_smth()
end)
```

Is there a precedent in non-Lua world calling for placing the timeout first?

kmike: Yeah, it is easier for my eyes as well. But there are some stupid tech difficulties in having timeout first ([sets_callback](https://github.com/scrapinghub/splash/blob/ad656e3ee22f7b3161279777cb7cc0d12a47d786/splash/lua_modules/wraputils.lua#L100) uses first argument), so for `call_later` I made timeout the last argument. It is also consistent with  JavaScript functions. In Python it is more common to have timeout a first argument, e.g. this is the case in asyncio and twisted.

splash/qtrender_lua.py
```diff
@@ -487,6 +493,75 @@ def error(error_info):
         ))
         return cmd
 
+    @command(async=True, sets_callback=True, decode_arguments=False, pack_results=True)
+    def with_timeout(self, callback, timeout):
```
immerrr: Is the passed function really a `callback`?

mike1808: It's more `handler` rather than `callback`.

immerrr: Maybe we should call it differently everywhere then? Lua manual takes no stance as to how it should be called, naming it just `f`: [f must be a function](https://www.lua.org/manual/5.3/manual.html#pdf-coroutine.create). Maybe we should be a little more expressive about that and call it `func`?

mike1808: `func` is the nice name. I will rename it.

splash/lua_modules/wraputils.lua
```diff
@@ -109,6 +109,31 @@ local function sets_callback(func, storage)
...
+
+    if type(cb) == 'function' then
+      mcb = function(...)
+        return { cb(...) }
+      end
+    end
+
+    return func(mcb, ...)
+  end
+end
```
immerrr: Why is it necessary to pack & unpack it here? If we're returning an error, it's a single value and shouldn't be wrapped, otherwise the values originate from user-specified code chunk, shouldn't they pass through our executor intact?

mike1808: The reason is why we pack/inpuck in Lua is that if try to return several values from the passed function (callback) it will be converted to the table, because Python cannot know whether it got several values or a table. Then it passed back to Lua [here](https://github.com/mike1808/splash/blob/014b6f7c0d63ee24eede20ca7eb9289adab9a928/splash/browser_tab.py#L654)

immerrr: > Python cannot know whether it got several values or a table

Could you point me to where the automatical transformation is happening, because i see them easily discerned:
```
In [49]: list(lupa.LuaRuntime().execute('c = require "coroutine"; coro = c.create(function() c.yield(1,2); c.yield{3,4}; return 5,6,{7,8}; end); return coro').coroutine())
Out[49]: [(1, 2), <Lua table at 0x2cc3f60>, (5, 6, <Lua table at 0x2cbf520>)]

```

There's a problem when you return a tuple from python code to lua runtime, that's where it may be tricky to decide whether or not it should be unpacked, but the other way around there should be no problem.

mike1808: Here is that part
https://github.com/scrapinghub/splash/blob/master/splash/lua_runner.py#L176

docs/scripting-ref.rst
```diff
@@ -2230,3 +2230,59 @@ Trigger mouse release event in web page.
...
+**Async:** yes.
+
+Example 1:
+
+.. literalinclude:: ../splash/examples/with-timeout.lua
+   :language: lua
+
+
+Note that Splash can interrupt the running function only if some async function is called.
+Otherwise you should wait until the operation is finished.
```
immerrr: This paragraph might need to talk specifically about the fact that "timeout event can interrupt". And how one makes a function "async" without using splash methods by using `coro.yield`. It also might help to have asyncio-like comments in the code snippets saying "timeout event can fire here".

splash/lua_modules/wraputils.lua
immerrr: I think we can find a smarter way of doing it automatically, but don't have anything specific to back my claim. Let's keep it for now then.

splash/qtrender_lua.py
```diff
@@ -1094,9 +1169,12 @@ def get_wrapped(self):
...
     def run_async_command(self, cmd):
-        """ Execute _AsyncBrowserCommand """
-        meth = getattr(self.tab, cmd.name)
-        return meth(**cmd.kwargs)
+        """ Execute _AsyncBrowserCommand or _AsyncCallbackCommand"""
+        if isinstance(cmd, AsyncBrowserCommand):
+            meth = getattr(self.tab, cmd.name)
+            return meth(**cmd.kwargs)
+        elif isinstance(cmd, AsyncFunctionCommand):
+            cmd.kwargs["callback"]()
```
immerrr: is it intentional that callback retval is ignored here? is it intentional that the parameter is still called "callback" here?

should we put a `raise TypeError` at the end of the function itself to catch unhandled cases?

mike1808: Good catch. Usually async functions doesn't return any value and even if they do it isn't handled anywhere. 
I will return the `callback` value to make it look like the previous handler.

mike1808:
There is one "red" [line](https://github.com/scrapinghub/splash/pull/465/files#diff-f165607d210fc68f95a1e92d9aa15a3bR545) in codecov-io that I cannot cover. I didn't find any case where this line can be executed.

immerrr:
LGTM!

immerrr:
Merged! Thank you!

mike1808:
Nice, thank you!

PR #469, opened 2016-06-16
minor code style

dvska:


dockerfiles/splash/provision.sh
```diff
@@ -215,7 +215,7 @@ for cmd in "$@"; do
     fi
 done
 
-if [ $UNKNOWN -eq 1 ]; then
+if [ ${UNKNOWN} -eq 1 ]; then
```
kmike: 👍

splash/tests/mockserver.py
```diff
@@ -764,7 +764,7 @@ class Image(Resource):
         @use_chunked_encoding
         def render_GET(self, request):
             request.setHeader(b"Content-Type", b"image/gif")
-            return base64.decodestring(b'R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=')
+            return base64.decodebytes(b'R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=')
```
kmike: We're still trying to support Python 2.x; it looks like in 2.x there is not base64.decodebytes method.

kmike:
Thanks for the cleanup, and sorry it took so long to review! Most changes are good, but I'm not sure I like `@staticmethod` changes: I don't see benefits, and it is more code.

PR #471, opened 2016-06-22
[WIP] Add Element class for manipulating DOM elements in Lua

mike1808:
some test lua code

```lua
function main(splash)
  assert(splash:go('http://localhost:8000/'))

  local element = splash:select('button')
  element:mouse_click()
  assert(splash:wait(0.0001))

  local title = splash:select('#title')

  return {title:get_styles()["display"], title:node_property('nodeName')}
end
```
for the following html
```html
<!DOCTYPE html>
<html>
<head>
   <title>Index 1</title>
</head>
<body>
<button type="button" onclick="document.getElementById('title').style.display = 'block'">Click me</button>
<h1 id="title" style="display: none">Index 1</h1>
</body>
</html>
```

### DIFF
--- a/docs/scripting-element-object.rst
+++ b/docs/scripting-element-object.rst
@@ -1,0 +1,395 @@
+.. _splash-element:
+
+Element Object
+==============
+
+Element objects are instanced by :ref:`splash-select`;
+
+.. _splash-element-attributes:
+
+Attributes
+~~~~~~~~~~
+
+The following fields are read-only.
+
+.. _splash-element-id:
+
+element.id
+----------
+
+Id of the inner representation of the element.
+
+Methods
+~~~~~~~
+
+To modify or retrieve some information about the element you can use the following methods.
+Note that all of the following methods return flag indicating whether the operation was
+successful or not. The methods returns ``ok, reason_or_value`` pair. If ``ok`` is nil
+then error happened during the operation; ``reason`` provides an information about error type;
+otherwise ``ok`` is ``true`` and the returned value is stored in the second variable.
+
+
+.. _splash-element-exists:
+
+element:exists
+--------------
+
+Check whether the element exists in DOM. If the element doesn't exist almost all other
+methods will fail.
+
+**Signature:** ``ok, exists = element:exists()``
+
+**Returns:** ``ok, exists`` pair. If ``ok`` is nil then error happened during the check;
+``exists`` provides an information about error type; otherwise ``exists`` indicated whether
+the element exists.
+
+**Async:** no.
+
+
+.. _splash-node-property:
+
+element:node_property
+---------------------
+
+Return the value of the specified property of the element
+
+**Signature:** ``ok, value = element:node_property(property_name)``
+
+**Parameters:**
+
+* property_name - name of the node property
+
+**Returns:** ``ok, value`` pair. If ``ok`` is nil then error happened during the function
+call; ``value`` provides an information about error type; otherwise ``value`` is a value of
+the specified property
+
+**Async:** no.
+
+
+.. _splash-node-method:
+
+element:node_method
+-------------------
+
+Return the callable Lua function which call the specified node method.
+
+**Signature:** ``ok, method = element:node_method(method_name)``
+
+**Parameters:**
+
+* method_name - name of the node method
+
+**Returns:** ``ok, method`` pair. If ``ok`` is nil then error happened during the function
+call; ``method`` provides an information about error type; otherwise ``value`` is a function
+that can be called from Lua to execute node method in page context.
+
+**Async:** no.
+
+See more in :ref:`splash-jsfunc`.
+
+.. note::
+
+    If the specified method returns another HTML element it will be represented as a table with
+    ``type`` property which equals to ``'node'``
+
+
+.. _splash-element-mouse-click:
+
+element:mouse_click
+-------------------
+
+Trigger mouse click event on the top-left corner of the element.
+
+**Signature:** ``ok, reason = element:mouse_click()``
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
+function call; ``reason`` provides an information about error type.
+
+**Async:** no.
+
+See more about mouse events in :ref:`splash-mouse-click`.
+
+
+.. _splash-element-mouse-hover:
+
+element:mouse_click
+-------------------
+
+Trigger mouse hover (JavaScript mouseover) event on the top-left corner of the element.
+
+**Signature:** ``ok, reason = element:mouse_hover()``
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
+function call; ``reason`` provides an information about error type.
+
+**Async:** no.
+
+See more about mouse events in :ref:`splash-mouse-click`.
+
+
+.. _splash-element-get-styles:
+
+element:get_styles
+------------------
+
+Return the computed styles of the element.
+
+**Signature:** ``ok, styles = element:get_styles()``
+
+**Returns:** ``ok, styles`` pair. If ``ok`` is nil then error happened during the
+function call; ``styles`` provides an information about error type; otherwise
+``styles`` is a table with computed styles of the element.
+
+**Async:** no.
+
+Example of getting the font size of the element.
+
+.. code-block:: lua
+
+    function main(splash)
+        local element = splash:select('.element')
+        local ok, styles = assert(element:get_styles())
+
+        return styles['font-size']
+    end
+
+
+.. _splash-element-get-bounds:
+
+element:get_bounds
+------------------
+
+Return the bounding client rectangle of the element
+
+**Signature:** ``ok, styles = element:get_bounds()``
+
+**Returns:** ``ok, bounds`` pair. If ``ok`` is nil then error happened during the
+function call; ``bounds`` provides an information about error type; otherwise
+``bounds`` is a table with the client bounding rectangle with the ``top``, ``right``,
+``bottom`` and ``left`` coordinates.
+
+**Async:** no.
+
+Example of getting the bounds of the element.
+
+.. code-block:: lua
+
+    function main(splash)
+        local element = splash:select('.element')
+        local ok, bounds = assert(element:get_bounds())
+        -- e.g. bounds is { top = 10, right = 20, bottom = 20, left = 10 }
+        return bounds
+    end
+
+
+.. _splash-element-png:
+
+element:png
+-----------
+
+Return a screenshot of the element in PNG format
+
+**Signature:** ``ok, shot = element:png{width=nil, height=nil, scale_method='raster'}``
+
+**Parameters:**
+
+* width - optional, width of a screenshot in pixels;
+* height - optional, height of a screenshot in pixels;
+* scale_method - optional, method to use when resizing the image, ``'raster'``
+  or ``'vector'``;
+
+**Returns:** ``ok, shot`` pair. If ``ok`` is nil then error happened during the
+function call; ``shot`` provides an information about error type; otherwise
+``shot`` is a PNG screenshot data, as a :ref:`binary object <binary-objects>`.
+When the result is empty (e.g. if the element is not visible) ``nil`` is returned.
+
+**Async:** no.
+
+See more in :ref:`splash-png`.
+
+
+
+.. _splash-element-jpeg:
+
+element:jpeg
+------------
+
+Return a screenshot of the element in JPEG format
+
+**Signature:** ``ok, shot = element:jpeg{width=nil, height=nil, scale_method='raster', quality=75, region=nil}``
+
+**Parameters:**
+
+* width - optional, width of a screenshot in pixels;
+* height - optional, height of a screenshot in pixels;
+* scale_method - optional, method to use when resizing the image, ``'raster'``
+  or ``'vector'``;
+* quality - optional, quality of JPEG image, integer in range from ``0`` to ``100``;
+
+**Returns:** ``ok, shot`` pair. If ``ok`` is nil then error happened during the
+function call; ``shot`` provides an information about error type; otherwise
+``shot`` is a JPEG screenshot data, as a :ref:`binary object <binary-objects>`.
+When the result is empty (e.g. if the element is not visible) ``nil`` is returned.
+
+**Async:** no.
+
+See more in :ref:`splash-jpeg`.
+
+
+.. _splash-element-visible:
+
+element:visible
+---------------
+
+Check whether the element is visible.
+
+**Signature:** ``ok, visible = element:visible()``
+
+**Returns:** ``ok, visible`` pair. If ``ok`` is nil then error happened during the function
+call; ``visible`` provides an information about error type; otherwise ``visible`` indicated whether
+the element is visible.
+
+**Async:** no.
+
+
+.. _splash-element-fetch-text:
+
+element:fetch_text
+------------------
+
+Fetch a text information from the element
+
+**Signature:** ``ok, visible = element:fetch_text()``
+
+**Returns:** ``ok, text`` pair. If ``ok`` is nil then error happened during the function call;
+``text`` provides an information about error type; otherwise ``text`` is a text content
+of the element.
+
+**Async:** no.
+
+It tries to return the value of the following JavaScript ``Node`` properties:
+
+* textContent
+* innerText
+* value
+
+If all of them are empty an empty string is returned.
+
+
+.. _splash-element-info:
+
+element:info
+------------
+
+Get useful information about the element.
+
+**Signature:** ``ok, info = element:info()``
+
+**Returns:** ``ok, info`` pair. If ``ok`` is nil then error happened during the function call;
+``info`` provides an information about error type; otherwise ``info`` is a table with info.
+
+**Async:** no.
+
+Info is a table with the following fields:
+
+* nodeName - node name in a lower case (e.g. *h1*)
+* attributes - table with attributes names and its values
+* tag - html string representation of the element
+* html - inner html of the element
+* text - inner text of the element
+* x - x coordinate of the element
+* y - y coordinate of the element
+* width - width of the element
+* height - height of the element
+* visible - flag representing if the element is visible
+
+
+.. _splash-field-value:
+
+element:field_value
+-------------------
+
+Get value of the field element (input, select).
+
+**Signature:** ``ok, info = element:field_value()``
+
+**Returns:** ``ok, value`` pair. If ``ok`` is nil then error happened during the function call;
+``value`` provides an information about error type; otherwise ``value`` is a value of the
+element.
+
+**Async:** no.
+
+The value can be a
+* string - for text/radio input and select and other element types
+* array of strings - for multi select
+* boolean - for checkbox input
+
+
+.. _splash-form-values:
+
+element:form_values
+-------------------
+
+Return a table with form values if the element type is *form*
+
+**Signature:** ``ok, info = element:form_values()``
+
+**Returns:** ``ok, values`` pair. If ``ok`` is nil then error happened during the function call
+or node type is not *form*; ``values`` provides an information about error type; otherwise
+``values`` is a table of values.
+
+**Async:** no.
+
+
+.. _splash-send-keys:
+
+element:send_keys
+-----------------
+
+Send keyboard events to the element.
+
+**Signature:** ``ok = element:send_keys(keys)``
+
+**Parameters**
+
+* keys - string representing the keys to be sent as keyboard events.
+
+**Returns:** ``ok`` pair. If ``ok`` is nil then error happened during the function call.
+
+**Async:** no.
+
+This methods do the following:
+
+* checks whether the selected element is editable
+* clicks on the element
+* send the specified keyboard events
+
+See more about keyboard events in in :ref:`splash-send-keys`.
+
+
+.. _splash-send-text:
+
+element:send_text
+-----------------
+
+Send keyboard events to the element.
+
+**Signature:** ``ok = element:send_text(text)``
+
+**Parameters**
+
+* text - string to be sent as input.
+
+**Returns:** ``ok`` pair. If ``ok`` is nil then error happened during the function call.
+
+**Async:** no.
+
+This methods do the following:
+
+* checks whether the selected element is editable
+* clicks on the element
+* send the specified text to the element
+
+See more about it in :ref:`splash-send-text`.
+
+
--- a/docs/scripting-element-object.rst
+++ b/docs/scripting-element-object.rst
@@ -37,11 +37,9 @@ element:exists
 Check whether the element exists in DOM. If the element doesn't exist almost all other
 methods will fail.
 
-**Signature:** ``ok, exists = element:exists()``
+**Signature:** ``exists = element:exists()``
 
-**Returns:** ``ok, exists`` pair. If ``ok`` is nil then error happened during the check;
-``exists`` provides an information about error type; otherwise ``exists`` indicated whether
-the element exists.
+**Returns:** ``exists`` indicated whether the element exists.
 
 **Async:** no.
 
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -2436,3 +2436,21 @@ arguments passed to splash, `username` and `password`.
         -- Usually, wait for the submit request to finish
         -- ...
     end
+
+
+.. _splash-select:
+
+
+splash:select
+-------------
+
+Select HTML element from DOM of the current web page
+
+**Signature:** ``splash:select(selector)``
+
+**Parameters:**
+
+* selector - valid CSS selector
+
+**Returns:** a :ref:`splash-element`.
+
--- a/splash/browser_tab.py
+++ b/splash/browser_tab.py
@@ -31,8 +31,9 @@ from splash.jsutils import (
     SANITIZE_FUNC_JS,
     get_process_errors_js,
     escape_js,
+    store_dom_elements,
 )
-
+from splash.html_element import HTMLElement
 
 def skip_if_closing(meth):
     @functools.wraps(meth)
@@ -89,6 +90,7 @@ class BrowserTab(QObject):
 
         self._set_default_webpage_options(self.web_page)
         self._setup_webpage_events()
+        self._init_elements_storage()
 
         self.web_view = SplashQWebView()
         self.web_view.setPage(self.web_page)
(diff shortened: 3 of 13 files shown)


splash/browser_tab.py
```diff
@@ -1078,6 +1093,65 @@ def log(self, message, min_level=None):
...
+
+    @pyqtSlot(name="getId", result=str)
+    def get_id(self):
+        return str(uuid.uuid1())
+
+
+class HTMLElement(object):
+    DIMENSIONS_JS_FUNC = """
+        (function(element) {
+            var rect = element.getBoundingClientRect();
```
Youwotma: better `getClientRects()[0]`

splash/browser_tab.py
```diff
@@ -1078,6 +1093,65 @@ def log(self, message, min_level=None):
...
+        except JsError:
+            raise RuntimeError("Element no longer exists in DOM")
+
+    def node_property(self, property_name):
+        self.assert_element_exists()
+        return self.tab.evaljs("%s[%s]" % (self.element_js, escape_js(property_name)))
+
+    def node_method(self, method_name):
+        self.assert_element_exists()
+        return self.tab.evaljs("%s[%s]()" % (self.element_js, escape_js(method_name)))
```
Youwotma: Should we use `jsfunc` and return the function, so the user can call it with whatever arguments they want? You might need to bind the element to the context or wrap the function for it to work properly.

splash/browser_tab.py
mike1808: Why it is better?

Youwotma: Consider this link: screenshoot

The blue semi-transparent rectangles are the client rects (it has two), the dotted line the bounding client rect (the rect that contains the client rects).

The top left corner of the bounding client rect is not "on" the element, the top-left of one of the client rects is.

splash/browser_tab.py
mike1808: You are right. I'll change it.

splash/html_element.py
```diff
@@ -0,0 +1,222 @@
...
+
+        self.tab.mouse_hover(dimensions["x"], dimensions["y"])
+
+    def get_styles(self):
+        self.assert_element_exists()
+        return self.tab.evaljs("getComputedStyle(%s)" % self.element_js, False, False)
+
+    def get_bounds(self):
+        self.assert_element_exists()
+        return self.tab.evaljs("%s.getClientRects()[0]" % self.element_js, result_protection=False)
```
immerrr: This getClientRects/getBoundingClientRect debate makes me wonder:
- if we want to expose getBoundingClientRect to the users
- if we want to expose getClientRects returning an array to the users
- if we want to capture (png) all client rects or the bounding client rect to the users

splash/browser_tab.py
immerrr: That might be a problem if you want to "click" the element. If you want to take a screenshot, you most likely won't be satisfied with just the first client rect.

Youwotma: Yes, exactly, both of the methods exist for a reason. It might also be
worth it to perform a hittest on the coordinates and raise an error if its
not successful. The element might be hidden behind another element or
transparent in that coordinates, so raising an error would be better than
failing silently and possibly causing side effects.

immerrr: True, but I still think that for getting "dimensions" the bounding variant is better. For clicking or interacting with it somehow, probably not, but this method is used later on for capturing the elements and like I said, the first client rect is most certainly not what you'd expect when you'd do `element.png()`.

immerrr: As a sidenote, @mike1808, would it be useful to return `nil` from `png` method if the element is not visible?

mike1808: @immerrr it sounds reasonable.

mike1808:
@Youwotma @immerrr I've changed the method of getting boundary rectangle for `png` and `jpeg` methods to `getBoundaryRectangle`. Now those methods return `None` if element is not visible.

Also I moved all used Casper.JS methods to separate file and added MIT license to them. 

@Youwotma check please the `node_method` both in Lua and Python. I have 2 different implementation for Python API and Lua.

splash/qtrender_lua.py
```diff
@@ -328,7 +329,7 @@ def __call__(self, *args):
             args=escape_js(*args)
         )
         wrapper_script = get_process_errors_js(expr)
-        res = self.tab.evaljs(wrapper_script)
+        res = self.tab.evaljs(wrapper_script, result_protection=self.js_result_protection)
```
mike1808: If you turn on `result_protection` some objects cannot be retreived, for example `getBoundaryRect()` returns an instance of [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) which is ignored in sanitize function.

kmike: I think the way to go is to wrap objects like DOMRect explicitly.

mike1808: @kmike what do you mean? Should new rules be added to [sanitize](https://github.com/scrapinghub/splash/blob/master/splash/jsutils.py#L12) function? Or it should be written like
```js
var rect = elem. getBoundingClientRect();
return { top: rect.x, right: rect.right, bottom: rect.bottom, left: rect.left };
```

kmike: I think both these options are fine. Adding it to sanitize may be a bit more convenient for users because they won't have to write `{ top: rect.x, right: rect.right, bottom: rect.bottom, left: rect.left }` themselves. A third option is to convert DOMRect to a special DOMRect Lua object - just add a few useful methods to `{ top=rect.x, right=rect.right, bottom=rect.bottom, left=rect.left }` table.

mike1808:
Fixed bugs, rebased code, added tests. Some of them fails because of #488 
This patch fixes that but I didn't want to include it in this PR:
```
From d9b0204eaa74d8836cd9ae66567836b5f737a34a Mon Sep 17 00:00:00 2001
From: Michael Manukyan <arm.localhost@gmail.com>
Date: Tue, 26 Jul 2016 20:39:25 +0400
Subject: [PATCH] store private methods in the instance itself rather than in
 the separate table

---
 splash/lua_modules/extras.lua    |  3 +--
 splash/lua_modules/request.lua   |  3 +--
 splash/lua_modules/response.lua  |  3 +--
 splash/lua_modules/splash.lua    | 22 ++++++++++------------
 splash/lua_modules/wraputils.lua | 16 +++++++++-------
 5 files changed, 22 insertions(+), 25 deletions(-)

diff --git a/splash/lua_modules/extras.lua b/splash/lua_modules/extras.lua
index 61bf6fa..8a0e0f1 100644
--- a/splash/lua_modules/extras.lua
+++ b/splash/lua_modules/extras.lua
@@ -6,12 +6,11 @@
 local wraputils = require("wraputils")
 
 local Extras = {}
-local Extras_private = {}
 wraputils.set_metamethods(Extras)
 
 function Extras._create(py_extras)
   local extras = {}
-  wraputils.wrap_exposed_object(py_extras, extras, Extras, Extras_private, false)
+  wraputils.wrap_exposed_object(py_extras, extras, Extras)
   return extras
 end
 
diff --git a/splash/lua_modules/request.lua b/splash/lua_modules/request.lua
index 79d5783..00caeb8 100644
--- a/splash/lua_modules/request.lua
+++ b/splash/lua_modules/request.lua
@@ -5,7 +5,6 @@ local wraputils = require("wraputils")
 local treat = require("libs/treat")
 
 local Request = wraputils.create_metatable()
-local Request_private = {}
 wraputils.set_metamethods(Request)
 
 function Request._create(py_request)
@@ -16,7 +15,7 @@ function Request._create(py_request)
     method=py_request.method,
   }
 
-  wraputils.wrap_exposed_object(py_request, request, Request, Request_private, false)
+  wraputils.wrap_exposed_object(py_request, request, Request)
   return request
 end
 
diff --git a/splash/lua_modules/response.lua b/splash/lua_modules/response.lua
index 65ac2d8..d8c44d8 100644
--- a/splash/lua_modules/response.lua
+++ b/splash/lua_modules/response.lua
@@ -6,7 +6,6 @@ local treat = require("libs/treat")
 local Request = require("request")
 
 local Response = wraputils.create_metatable()
-local Response_private = {}
 wraputils.set_metamethods(Response)
 
 function Response._create(py_response)
@@ -15,7 +14,7 @@ function Response._create(py_response)
     request=Request._create(py_response.request),
   }
 
-  wraputils.wrap_exposed_object(py_response, response, Response, Response_private, false)
+  wraputils.wrap_exposed_object(py_response, response, Response)
   return response
 end
 
diff --git a/splash/lua_modules/splash.lua b/splash/lua_modules/splash.lua
index 2598ead..933992f 100644
--- a/splash/lua_modules/splash.lua
+++ b/splash/lua_modules/splash.lua
@@ -12,12 +12,11 @@ local repr = require("repr")
 -- Lua wrapper for Splash Python object.
 --
 local Splash = wraputils.create_metatable()
-local Splash_private = {}
 wraputils.set_metamethods(Splash)
 
   local splash = { args = py_splash.args }
-  wraputils.wrap_exposed_object(py_splash, splash, Splash, Splash_private, true)
+  wraputils.wrap_exposed_object(py_splash, splash, Splash)
   return splash
 end
 
@@ -26,7 +25,7 @@ end
 -- It is required to handle errors properly.
 --
 function Splash:jsfunc(...)
-  local func = Splash_private.jsfunc(self, ...)
+  local func = self.__private.jsfunc(self, ...)
   return wraputils.unwraps_python_result(func)
 end
 
@@ -37,7 +36,7 @@ function Splash:on_request(cb)
   if type(cb) ~= 'function' then
     error("splash:on_request callback is not a function", 2)
   end
-  Splash_private.on_request(self, function(py_request)
+  self.__private.on_request(self, function(py_request)
     local req = Request._create(py_request)
     return cb(req)
   end)
@@ -47,7 +46,7 @@ function Splash:on_response_headers(cb)
   if type(cb) ~= 'function' then
     error("splash:on_response_headers callback is not a function", 2)
   end
-  Splash_private.on_response_headers(self, function(response)
+  self.__private.on_response_headers(self, function(response)
     local res = Response._create(response)
     return cb(res)
   end)
@@ -57,7 +56,7 @@ function Splash:on_response(cb)
   if type(cb) ~= 'function' then
     error("splash:on_response callback is not a function", 2)
   end
-  Splash_private.on_response(self, function(response)
+  self.__private.on_response(self, function(response)
     local res = Response._create(response)
     return cb(res)
   end)
@@ -68,17 +67,16 @@ end
 -- Timer Lua wrapper
 --
 local Timer = wraputils.create_metatable()
-local Timer_private = {}
 wraputils.set_metamethods(Timer)
 
 function Timer._create(py_timer)
   local timer = {}
-  wraputils.wrap_exposed_object(py_timer, timer, Timer, Timer_private, true)
+  wraputils.wrap_exposed_object(py_timer, timer, Timer)
   return timer
 end
 
 function Splash:call_later(cb, delay)
-  local py_timer = Splash_private.call_later(self, cb, delay)
+  local py_timer = self.__private.call_later(self, cb, delay)
   return Timer._create(py_timer)
 end
 
@@ -97,7 +95,7 @@ function Element._create(py_element)
 end
 
 function Element:node_method(...)
-  local ok, func = Element_private.node_method(self, ...)
+  local ok, func = self.__private.node_method(self, ...)
 
   if not ok then
     return ok, func
@@ -108,7 +106,7 @@ end
 
 
 function Element:node_property(...)
-  local ok, result, is_element = Splash_private.node_property(self, ...)
+  local ok, result, is_element = self.__private.node_property(self, ...)
 
   if not ok then
     return ok, result
@@ -123,7 +121,7 @@ end
 
 
 function Splash:select(selector)
-  local py_element = Splash_private.select(self, selector)
+  local py_element = self.__private.select(self, selector)
   return Element._create(py_element)
 end
 
diff --git a/splash/lua_modules/wraputils.lua b/splash/lua_modules/wraputils.lua
index cd8a631..671b916 100644
--- a/splash/lua_modules/wraputils.lua
+++ b/splash/lua_modules/wraputils.lua
@@ -97,10 +97,12 @@ end
 -- * Lua methods are created for Python methods wrapped in @command.
 -- * Async methods are wrapped with `coroutine.yield`.
 -- * Lua <-> Python error handling is fixed.
--- * Private methods are stored in `private_self`, public methods are
+-- * Private methods are stored in `self.__private`, public methods are
 --   stored in `self`.
 --
-local function setup_commands(py_object, self, private_self)
+local function setup_commands(py_object, self)
+  rawset(self, '__private', {})
+
   -- Create lua_object:<...> methods from py_object methods:
   for key, opts in pairs(py_object.commands) do
     local command = py_object[key]
@@ -121,7 +123,7 @@ local function setup_commands(py_object, self, private_self)
 
     if is_private_name(key) then
       local short_key = string.sub(key, PRIVATE_PREFIX:len() + 1)
-      private_self[short_key] = command
+      self.__private[short_key] = command
     else
       -- avoid custom setter
       rawset(self, key, command)
@@ -133,7 +135,7 @@ end
 --
 -- Handle @lua_property decorators.
 --
-local function setup_property_access(py_object, self, cls)
+local function setup_property_access(py_object, self)
   rawset(self, '__getters', {})
   rawset(self, '__setters', {})
 
@@ -153,10 +155,10 @@ end
 --
 -- Create a Lua wrapper for a Python object.
 --
-local function wrap_exposed_object(py_object, self, cls, private_self)
+local function wrap_exposed_object(py_object, self, cls)
   setmetatable(self, cls)
-  setup_commands(py_object, self, private_self)
-  setup_property_access(py_object, self, cls)
+  setup_commands(py_object, self)
+  setup_property_access(py_object, self)
 end
 
 
-- 
2.7.4 (Apple Git-66)
```

mike1808:
@immerrr @kmike I've added docs and test.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,395 @@
...
+--------------
+
+Check whether the element exists in DOM. If the element doesn't exist almost all other
+methods will fail.
+
+**Signature:** ``ok, exists = element:exists()``
+
+**Returns:** ``ok, exists`` pair. If ``ok`` is nil then error happened during the check;
+``exists`` provides an information about error type; otherwise ``exists`` indicated whether
+the element exists.
```
immerrr: What errors  can occur here?

mike1808: You are right. Particularly this method has very little chance to fail (or maybe doesn't have at all).

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
...
+
+.. _splash-element-exists:
+
+element:exists
+--------------
+
+Check whether the element exists in DOM. If the element doesn't exist almost all other
+methods will fail.
+
+**Signature:** ``ok, exists = element:exists()``
```
immerrr: The flag is still here :)

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
+.. _splash-element:
+
+Element Object
+==============
+
+Element objects are instanced by :ref:`splash-select`;
```
kmike: instantiated, created

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
...
+that can be called from Lua to execute node method in page context.
+
+**Async:** no.
+
+See more in :ref:`splash-jsfunc`.
+
+.. note::
+
+    If the specified method returns another HTML element it will be represented as a table with
+    ``type`` property which equals to ``'node'``
```
kmike: why not Element class?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
...
+
+
+.. _splash-element-mouse-click:
+
+element:mouse_click
+-------------------
+
+Trigger mouse click event on the top-left corner of the element.
+
+**Signature:** ``ok, reason = element:mouse_click()``
```
kmike: It'd be nice to support relative coordinates here, defaulting to 0,0

docs/scripting-element-object.rst
mike1808: It's already fixed in the other PR. I am going to refactor this PR.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
...
+
+
+.. _splash-element-png:
+
+element:png
+-----------
+
+Return a screenshot of the element in PNG format
+
+**Signature:** ``ok, shot = element:png{width=nil, height=nil, scale_method='raster'}``
```
kmike: I used a similar function in a project, and it was really nice to be able to take a screenshot of an area which is slightly larger than element itself. What do you think about adding 'pad' argument - a number of pixels to add to the screenshot around the element? Maybe it can also be a table with {top, left, bottm, right} padding widths. I know, more features.. :)

In implementation it'd be nice to have padding function separate and public, so that people can use it e.g. with get_bounds.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,393 @@
...
+
+* keys - string representing the keys to be sent as keyboard events.
+
+**Returns:** ``ok`` pair. If ``ok`` is nil then error happened during the function call.
+
+**Async:** no.
+
+This methods do the following:
+
+* checks whether the selected element is editable
```
kmike: Why is this check needed?

kmike: E.g. element may become editable only after a click, or it may listen to keyboard events without being editable

docs/scripting-ref.rst
```diff
@@ -2436,3 +2436,21 @@ arguments passed to splash, `username` and `password`.
...
+
+Select HTML element from DOM of the current web page
+
+**Signature:** ``splash:select(selector)``
+
+**Parameters:**
+
+* selector - valid CSS selector
+
+**Returns:** a :ref:`splash-element`.
```
kmike: Async flag is missing

splash/browser_tab.py
```diff
@@ -104,6 +106,11 @@ def _init_webpage(self, verbosity, network_manager, splash_proxy_factory, render
         self.web_view.resize(
             QSize(*map(int, defaults.VIEWPORT_SIZE.split('x'))))
 
+    def _init_elements_storage(self):
```
kmike: It'd be nice to have this work lazily, when a first HtmlElement instance is created. This way Splash won't be changing global scope unless required.

docs/scripting-element-object.rst
immerrr: What about making the "padding" relative coordinates, i.e. positive values mean normal html padding, negative values - pad the image "inside" of the element.

On a sidenote, i don't think it's necessary to keep "width" and "height"  parameters with the same, quite obscure meaning as in the api. We might want to seize the opportunity to think of better names or better ways to describe the use cases for these parameters, not necessarily within this project.

splash/browser_tab.py
```diff
@@ -1095,6 +1121,16 @@ def log(self, message, min_level=None):
...
 
 
+class ElementsStorage(QObject):
+    def __init__(self, parent):
+        self.name = str(uuid.uuid1())
+        super(ElementsStorage, self).__init__(parent)
+
+    @pyqtSlot(name="getId", result=str)
+    def get_id(self):
+        return str(uuid.uuid1())
```
kmike: It seems I'm missing something - why is there two uuid.uuid1() calls, one to set self.name, and another in get_id()?

splash/casperjs_utils.py
```diff
@@ -0,0 +1,114 @@
+# Copyright (c) 2011-2015 Nicolas Perriault
```
kmike: could you please add a comment with a link to the casperjs file with relevant source code?

splash/jsutils.py
```diff
@@ -39,6 +40,15 @@ def escape_js(*args):
             else if (o instanceof Date) {
                 return o.toJSON();
             }
+            else if (o instanceof ClientRect) {
+                return {
+                    top: o.top, left: o.left, bottom: o.bottom, right: o.right,
+                    width: o.width, height: o.height,
```
kmike: 👍 

I wonder if it is possible to assign a non-float value to a ClientRect attribute in webkit used in Splash, bypassing this sanitizing function

splash/qtrender_lua.py
```diff
@@ -80,7 +80,7 @@ def get_last(self):
 
 
 def command(table_argument=False, sets_callback=False,
-            decode_arguments=True):
+            decode_arguments=True, flag=False):
```
kmike: could you please use a longer name for `flag` variable? e.g. `errors_as_flags` or something like this (I'm not exactly sure about its meaning)?

splash/tests/test_execute.py
```diff
@@ -3925,3 +3925,433 @@ def test_key_error(self):
...
+    def test_select(self):
+        resp = self.request_lua("""
+        function main(splash)
+            assert(splash:go(splash.args.url))
+            assert(splash:wait(0.1))
+
+            local p = splash:select('p')
+            local form = splash:select('form#login')
+            local username = splash:select('input[name="username"]')
+            local password = splash:select('input[name="password"]')
```
kmike: From this example it looks like element:select() is a good idea - I'd better write `form:select(...)` here, as ideally the lookup should be limited to `form#login`.

splash/tests/test_execute.py
```diff
@@ -3925,3 +3925,433 @@ def test_key_error(self):
...
+
+    def test_node_property_returns_element(self):
+        resp = self.request_lua("""
+        function main(splash)
+            assert(splash:go(splash.args.url))
+            assert(splash:wait(0.1))
+
+            local p = splash:select('p')
+            local ok, parent = p:node_property('parentNode')
+            local ok, remove = parent:node_method('remove')
```
kmike: `splash:select('p'):node_property('parentNone'):node_method('remove')()` is so much nicer when you're writing quick scraping scripts :) I woder if we should sacrifice `ok, result` consistency and start returning nil objects or raising errors

splash/utils.py
```diff
@@ -89,7 +89,7 @@ def get_alive():
         'SplashCookieJar', 'OneShotCallbackProxy',
         '_ExposedRequest', '_ExposedBoundRequest',
         '_ExposedResponse', '_ExposedBoundResponse',
-        '_ExposedTimer',
+        '_ExposedTimer', '_ExposedElement',
```
kmike: could you please add ElementsStorage here?

docs/scripting-element-object.rst
kmike: > What about making the "padding" relative coordinates, i.e. positive values mean normal html padding, negative values - pad the image "inside" of the element.

yeah, that's what I meant, but haven't expressed clearly!

> On a sidenote, i don't think it's necessary to keep "width" and "height" parameters with the same, quite obscure meaning as in the api. We might want to seize the opportunity to think of better names or better ways to describe the use cases for these parameters, not necessarily within this project.

What do you have in mind? It is nice to be able to rescale element when rendering; do you have a better API proposal?

docs/scripting-element-object.rst
kmike: 👍

splash/tests/test_execute.py
immerrr: There's a branch that would allow `splash:select('p').parentNode:remove()`, that raises errors if something goes wrong, and I have high hopes for it. +1 from me.

Also, while writing it made me wonder if we want to have `.parentNode` for attributes and `:remove` for methods.

mike1808:
@immerrr @kmike I've moved all the work to #490 because I've did quite much refactoring. I'm going to fix all your notes in that branch. And if it OK for you I'm going to close this PR.

kmike:
Makes sense 👍

splash/tests/test_execute.py
mike1808: In my PR #490 you can do "form.node:querySelector()"

splash/jsutils.py
mike1808: Why do you want to assign value to the client rect? It won't affect on anything.

splash/browser_tab.py
mike1808: `get_id` is used in JS for retrieving a unique ID for the element. However, it's a good idea to call the same function in the constructor.

splash/browser_tab.py
mike1808: Good point.

docs/scripting-ref.rst
mike1808: Should I write some small guide why this feature is useful?

docs/scripting-element-object.rst
mike1808: The overall idea of "sending keys to element" is so obvious. What does it mean to send keys to the element? 

The original idea was taken from casperj's "setField" method which do almost the same:
- checks whether the element is editable 
- focuses on it 
- send key events

However, in Splash we have more control over the web page (e.g. we can simulate a "real" mouse click, w/o using JS). So, I think you point is good. :)

docs/scripting-element-object.rst
mike1808: Nice idea about the padding. What do think, what should we in the following cases:

- for positive pads, if the padded image is going out of the web page bounds should we throw error or just fill with some color (maybe user-specified) 
- for negative pads, if the padding image has no width or height at all: throw error or return an empty image?

mike1808: > What do you have in mind? It is nice to be able to rescale element when rendering; do you have a better API proposal?

What if we rename the `size` to `rescale_to`?

docs/scripting-element-object.rst
mike1808: Ok
What if that coordinate exceed the element width or height? Just ignore it and click on the other element or throw an error?

kmike: I'd say click anyways, without throwing an error.

docs/scripting-element-object.rst
kmike: > for positive pads, if the padded image is going out of the web page bounds should we throw error or just fill with some color (maybe user-specified)

+1 to fill with a color. I recall there was some filling in existing rendering methods; imho we should do the same here (/cc @immerrr)

> for negative pads, if the padding image has no width or height at all: throw error or return an empty image?

my vote is to return an empty image.

> What if we rename the size to rescale_to? Will it make easier to understand the meaning of that argument?

there is no 'size' argument, there are width and height. User can pass wither both of them, or only one of them. Also, in splash:png and splash:jpeg they work differently: width causes page to be rescaled, while height causes page to be truncated at the bottom. So rescale_to is not the same as width/height.

docs/scripting-ref.rst
kmike: Yes, and more examples could be nice!

splash/browser_tab.py
kmike: Ah, so self.name is an id of elements storage element, and get_id generates an id of an element inside this storage? This makes sense. Could you please add docstrings to make it more clear?

splash/jsutils.py
kmike: Someone may want to assign a value to client rect in order to bypass sanitanization.

splash/tests/test_execute.py
kmike: hm, by the way, is there a difference between `node:querySelector(css)` and `:select(css)`? Can we make `:select(css)` a shortcut for `.node:querySelector(css)`? Are there extra features `:select()` provides?

PR #472, opened 2016-06-24
DOC use getClientRects instead of getBoundingClientRect

Youwotma:
The bounding client rect's coordinates might not be part of the element
if it's comprised of several client rects.


kmike:
Thanks @Youwotma!

PR #475, opened 2016-06-27
Add PyResult to pass retvals and exceptions to Lua

immerrr:
This PR is the first step to streamline Lua<->Python interaction.

It replaces multiple specific kwargs with one generic way of passing values and exceptions from Python to Lua interpreter: `PyResult`. There are three ways the result can be handled in Lua:

- `PyResult(*args)` (or `PyResult.return_(*args)`)

      Returns args to Lua interpreter.

- `PyResult.raise_(error)`

      Raises an error in Lua interpreter.

- `PythonResult.yield_(*args)`

      Returns args asynchronously via ``coroutine.yield``

In Lua it is paired by `wraputils:unwrap_python_result` function that does what's necessary to propagate the desired result.

### DIFF
--- a/splash/lua_runner.py
+++ b/splash/lua_runner.py
@@ -7,8 +7,8 @@ import six
 import lupa
 
 from splash.exceptions import ScriptError
-from splash.lua import parse_error_message
-from splash.utils import truncated, PyResult, ensure_tuple
+from splash.lua import parse_error_message, PyResult, ensure_tuple
+from splash.utils import truncated
 
 
 class AsyncCommand(object):
--- a/splash/lua_runner.py
+++ b/splash/lua_runner.py
@@ -7,8 +7,8 @@ import six
 import lupa
 
 from splash.exceptions import ScriptError
-from splash.lua import parse_error_message, PyResult, ensure_tuple
-from splash.utils import truncated
+from splash.lua import parse_error_message, PyResult
+from splash.utils import truncated, ensure_tuple
 
 
 class AsyncCommand(object):
--- a/splash/lua_runner.py
+++ b/splash/lua_runner.py
@@ -8,12 +8,7 @@ import lupa
 
 from splash.exceptions import ScriptError
 from splash.lua import parse_error_message
-from splash.utils import truncated
-
-
-class ImmediateResult(object):
-    def __init__(self, value):
-        self.value = value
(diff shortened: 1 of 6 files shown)


immerrr:
Replaced `op=` kwarg with named constructors, now it should be even less typing.

splash/lua_runner.py
```diff
@@ -90,31 +87,51 @@ def on_lua_error(self, lua_exception):
 
     def dispatch(self, cmd_id, *args):
         """ Execute the script """
-        args = args or None
-        args_repr = truncated("{!r}".format(args), max_length=400, msg="...[long arguments truncated]")
-        self.log("[lua_runner] dispatch cmd_id={}, args={}".format(cmd_id, args_repr))
+        args = args or ()
+
+        def truncated_repr(x):
```
kmike: This can be used in a few other places - what do you think about moving this function to utils?

splash/qtrender_lua.py
```diff
@@ -477,23 +472,24 @@ def wait(self, time, cancel_on_redirect=False, cancel_on_error=True):
             })
 
         def success():
-            cmd.return_result(True)
+            cmd.return_result(PyResult(True))
```
kmike: Is it possible to fix this as well, i.e. not to require wrapping here?

splash/utils.py
```diff
@@ -185,3 +185,66 @@ def wrapper(self, *args, **kwargs):
             return meth(self, *args, **kwargs)
         return wrapper
     return decorator
+
+
+class PyResult(object):
```
kmike: What about moving it to qtrender_lua?

kmike: Ah, it is used in splash.lua. What about moving it there then?

kmike: I'm asking because so far splash.utils was not aware of Lua (with a small exception of `get_alive`); PyResult looks specific to Lua scripting support.

immerrr: Fair point, I'll see what I can o

splash/lua_runner.py
immerrr: There's already `utils.truncated`, this func just adds a few args common for this specific method.

kmike:
Thanks for the cleanup @immerrr!

PR #482, opened 2016-07-17
wraputils.lua setup_property_access shouldn't override class methods

mike1808:
This PR fixes bug when you create multiple instances for exposed objects. For example, when you call `splash#call_later` you create a new instance of `_ExposedTimer` Python class and `Timer` Lua class.

### DIFF
--- a/splash/lua_modules/splash.lua
+++ b/splash/lua_modules/splash.lua
@@ -10,7 +10,7 @@ local Request = require("request")
 --
 -- Lua wrapper for Splash Python object.
 --
-local Splash = wraputils.create_metatable(Splash)
+local Splash = wraputils.create_metatable()
 local Splash_private = {}
 
 function Splash._create(py_splash)
@@ -65,7 +65,7 @@ end
 --
 -- Timer Lua wrapper
 --
-local Timer = wraputils.create_metatable(Timer)
+local Timer = wraputils.create_metatable()
 local Timer_private = {}
 
 function Timer._create(py_timer)
--- a/splash/lua_modules/splash.lua
+++ b/splash/lua_modules/splash.lua
@@ -10,14 +10,11 @@ local Request = require("request")
 --
 -- Lua wrapper for Splash Python object.
 --
-local Splash = wraputils.create_metatable()
+local Splash = wraputils.create_metatable(Splash)
 local Splash_private = {}
 
 function Splash._create(py_splash)
   local splash = { args = py_splash.args }
-  setmetatable(splash, Splash)
-  Splash.__index = Splash
-  Splash.__newindex = rawset
   wraputils.wrap_exposed_object(py_splash, splash, Splash, Splash_private, true)
   return splash
 end
@@ -68,14 +65,11 @@ end
 --
 -- Timer Lua wrapper
 --
-local Timer = wraputils.create_metatable()
+local Timer = wraputils.create_metatable(Timer)
 local Timer_private = {}
 
 function Timer._create(py_timer)
   local timer = {}
-  setmetatable(timer, Timer)
-  Timer.__index = Timer
-  Timer.__newindex = rawset
   wraputils.wrap_exposed_object(py_timer, timer, Timer, Timer_private, true)
   return timer
 end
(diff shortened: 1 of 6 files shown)


splash/tests/test_execute_callbacks.py
```diff
@@ -834,6 +834,35 @@ def test_error_handled_in_callback(self):
...
+          timer6:cancel()
+
+          splash:wait(1)
+
+          return treat.as_array(o)
+        end
+        """)
+
+        self.assertStatusCode(resp, 200)
+        self.assertEqual(resp.json(), [False, False, False, False, False, False])
```
mike1808: If you run this test w/o this fix you will get `[False, True, True, True, True, False]`. This happens because when `setup_property_access` is called it sets `__index` and `__setindex` metamethods to the provided class. And that way only the first and the last created instances will work correctly, because for the class (table) of the first instance is created without any overrides and the last instance works because its class methods  has right `self` arguments.

splash/lua_modules/wraputils.lua
```diff
@@ -134,32 +134,33 @@ end
 -- Handle @lua_property decorators.
 --
 local function setup_property_access(py_object, self, cls)
-  local setters = {}
-  local getters = {}
+  self.setters = {}
+  self.getters = {}
```
mike1808: I suppose that we should somehow hide this objects from `self`. The previous closure method didn't work.

mike1808:
Also, I didn't find a way to test `@lua_property` methods (getters and setters).

splash/lua_modules/splash.lua
```diff
@@ -14,10 +14,12 @@ local Splash = wraputils.create_metatable()
...
 
 function Splash._create(py_splash)
-  local self = {args=py_splash.args}
-  setmetatable(self, Splash)
-  wraputils.wrap_exposed_object(py_splash, self, Splash, Splash_private, true)
-  return self
+  local splash = { args = py_splash.args }
+  setmetatable(splash, Splash)
+  Splash.__index = Splash
+  Splash.__newindex = rawset
```
mike1808: This is used to reset the result of the previous `setup_property_access` call.

splash/tests/test_execute_callbacks.py
immerrr: This comment belongs in the source code, not here.

splash/lua_modules/wraputils.lua
```diff
@@ -169,6 +170,9 @@ end
 -- Create a Lua wrapper for a Python object.
 --
 local function wrap_exposed_object(py_object, self, cls, private_self, async)
+  setmetatable(self, cls)
+  cls.__newindex = rawset
```
immerrr: hmm, does this metamethod only live up to `setup_property_access` where it's reset with setter-related one?

splash/lua_modules/wraputils.lua
```diff
@@ -134,32 +135,33 @@ end
...
       end
     end
   end
 
-  function cls:__newindex(index, value)
-    if setters[index] then
-      return setters[index](self, value)
+  function cls:__index(index)
+    if self.__getters[index] then
+      return self.__getters[index](self)
```
immerrr: I still stumble upon this function's header each time I revisit this code.

Can we rewrite it (and `__newindex` below) to `function cls.__index(self, index)` or `cls.__index = function(self, index)` so that it's more obvious that `self` is an argument to that function rather than `cls` or `self` from the outside scope?

immerrr: Also, can we move these metamethods to `create_metatable` so that they're executed only once per class?

mike1808: > Can we rewrite it (and __newindex below) to function cls.__index(self, index) or cls.__index = function(self, index) so that it's more obvious that self is an argument to that function rather than cls or self from the outside scope?

Sure, we can.

> Also, can we move these metamethods to create_metatable so that they're executed only once per class?

Unfortunately, I cannot do it in `create_metatable` function, because I need a class.

immerrr: > Unfortunately, I cannot do it in create_metatable function, because I need a class.

Why not? Our metatables *are* classes, aren't they? :)

mike1808: I know, they are 😄 However, I've tried to put their and it didn't work.

immerrr: ok, let's keep it that way for now (but please add a FIXME there)

immerrr:
LGTM

PR #484, opened 2016-07-25
wraputils.create_metatable: add index & newindex metamethods

immerrr:
Should fix the FIXME in the wraputils code.


mike1808:
I've tried to do in this way and it didn't work. Everything hangs when I start my tests.

PR #487, opened 2016-07-26
wraputils.setup_commands store private methods in the instance itself rather than in the separate table

mike1808:


splash/lua_modules/splash.lua
```diff
@@ -25,7 +24,7 @@ end
 -- It is required to handle errors properly.
 --
 function Splash:jsfunc(...)
-  local func = Splash_private.jsfunc(self, ...)
+  local func = self.__private.jsfunc(self, ...)
```
kmike: They are private because users shouldn't be able to access them; it seems that adding `__private` attribute breaks that.

mike1808: In that case we should fix `drops_self_argument` because it make useless the usage of self.

kmike:
Fixed by #495.

Issue #229, opened 2015-06-01
error info should be a table in splash:go and splash:set_content 


kmike:
It should make debugging easier. Error handling code will also become more straightforward. Instead of 
`if reason:sub(0,4) == 'http'` users will be able to write something like `if error.type == 'http'`.

Two downsides / things to consider:

1. By default, Lua tables are displayed as `table: 0x7fc251510e10`. It is less useful than e.g. `http404`. We should check where can user see this value, and make sure a nice repr is returned: Splash-Jupyter? logs? Splash UI? result returned when `assert(splash:go(...))` raises an error?
2. This change is backwards incompatible. It can be made more compatible if we add `sub` method and override `__eq` metamethod to the resulting table. It may not worth the trouble to keep backwards compatibility though.

pawelmhm:
What would be in this Lua table aside from status code? One thing that is missing from error responses now is url that caused go() to fail. Currently when go fails you get something like this:

`unhandled Lua error: [string "function main(splash) local first = true spla..."]:1: http404`

if you have many go() in script you don't know which go() call caused failure, you can investigate traceback but it's not very intuitive, would be nicer to have something like this:

```
{
    "traceback": 'unhandledLuaerror: [string"functionmain(splash)localfirst=truespla...]: 1: http404',
    "target_response": {
        "url": "foobar.com",
        "status": "404"
    }
}```

this would be for go() I'm not sure about set_content.

kmike:
Removed 2.0 milestone because it looks like this change can be made in a backwards-compatible way.

PR #490, opened 2016-08-02
[WIP] splash:select, DOM elements manipulating helpers

mike1808:


splash/qtrender_lua.py
```diff
@@ -47,6 +47,17 @@
...
 
+try:
+    from functools import partialmethod
+except ImportError:
+    class partialmethod(functools.partial):
+        def __get__(self, instance, owner):
+            if instance is None:
+                return self
+            return functools.partial(self.func, instance,
+                                     *(self.args or ()), **(self.keywords or {}))
```
immerrr: not needed anymore, i guess?

splash/qtrender_lua.py
```diff
@@ -1273,13 +1292,85 @@ def clear(self):
...
+
+    ]
 
     def __init__(self, lua, exceptions, element, splash):
         self.element = element
         self.splash = splash
         super(_ExposedElement, self).__init__(lua, exceptions)
 
+    @staticmethod
+    def init_properties():
```
immerrr: You can make it
```
@classmethod
def init_properties(cls):
    ....
```
and refer to the class as `cls` if that's more convenient (but i don't insist)

splash/qtrender_lua.py
mike1808: Oh, yeah. You are right. I will remove this part.

splash/html_element.py
```diff
@@ -0,0 +1,283 @@
...
+        return call
+
+    def mouse_click(self, x=0, y=0, button="left"):
+        """ Click on the element """
+        self.assert_element_exists()
+        dimensions = self.tab.evaljs(
+            DIMENSIONS_JS_FUNC % self.element_js,
+        )
+
+        in_bounds = 0 <= x < dimensions["width"] and 0 <= y < dimensions["height"]
```
mike1808: Is this validation needed? Maybe we should allow user to click everywhere he wants? For example, I want to click on the element which is located above of my element and I will pass negative `y` to this method.

mike1808:
I've added `element:fill(values)` method which set the values of the form using the provided `values` table. However, in comparison with casperjs it doesn't automatically submit the form and only supports the name selector type (the keys of the table are used to create the following selectors `[name="<key>"]`).

splash/html_element.py
kmike: +1 to remove validation

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,550 @@
...
+
+.. _splash-element-mouse-click:
+
+
+element:mouse_click
+-------------------
+
+Trigger mouse click event on the element.
+
+**Signature:** ``ok, reason = element:mouse_click(x, y)``
```
kmike: It may be better to write it with default values: `ok, reason = element:mouse_click{x=0, y=0}`

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,550 @@
...
+
+**Returns:** ``visible`` indicated whether
+the element is visible.
+
+**Async:** no.
+
+
+.. _splash-element-fetch-text:
+
+element:fetch_text
```
kmike: I don't quite like the `fetch_text` name because `fetch` usually implies some async action, e.g. a HTTP request. What about `element:get_text()` or just `element:text()`?

By the way, it looks like some of the methods use `get_` prefix, and some of them don't. Is it intentional, it there a pattern?

mike1808: `fetch_text` was taken from casperjs

>it looks like some of the methods use get_ prefix, and some of them don't. Is it intentional, it there a pattern?

Nice catch. No, there wasn't any intended pattern. I will remove all `get_` prefixes and rename `fetch_text` to `text`.

splash/html_element.py
```diff
@@ -0,0 +1,285 @@
...
+        return self.tab.evaljs(u"({fill_form_values_func})({element}, {values})".format(
+            fill_form_values_func=FILL_FORM_VALUES_JS,
+            element=self.element_js,
+            values=escape_js(values)
+        ))
+
+    def send_keys(self, text):
+        """ Send key events to the element separated by whitespaces """
+        self.mouse_click()
+        self.tab.send_keys(text)
```
immerrr: Why do we do `mouse_click` and not `focus` here? (asking as a person who rarely uses his mouse and prefers to TAB-TAB-TAB around form fields)

mike1808: What is the significant difference between clicking on an input and triggering focus event?

immerrr: In a normal scenario, not much, clicking an input element should "focus" it.

But it might do strange things when you're trying to `send_keys` to clickable elements like buttons and links or when mouse-related event handlers define different behaviour.

mike1808: I remembered why I used `mouse_click` here. If try to use `focus` using `self.node_method('focus')` it doesn't work every time. Because when you call `node_method` it evaluates JS which is not immediate operation. On the contrary, calling `mouse_click` generates mouse click event using Qt.

splash/vendor/harviewer
```diff
@@ -1 +1 @@
-Subproject commit 5d0f5e2f121f5d7de7194eda0059b6a4c50c5561
```
kmike: Is the downgrade intended?

splash/tests/test_execute.py
```diff
@@ -3937,3 +3936,1153 @@ def test_key_error(self):
         """)
         self.assertScriptError(resp, ScriptError.SPLASH_LUA_ERROR,
                                message="Unknown key")
+
+
+class HTMLElementTest(BaseLuaRenderTest):
```
kmike: Could you please move these tests to a new file?

splash/vendor/harviewer
mike1808: No, I forgot to upgrades submodules in my local repo.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
+.. _splash-element:
+
+Element Object
+==============
+
+Element objects are created by :ref:`splash-select` and :ref:`splash-select-all`.
```
kmike: Can they be created by splash:evaljs and splash:jsfunc calls as well?

mike1808: No, they cannot. If you return node from these methods you will receive objects like this:

```js
{ type: 'node', id: '<uuid>' }
```

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+    - isDefaultNamespace
+    - isEqualNode
+    - isSameNode
+    - lookupPrefix
+    - lookupNamespaceURI
+    - normalize
+    - removeChild
+    - replaceChild
+
+Also, you can attach event handlers to the specified event. When the handler is called it will
```
kmike: it could be helpful to explain what's the difference elem.node:from addEventListener

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+.. _Event: https://developer.mozilla.org/en-US/docs/Web/API/Event
+.. _EventTarget: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget
+
+
+.. _splash-element-inner_id:
+
+element.inner_id
+----------------
+
+Id of the inner representation of the element.
```
kmike: Is it the internal element id used by Splash, as opposed to HTML element id attribute? What does 'inner representation' mean?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+Methods
+~~~~~~~
+
+To modify or retrieve some information about the element you can use the following methods.
+Note that all of the following methods return flag indicating whether the operation was
+successful or not. The methods returns ``ok, reason_or_value`` pair. If ``ok`` is nil
```
kmike: it seems `ok, reason_or_value` API is not used for many of the methods; maybe it makes sense to just document it for individual methods.

docs/scripting-element-object.rst
mike1808: Yes, it's the internal element id used by Splash to access the element object in the JS window object.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+then error happened during the operation; ``reason`` provides an information about error type;
+otherwise ``ok`` is ``true`` and the returned value is stored in the second variable.
+
+
+.. _splash-element-exists:
+
+element:exists
+--------------
+
+Check whether the element exists in DOM. If the element doesn't exist some of the methods will fail raising
```
kmike: When can element disappear? I think that'd be good to explain that, or maybe just say something like "check whether element *still* exists"

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+**Signature:** ``ok, reason = element:mouse_click{x=0, y=0}``
+
+**Parameters:**
+
+* x - optional, x coordinate relative to the left corner of the element
+* y - optional, y coordinate relative to the top corner of the element
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
+function call; ``reason`` provides an information about error type.
```
kmike: When can click fail? Should user care about it when writing a script?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+**Async:** no.
+
+If x or y coordinate is not provided they will be set to 0 and the click will be triggered
```
kmike: mouse_hover doesn't trigger a click

docs/scripting-element-object.rst
mike1808: When he tries to click on element which doesn't exist.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+on the left-top corner of the element. The coordinates can have a negative value which means
+the click will be triggered outside of the element.
+
+Mouse events are not propagated immediately, to see consequences of click
```
kmike: there is no click in mouse_hover :)

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+**Async:** no.
+
+Example of getting the font size of the element using this method.
+
+.. code-block:: lua
+
+    function main(splash)
+        local element = splash:select('.element')
+        return element:styles()['font-size']
+    end
```
kmike: What's the difference between `element:styles()` and `element.node.style`? Could you please add it to docs?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+* width - optional, width of a screenshot in pixels;
+* scale_method - optional, method to use when resizing the image, ``'raster'``
+  or ``'vector'``;
+* pad - optional, integer or ``{left, top, right, bottom}`` values of padding
+
+**Returns:** ``ok, shot`` pair. If ``ok`` is nil then error happened during the
+function call; ``shot`` provides an information about error type; otherwise
+``shot`` is a PNG screenshot data, as a :ref:`binary object <binary-objects>`.
+When the result is empty (e.g. if the element is not visible) ``nil`` is returned.
```
kmike: What do you think about removing ok flag for this function? It looks like for practical purposes there is no much difference between invisible HTML element and an element which no longer present on a page.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+**Async:** no.
+
+*pad* parameter sets the padding of the resulting image. If it is a single integer then the
+padding from all sides will be equal. If the value of the padding is positive the resulting screenshot
```
kmike: What else can it be, besides a single integer?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+
+.. _splash-element-field-value:
+
+element:field_value
+-------------------
+
+Get value of the field element (input, select).
+
+**Signature:** ``ok, info = element:field_value()``
```
kmike: it is documented as `ok, value` below

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+* height - height of the element
+* visible - flag representing if the element is visible
+
+
+.. _splash-element-field-value:
+
+element:field_value
+-------------------
+
+Get value of the field element (input, select).
```
kmike: what about textarea?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+element:form_values
+-------------------
+
+Return a table with form values if the element type is *form*
+
+**Signature:** ``ok, info = element:form_values()``
+
+**Returns:** ``ok, values`` pair. If ``ok`` is nil then error happened during the function call
+or node type is not *form*; ``values`` provides an information about error type; otherwise
```
kmike: So there are two possible errors; it means `values` error message is not purely for debugging - developer may want to check the error type in a Lua script; it means we should document which values can it take.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+``values`` is a table of values.
```
kmike: Is it just a table of values (i.e. a plain array with values), or a table with element names as keys and values as values?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+    <form id="login">
+        <input type="text" name="username" />
+        <input type="password" name="password" />
+    </form>
+
+.. code-block:: lua
+
+    function main(splash)
+        assert(splash:select('.login'):fill({ username="admin", password="pass" }))
+    end
```
kmike: Is it possible to get form values using elem:form_values(), change a couple of field values and call splash:fill() - are values compatible?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+Send keyboard events to the element.
+
+**Signature:** ``ok = element:send_keys(keys)``
+
+**Parameters**
+
+* keys - string representing the keys to be sent as keyboard events.
+
+**Returns:** ``ok`` pair. If ``ok`` is nil then error happened during the function call.
```
kmike: ok is a flag, not a pair

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+
+**Async:** no.
+
+This methods do the following:
```
kmike: this method does the following

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,566 @@
...
+.. _splash-element-submit:
+
+element:submit
+--------------
+
+Submit the form element.
+
+**Signature:** ``ok = element:submit()``
+
+**Returns:** ``ok`` pair. If ``ok`` is nil then error happened during the function call.
```
kmike: What are possible errors, and why isn't there a reason field?

docs/scripting-ref.rst
```diff
@@ -2477,3 +2477,84 @@ arguments passed to splash, `username` and `password`.
...
     end
+
+
+.. _splash-select:
+
+
+splash:select
+-------------
+
+Select the first HTML element from DOM of the current web page that matches the specified selector.
```
kmike: I think we should say that it is a CSS selector

docs/scripting-ref.rst
```diff
@@ -2477,3 +2477,84 @@ arguments passed to splash, `username` and `password`.
...
+
+        local el = splash:select('.element')
+        local hash = {}
+        local classNames = {}
+
+        while el do
+          local classList = el.node.classList
+          if classList then
+            for _, v in ipairs(classList) do
+              if (not hash[v]) then
```
kmike: `if not seen[v]` can be easier to read

docs/scripting-element-object.rst
mike1808: it do the following:

if the element is `select` and it has multiples values it returns a table with that values
if the element is `select` and it has single value it returns it
if the element has `type = "radio"` it returns it's value if it's checked and `nil` otherwise
if the element has `type = "checkbox"` it returns bool
otherwise it returns the `value` property value or empty string if it doesn't exist

docs/scripting-element-object.rst
mike1808: it's a table with keys as element's names and values as its values

docs/scripting-element-object.rst
mike1808: Yes, they are.

splash/browser_tab.py
```diff
@@ -90,6 +92,7 @@ def _init_webpage(self, verbosity, network_manager, splash_proxy_factory, render
 
         self._set_default_webpage_options(self.web_page)
         self._setup_webpage_events()
+        self._init_elements_storage()
```
kmike: it is also called in _init_js_objects_storage - is that ok?

mike1808: No, this line should be removed.

splash/browser_tab.py
```diff
@@ -1116,6 +1185,140 @@ def log(self, message, min_level=None):
         log.msg(message, system='render')
 
 
+class ElementsStorage(QObject):
```
kmike: Could you please add docstrings for these classes (ElementsStorage, Event, EventHandlersStorage, EventsStorage)?

splash/browser_tab.py
```diff
@@ -1116,6 +1185,140 @@ def log(self, message, min_level=None):
...
+
+class EventHandlersStorage(QObject):
+    def __init__(self, parent, events_storage):
+        self.name = str(uuid.uuid1())
+        self.events_storage = events_storage
+        self.storage = {}
+        super(EventHandlersStorage, self).__init__(parent)
+
+    def add(self, func):
+        func_id = str(uuid.uuid1())
```
kmike: calling `str(uuid.uuid1())` looks common - what do you think about extracting it to a function? it is also used in wait_for_resume implementation.

docs/scripting-element-object.rst
mike1808: Can you explain about what difference do you want me to write?

docs/scripting-element-object.rst
mike1808: It can disappear if some JS removes it from DOM. But it will be still present in memory. 
You can get non-existing elements using the selected element. E.g with `splash.node:clone()`

docs/scripting-element-object.rst
mike1808: Sounds reasonable.

docs/scripting-element-object.rst
mike1808: The reason is absent my mistake. The reason can be if you want to submit non-form element.

splash/browser_tab.py
mike1808: No, it'd be removed.

docs/scripting-element-object.rst
mike1808: Here is a quote from MDN:

>The HTMLElement.style property returns a CSSStyleDeclaration object that represents only the element's inline style attribute, ignoring any applied style rules.

docs/scripting-element-object.rst
mike1808: @kmike should we convert the string representation of an exception to dictionary (table in Lua) to allow the developer to do something following:

```lua
local ok, error = foo()

if not ok then
 if error.type == 'DOMError' then
   -- handle error
  end
end 
-- ...
```

docs/scripting-element-object.rst
kmike: Is it just a missing feature, or is it by design? What's the reason for that?

docs/scripting-element-object.rst
kmike: Ouch, my comment was gibberish, I edited it now.

docs/scripting-element-object.rst
kmike: Could you please add this to docs? The current docs are good, and they are correct; I was just reading them and asking you questions that popped up. I think other users may have simila questions, and adding these explanations to the docs makes sense.

docs/scripting-element-object.rst
kmike: Could you please add this explanation to the element:exists docs?

docs/scripting-element-object.rst
kmike: It makes sense (both textarea and button have value attribute with a right content); we should document that this function works for all form elements, not only for input and select.

docs/scripting-element-object.rst
kmike: That's a good idea. See also: https://github.com/scrapinghub/splash/issues/229.

docs/scripting-element-object.rst
mike1808: It's more like a missing feature, because I didn't think about it.

kmike: How hard is it to fix that? I think it is better to do before the merge because we'll get a backwards compatibility issue if we change it later.

docs/scripting-element-object.rst
mike1808: wdyt, if I add table errors only for these table  methods? adding for other ones will make this PR more bigger and harder to review :)

kmike: Sure, I was mentiong this PR to support the idea, not asking you to fix all Splash issues :)

docs/scripting-element-object.rst
mike1808: @kmike added.

docs/scripting-element-object.rst
mike1808: You are right, but you've commented on wrong line :)

docs/scripting-element-object.rst
mike1808: Fixed

docs/scripting-element-object.rst
mike1808: It took more time than I had expected. Let's move it to another PR,

mike1808:
@kmike is it time for squashing the commits?

Dockerfile
```diff
@@ -39,3 +39,5 @@ ENTRYPOINT [ \
     "--filters-path", "/etc/splash/filters", \
     "--lua-package-path", "/etc/splash/lua_modules/?.lua" \
 ]
+
+CMD pip3 install -r /app/requirements-dev.txt /app
```
kmike: What's this for? That's not good to ship unneeded dependencies for end users.

mike1808: Oh, sorry. I did that for local tests and forgot to remove it.

splash/utils.py
```diff
@@ -197,3 +200,35 @@ def ensure_tuple(val):
...
+
+    if obj is None:
+        return None
+
+    if predicate(obj):
+        return convert(obj)
+
+    if isinstance(obj, dict):
+        return {
+            key: traverse_obj(value, predicate, convert, max_depth=max_depth - 1)
```
kmike: Shouldn't it traverse key as well?

mike1808: In what cases a key can be a HTML element?

kmike: This function is un utils.py, its description says nothing about HTML elements. If it is only for traversing specific input (a Python data structure which possibly contain HTML Element wrappers?) then it could be better to move it somewhere else.

mike1808: Ok, I've added docstring and changed that.

docs/scripting-element-object.rst
kmike: Great! Could you please update the docs as well?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,742 @@
...
+            y = event.clientY
+        end
+
+        assert(splash:wait(10))
+
+        return x, y
+    end
+
+The another way to attach event handlers is to use ``element.node:addEventListener(event, listener)``.
+It allows you to add more than a single event handler for an event.
```
kmike: 👍

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,742 @@
...
+
+.. code-block:: lua
+
+    function main(splash)
+        -- ...
+        local element = splash:select('.element')
+        local cloned = element.node:cloneNode() -- the cloned element isn't in DOM
+        local exists = cloned:exists() -- exists will be `false`
+        -- ...
+    end
```
kmike: 👍

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+This method returns the result of JavaScript `Window.getComputedStyle()`_ applied on the element.
+
+Example: get all computed styles and return the ``font-size`` property.
+
+.. code-block:: lua
+
+    function main(splash)
+        -- ...
+        local element = splash:select('.element')
+        return element:styles()['font-size']
```
kmike: note for the future: probably `element:style('font-size')` makes more sense - this is usually what developer wants, it is less code and less JS->Python->Lua conversions

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+        <input type="text" name="username" value="admin" />
+        <input type="password" name="password" value="pass" />
+    </form>
+
+.. code-block:: lua
+
+    function main(splash)
+        -- ...
+        local form = splash:select('#login')
+        local ok, values = assert(form:form_values())
```
kmike: isn't assert removing ok flag?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+
+
+.. _splash-element-form-values:
+
+element:form_values
+-------------------
+
+Return a table with form values if the element type is *form*
+
+**Signature:** ``ok, info = element:form_values()``
```
kmike: return value is called `ok, values` below; it is better to use the same name

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+        <input type="text" name="username" value="admin" />
+        <input type="password" name="password" value="pass" />
+    </form>
+
+.. code-block:: lua
+
+    function main(splash)
+        -- ...
+        local form = splash:select('#login')
+        local ok, values = assert(form:form_values())
```
kmike: isn't ok flag removed by assert?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+
+Send keyboard events to the element.
+
+**Signature:** ``ok = element:send_keys(keys)``
+
+**Parameters**
+
+* keys - string representing the keys to be sent as keyboard events.
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
```
kmike: in signature it is just 'ok'

docs/scripting-element-object.rst
mike1808: no, it isn't
in `wraputils.lua` we have method `assertx` that do it, but I didn't want to include it into sandbox and write into docs

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+
+Send keyboard events to the element.
+
+**Signature:** ``ok = element:send_text(text)``
+
+**Parameters**
+
+* text - string to be sent as input.
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
```
kmike: in signature it is just 'ok', not 'ok, reason' pair

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+function call; ``reason`` provides an information about error type.
+
+**Async:** no.
+
+This methods do the following:
```
kmike: this method does the following

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,744 @@
...
+.. _splash-element-submit:
+
+element:submit
+--------------
+
+Submit the form element.
+
+**Signature:** ``ok = element:submit()``
+
+**Returns:** ``ok, reason`` pair. If ``ok`` is nil then error happened during the
```
kmike: signature only mentions a single return value

docs/scripting-element-object.rst
kmike: ah, right, sorry for the noise

kmike: What do you think about removing 'ok' flag, and returning nil, reason instead of values table if there is an error? This is the way [splash:wait_for_resume](http://splash.readthedocs.io/en/stable/scripting-ref.html#splash-wait-for-resume) works; it allows to write just `value = assert(form:form_values())`

docs/scripting-ref.rst
```diff
@@ -2477,3 +2477,104 @@ arguments passed to splash, `username` and `password`.
...
+
+* selector - valid CSS selector
+
+**Returns:** a :ref:`splash-element`.
+
+**Async:** no.
+
+Using :ref:`splash-select` you can get the element that matches your specified CSS selector like using
+`document.querySelector`_ in your browser. The returned element is a :ref:`splash-element` which has many useful
+methods and almost all methods and attributes that element has in JavaScript.
```
kmike: What happens when an element can't be found - is this correct that nil is returned? I think it should be documented. Returning nil also means el = assert(splash:select("")) can be used.

docs/scripting-ref.rst
```diff
@@ -2477,3 +2477,104 @@ arguments passed to splash, `username` and `password`.
...
+
+Example: select an element which has ``element` class and return class
```
kmike: element has 2 qoute marks at the left, but only one at the right

splash/jsutils.py
```diff
@@ -76,6 +96,45 @@ def get_sanitized_result_js(expression, max_depth=0):
...
+    };
+  }
+
+  if (o instanceof Node) {
+    return storeNode(o);
+  } else if (o instanceof NodeList) {
+    return Array.prototype.slice.call(o).map(storeNode);
+  }
+
+  return o;
```
kmike: Does it mean user can return `{type: 'node', id: '...'}` and Splash will think a DOM element was returned? What do you think about always wrapping the result here - for non-node objects it can return e.g. `{type: 'any', value: ...}, and for node lists it can return `{type: 'nodelist', 'value': ...}`, and the unwrapping the result?

splash/browser_tab.py
```diff
@@ -905,6 +959,29 @@ def send_keys(self, text):
         for key in text.split():
             qt_send_key(key, self.web_page)
 
+    def select(self, selector):
+        """ Selects DOM element and returns an instance of `HTMLElement`
```
kmike: a very minor nitpick: pep8 says docstrings should use imperative mood ("Select ..", not "Selects ..")

docs/scripting-element-object.rst
mike1808: And what if the result is `nil`?

kmike: The result is a table; if there are no elements then it is an empty table, not nil. Assert won't fire an error on empty table.

mike1808: I am using a decorator for returning flags when the method raises an error, and I have methods that return flag but do not have any return values (like `mouse_click`). I was thinking about some generic solution for them. It can be the following:

- if the method successfully finished and it has a return value- return it as a first value
- if the method successfully finished and it doesn't have a return value - return `True` as a first value
- if the method failed - leave as is (`nil, reason`)

kmike: It makes sense, with one caveat: it only works if the return value can't be nil or false - e.g. if it is always a table or other object.

mike1808: I only have one method that can return `nil` or `false` – `field_value()`.

docs/scripting-ref.rst
mike1808: Updated docs.

docs/scripting-element-object.rst
mike1808: @kmike I've changed the signature. Now it returns `values, reason`.

splash/browser_tab.py
```diff
@@ -1116,6 +1193,176 @@ def log(self, message, min_level=None):
...
+class ElementsStorage(QObject):
+    """
+    Object that allows to store JavaScript Node objects.
+
+    This creates a JavaScript-compatible object (can be added to `window`)
+    that has `get_id()` function which can be called from JavaScript for
+    retrieving a unique id for each Node object
+    """
+    def __init__(self, parent):
+        self.name = self.get_id()
```
kmike: imho just `get_id()` is more clear here, as the id is not related

splash/browser_tab.py
```diff
@@ -1116,6 +1193,176 @@ def log(self, message, min_level=None):
...
+        window[{storage_name}].preventDefault.connect(window[{storage_name}].callMethod('preventDefault'))
+        window[{storage_name}].stopImmediatePropagation.connect(window[{storage_name}].callMethod('stopImmediatePropagation'))
+        window[{storage_name}].stopPropagation.connect(window[{storage_name}].callMethod('stopPropagation'))
+
+        window[{storage_name}].add = function(event) {{
+            var id = window[{storage_name}].getId()
+            window[{storage_name}].events[id] = event;
+            return id;
+        }}
+        """.format(storage_name=escape_js(self.name))))
```
kmike: the code can be made a bit less repetitive if you wrap the whole thing in an anonymous function and add a local variable `var storage=window[{storage_name}]`

splash/browser_tab.py
```diff
@@ -1116,6 +1193,176 @@ def log(self, message, min_level=None):
...
+    def remove_event(self, event_id):
+        frame = self.parent().web_page.mainFrame()
+        eval_expr = u"eval({})".format(escape_js("""
+        delete window[{storage_name}].events[{event_id}]
+        """.format(
+            storage_name=escape_js(self.name),
+            event_id=escape_js(event_id),
+        )))
+
+        frame.evaluateJavaScript(get_process_errors_js(eval_expr))
```
kmike: it looks like `eval_expr = "eval({})".format(escape_js(txt)); result = frame.evaluateJavaScript(get_process_errors_js(eval_expr)` is repeated a few time; what do you think about introducing a helper function for that? Or just for `get_process_errors_js(eval("{}".format(escape_js(txt))))`?

splash/html_element.py
```diff
@@ -0,0 +1,317 @@
...
+    return (elem.textContent || elem.innerText || elem.value || '').trim();
+})(%s)
+"""
+
+FILL_FORM_VALUES_JS = """
+function fill(form, values) {
+    Object.keys(values).forEach(function(name) {
+        var input = form.querySelector("input[name='" + name + "']")
+            || form.querySelector("select[name='" + name + "']")
+            || form.querySelector("textarea[name='" + name + "']");
```
kmike: ideally this should use CSS.escape for field names; it is a big [copy-paste](https://github.com/mathiasbynens/CSS.escape/blob/ff41bbb0702cdb4ca3e7202cc123b24bd472e1c0/css.escape.js#L21) though :(

splash/qtrender_lua.py
```diff
@@ -234,8 +251,15 @@ def exceptions_as_return_values_wrapper(self, *args, **kwargs):
...
                 res = (b'return',) + ensure_tuple(res)
+            if error_as_flag and (result_as_flag or res[1] is None or res[1] is False):
+                res = add_flag(res, True)
         except Exception as e:
-            res = (b'raise', repr(e))
+            if error_as_flag and self.FLAG_EXCEPTIONS is not None \
+                    and any(isinstance(e, x) for x in self.FLAG_EXCEPTIONS):
+                res = (b'return', False, repr(e).encode('utf-8'))
+            else:
+                res = (b'raise', repr(e))
```
kmike: it looks like sometimes exception repr is passed as bytes (`.encode('utf8')`), and sometimes as a str (`repr(e)`) - why is that?

splash/qtrender_lua.py
```diff
@@ -313,49 +360,39 @@ def __init__(self, splash, source):
...
     @exceptions_as_return_values
     @can_raise
     @emits_lua_objects
+    @create_html_elements_for_nodes
     @decodes_lua_arguments('utf8')
     def __call__(self, *args):
         expr = "eval('('+{func_text}+')')({args})".format(
             func_text=self.source,
-            args=escape_js(*args)
+            args=','.join([arg.element_js if isinstance(arg, HTMLElement) else escape_js(arg) for arg in args])
```
kmike: could you please extract this line to an utility function? It is used more than once

splash/qtrender_lua.py
mike1808: utf-8 encoding is redundant

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,743 @@
...
+element:form_values
+-------------------
+
+Return a table with form values if the element type is *form*
+
+**Signature:** ``values, reason = element:form_values()``
+
+**Returns:** ``values, reason`` pair. If ``values`` is nil then error happened during the function call
+or node type is not *form*; ``reason`` provides an information about error type; otherwise
+``values`` is a table with element names as keys and values as values.
```
kmike: I just realized that this data format won't work if there are several elements with the same name. Browsers send all key-value pairs in this case. This is a common pattern in PHP world:

```
<input name="foo[]" value="bar">
<input name="foo[]" value="egg">
<input name="foo[]" value="spam">
```

-- php even decodes the submission above to a PHP array.

Having a single value per key is more common, but I think we should support multiple values as well, because it is a perfectly standard feature. What do you think about adding `multi` flag to this function? When true, values will be arrays, not bare values. It'd be nice to have it also supported in `element:fill`, but it can be more tricky; maybe that's fine to just document this gotcha for `element:fill`.

splash/qtrender_lua.py
mike1808: actually, it isn't redundant, it's necessary. When I call `encode` function my string becomes a raw bytes list which is needed for py -> lua conversation.

docs/scripting-element-object.rst
mike1808: Hm, in that case what about the following forms? 

```html
<input name="names[].name" value="mikael" />
<input name="names[].surname" value="manukyan" />
```

mike1808: and in general, we need deserialization of this https://api.jquery.com/serialize/

kmike: These are different names, so they work as-in with the current code. I'm not proposing to handle `[]` in element names explicitly (it is a PHP thing), but we should handle multiple elements with the same `name` attribute.

mike1808: hm, this method supports the multiple values of the same name
I should change `element:fill()` to support it.

mike1808: I'm going to use casperjs methods for filling forms:
https://github.com/casperjs/casperjs/blob/512ff47bae055ca4f9b3321eb6e799b874f85af7/modules/clientutils.js#L910

kmike: aha, right! 

I'm not a fan of casperjs API here though - if value can be either a single value or a list then user code must always handle both cases if it uses the result. I.e. it can't always iterate over a value or use a value as-is, it always needs to check value type. With 'multi' flag user code may only handle a single case. But I'm on fence with it here because there is code available for this worse API :)

mike1808: Done, check `test_fill_complex` test case.

mike1808: Oops, I made a stupid thing :) Now user won't be able to set array value to the input. I will fix that by adding `multi` flag. Also, what do you think about treat.as_array usage? Should we consider that user tables are arrays by default?

kmike: yep; treat.as_array should be only required to suggest how to serialize values to JSON at return time

mike1808: and how should I distinguish between arrays and dictionaries?

```lua
form:fill({
  foo = { 'a', 'b', 'c' }
  bar = { 0: '1', foo = 'bar' }
})
```

kmike: hm, what does `bar = { egg = 'spam' }` mean? All values are text, right?

mike1808: in jQuery it will serialized to `bar.egg = spam`

kmike: Didn't know about that. It looks like a jQuery thing, there is nothing in standards about that; imho we shouldn't support it.

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,824 @@
...
+------------
+
+Fill the form with the provided values
+
+**Signature:** ``ok, reason = element:fill(values, multi=false)``
+
+**Parameters:**
+
+* values - table with input names as keys and values as input values
+* multi - bool which indicates that the form has input with the same name and its values are
```
kmike: hm, I think multi should be a flag of element:form_values, not a flag of element:fill because element:fill can check data format itself. 

Also, it shouldn't depend on whether the form has inputs with the same name or not - user may choose to use 1-item lists for values for consistency.

kmike: Aha, so the tricky case is `<select multiple>`? For me it looks fine that `{foo={'bar', 'baz'}` means either "set these values of `<select multiple>`" or "set these values on consequent inputs named `foo`" because in the end the same data will be submitted via HTTP in both cases: it will be `foo=bar&foo=baz`.

mike1808: No, the tricky part is described below in the examples.

splash/browser_tab.py
```diff
@@ -1116,6 +1198,176 @@ def log(self, message, min_level=None):
...
+        frame = self.parent().web_page.mainFrame()
+        eval_expr = u"eval({})".format(escape_js("""
+        (function() {{
+            var storage = window[{storage_name}];
+
+            storage.events = {{}};
+
+            storage.callMethod = function(methodName) {{
+                return function(eventId) {{
+                    var storage = window[{storage_name}].events;
```
kmike: a nitpick: it may be better to use another name, which doesn't shadow a name from the outer scope

docs/scripting-element-object.rst
kmike: Example 4 doesn't make sense to me, why does it join values from a table? "A single text/int value => set a single form field, a table with values => set multiple form fields" sounds like a better approach. It also means values used in element:fill() correspond more closely to what'd be sent by a browser after submitting a form.

splash/tests/test_execute_element.py
```diff
@@ -0,0 +1,1416 @@
...
+
+            return body.node.lastChild.node.id
+        end
+        """, {"url": self.mockurl("various-elements")})
+
+        self.assertStatusCode(resp, 200)
+        self.assertEqual(resp.text, 'mydiv')
+
+    def test_field_value_emtpy(self):
+        pass
```
kmike: test_field_value_empty is empty

kmike: Сould you please add a test for HTML elements after a redirect:

1) what happens with old elements after another splash:go?
2) what happens with old elements if a redirect is to the same page?
3) what happens with old elements if a redirect is initiated by a page itself (e.g. in /jsredirect resource)?

mike1808: I can answer you here: as all of our storages will be cleared (because `javaScriptWindowObjectCleared` signal will be raised) all of the element's methods will raise an exception or return false flag (

kmike: Yeah, this is a good behaviour, but it'd be a nice to have an explicit test for that.

kmike: 1st test checks that it works as expected, the second checks that we work on element instances and not e.g. store unique css selectors; 3rd test checks that we don't rely on splash:go and other methods to clear elements state

docs/scripting-element-object.rst
mike1808: Well, I understood what did you mean when asked for `mutli` flag in `fill` method.

Also, in CasperJS there is the following feature:

If the form has only one `checkbox` element with the same name then it returns boolean indicating whether it's checked or not.
If the forma has several `checkbox` elements with the same name it returns the list of the values of `value` attribute.

Do we need this in Splash?
I suggest the following:

- when `multi= nil | false` : return the boolean
- when `multi= true` :  return the value of `value` attribute

kmike: Sounds good! 

I'm not sure that with all this logic `multi` is still a good name for an argument. And ideas? Something like 'values = first|auto|list'?

docs/scripting-element-object.rst
```diff
@@ -0,0 +1,824 @@
...
+
+
+.. _splash-element-fill:
+
+element:fill
+------------
+
+Fill the form with the provided values
+
+**Signature:** ``ok, reason = element:fill(values, multi=false)``
```
kmike: it should use `{}` braces if there are keyword arguments

docs/scripting-element-object.rst
mike1808: > Something like 'values = first|auto|list'?

👍

docs/scripting-element-object.rst
mike1808: I know, but the problem in this method is that it doesn't support named arguments because it's first value is a table (lupa bug). 
Anyway, I am going to remove this parameter.

kmike: makes sense!

splash/tests/test_execute_element.py
mike1808: I've added some tests (see the last two), but  I don't understand what do you want.

>the second checks that we work on element instances and not e.g. store unique css selectors
like if we change the refresh the page and try to access some property and get error, right?
>3rd test checks that we don't rely on splash:go and other methods to clear elements state
should I check that I'm getting error on various type of page change (js redrect, splash:go)?

kmike: Existing tests are fine, thanks!

kmike:
It looks like the only item in TODO for this PR is to change the way DOM objects are passed from JS for Lua/Python, to fix issues like https://github.com/scrapinghub/splash/pull/490/files#r75652620 -  `result_protection=False` shouldn't be used in most cases. Is there anything else left @mike1808 @immerrr?

mike1808:
Also, I should make it work:

```lua
element.onclick = function(event)
    id =  event.target.node.id
end
``

kmike:
Merged in #531. Thanks @mike1808 for your work on this!

Issue #491, opened 2016-08-04
Docker image dbus issue


Mandeuki:
When running scrapy splash with docker I got that error message  :
   failed to open "/etc/machine-id": no such file or directory

I fixed it by running : 

`apt-get update`
`apt-get install dbus`
`dbus-uuidgen > /etc/machine-id`

I don't really from where that bug come, maybe add a fix to the Dockerfile ?

kmike:
The log message is annoying but harmless. Ideally we should strip dbus from qt (it is not needed AFAIK). Using dbus-uuidgen is an option, but it should be called on stratup, not at the time Docker container is built. From https://dbus.freedesktop.org/doc/dbus-uuidgen.1.html:

> The UUID should be different on two different virtual machines, because there are two different kernels.

PR to fix that is welcome :)

PR #493, opened 2016-08-10
Dbus dependency

Mandeuki:
Fix to issue #491


Dockerfile
```diff
@@ -31,11 +31,9 @@ VOLUME [ \
...
-    "--js-profiles-path", "/etc/splash/js-profiles", \
-    "--filters-path", "/etc/splash/filters", \
-    "--lua-package-path", "/etc/splash/lua_modules/?.lua" \
-]
+# Copy start script
+COPY bin/startup.sh /home/root/startup.sh
+RUN chmod +x /home/root/startup.sh
+
+# Start up script (Generate machine UUID)
+ENTRYPOINT ["/home/root/startup.sh"]
```
kmike: I think it is not the same as before: previously it was possible to pass command-line arguments to Splash when it is executed from a Docker container (see http://splash.readthedocs.io/en/stable/install.html#passing-custom-options); it looks like you need to handle extra arguments in a script.

Mandeuki: Ok I will look at it and commit again later.

kmike: Thanks! Other than that, the PR looks very clean.

PR #495, opened 2016-08-12
Store private fields on objects (take 2)

immerrr:
This is another take at what #487 was trying to accomplish.

The idea is to have the object represented by an empty (public) table forwarding all field accesses via a metatable to the real (private) table . The metatable would reject all unwanted accesses and the object methods would work on the private instance directly and thus have unhindered access to private fields.

"Privateness" is determined by a certain prefix, as before, but I had to change it to keep "private_mode" property name, so now it's "_" that marks a property or method as private.

I have also adjusted "is_wrapped" detection a bit. previously it would go and check the object's metatable for `__wrapped` field, but now the metatable is hidden, so I'm checking if the metatable is replaced by a certain placeholder.

### DIFF
--- a/splash/lua_modules/wraputils.lua
+++ b/splash/lua_modules/wraputils.lua
@@ -84,10 +84,9 @@ local function sets_callback(func, storage)
 end
 
 
-local PRIVATE_PREFIX = "private_"
-
-local function is_private_name(key)
-  return string.find(key, "^" .. PRIVATE_PREFIX) ~= nil
+local function is_private_name(name)
+  -- Method/attribute name is private true if it starts with an underscore.
+  return name:sub(1, 1) == "_"
 end
 
 
@@ -120,8 +119,7 @@ local function setup_commands(py_object, self, private_self)
     command = unwraps_python_result(command, nlevels)
 
     if is_private_name(key) then
-      local short_key = string.sub(key, PRIVATE_PREFIX:len() + 1)
-      private_self[short_key] = command
+      private_self[key] = command
     else
       -- avoid custom setter
       rawset(self, key, command)
--- a/splash/lua_modules/wraputils.lua
+++ b/splash/lua_modules/wraputils.lua
@@ -156,7 +156,7 @@ local function wrap_exposed_object(py_object, private_self, cls)
   setup_property_access(py_object, private_self)
 
   -- "Public" metatable that prevents access to private elements and to itself.
-  public_mt = {
+  local public_mt = {
     __index = function(self, key)
       if is_private_name(key) then
         return nil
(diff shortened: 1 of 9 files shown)


splash/tests/test_execute.py
```diff
@@ -1217,7 +1217,7 @@ def test_private_jsfunc_not_available(self):
     def test_private_jsfunc_attributes(self):
         resp = self.request_lua("""                                      -- 1
         function main(splash)                                            -- 2
-            local func = splash:private_jsfunc("function(){return 123}") -- 3 <-
+            local func = splash:_jsfunc("function(){return 123}") -- 3 <-
```
kmike: could you please re-indent line number comments?

splash/lua_modules/wraputils.lua
```diff
@@ -150,13 +143,73 @@ local function setup_property_access(py_object, self, cls)
...
+    __metatable = EXPOSED_OBJ_METATABLE_PLACEHOLDER,
+  }
+
+  -- Forward any metatable events not defined in the "public" table to the
+  -- actual class/metadatable.
+  setmetatable(public_mt, {__index = cls})
+
+  -- public_self should only contain a reference to the public metatable
+  -- forwarding all actual data to the "real" self object.
+  public_self = {
```
kmike: why not `local public_self`?

immerrr: good catch, my lua instincs have dwindled :)

splash/tests/test_execute.py
immerrr: done

mike1808:
Mind blown solution. I'm amazed how this works.
👍

mike1808:
Shouldn't this [property](https://github.com/scrapinghub/splash/blob/97504b38188dc4a0771c6c09aef467d5bae0c691/splash/qtrender_lua.py#L456) be changed?

immerrr:
This property was the reason I went for changing the "private" prefix from "private_" to "_", what do you have in mind?

mike1808:
I thought about decorator for marking the methods as private, but I remembered that in Lua you cannot add a new property to a function value.

immerrr:
Just tried to move method wrapping into object wrapping stage, but emulation tests started to fail, because they relied on modifying `Splash` class after the `splash` object was created. If a sandboxed piece of code could do `require("splash")` and mutate the class in the same manner, I think it caught a proper security problem.

If the class is not normally available, I guess we can move back to "on-access" wrapping and just disallow setting properties on `private_self` except for when it is done via property setters.

@kmike, what do you think?

kmike:
@immerrr sandboxed code can't run `require("splash")`, but scripts stored on server can. Scripts stored on server are not sandboxed, so doing unsafe things in them is fine. There shouldn't be a way to get Splash class from a script submitted via HTTP API.

immerrr:
Another quick stab at fixing this: I have made `wait_restart_on_redirects` public now that it's test-only.

Issue #512, opened 2016-10-14
Chrome blocking retrieval of jquery in the UI


dennisfischer:
Chrome is blocking the download of jquery resource files in the splash UI.
I'm accessing the splash UI during development from **localhost**.
The jquery files are linked with **https://**, thus the request will be blocked with response
**net::ERR_INSECURE_RESPONSE**.

PR #513, opened 2016-10-14
Changed jquery resources to respect location protocol

dennisfischer:
Fix for #512


kmike:
Thanks @dennisfischer!

PR #522, opened 2016-11-02
Fixed Core Dumped with bad encoded statusText (Fixes #521)

### LINKED ISSUE #521: Core Dumped with bad encoded statusText
report

```
2016-10-31 19:27:48.122537 [-] "127.0.0.1" - - [31/Oct/2016:18:27:47 +0000] "POST /render.json HTTP/1.1" 200 22263 "-" "python-requests/2.11.1"
  File "splash/network_manager.py", line 334, in _on_reply_headers
    har.store_reply_headers_received(self._get_request_id(request), reply)
  File "splash/har_builder.py", line 165, in store_reply_headers_received
    entry["response"].update(reply2har(reply))
  File "splash/har/qt.py", line 125, in reply2har
    res["statusText"] = bytes(status_text).decode('latin1')
UnicodeEncodeError: 'ascii' codec can't encode characters in position 18-19: ordinal not in range(128)
Aborted (core dumped)
```

The `repr()` of `status_text` before the error is `u'Przekierowanie sta\xc5\x82e'`

Reproducible, but I can't disclose the URL publicly. If someone whats it I can send it through email.

gabrii:
It's just ignoring the error. A better solution would be a general per request exception handler, which would return an error through the API, instead of Core Dumping everything.

### DIFF
--- a/splash/har/qt.py
+++ b/splash/har/qt.py
@@ -122,7 +122,10 @@ def reply2har(reply, content=None):
         try:
             res["statusText"] = bytes(status_text, 'latin1').decode('latin1')
         except TypeError:
-            res["statusText"] = bytes(status_text).decode('latin1')
+            try:
+                res["statusText"] = bytes(status_text).decode('latin1')
+            except ValueError:
+                res["statusText"] = ""
 
     else:
         res["statusText"] = REQUEST_ERRORS_SHORT.get(reply.error(), "?")


kmike:
Thanks for the PR!

Hm, I think a root cause if that status_text is unicode. The code was intended to be permissive: bytes_object.decode('latin1') should always succeed. But if status_code is already unicode there is no need to decode it. It seems the current code expects it to be either bytes (so that it is passed as-is) or QByteArray (so that bytes(..) converts it from PyQT object to a regular Python object). The whole 

```py
try:
    res["statusText"] = bytes(status_text, 'latin1').decode('latin1')
except TypeError:
    res["statusText"] = bytes(status_text).decode('latin1')
```
doesn't look good, because it is not clear which type do we expect, it looks more like ad-hoc fixing until it works.

I'm surprised we're getting an unicode object here; status_text is `reply.attribute(QNetworkRequest.HttpReasonPhraseAttribute)`, and qt [docs](http://doc.qt.io/qt-5/qnetworkrequest.html) say that this attribute returns QByteArray.

kmike:
Thanks for debugging it, it was the hard part of this issue. I've added tests and commited a fix.

Issue #497, opened 2016-08-15
tests fail with Twisted 16.3.0


kmike:
test_client_disconnects.py tests fail for me with Twisted 16.3.0, but pass with Twisted 16.2.0 and 16.1.x. 

Currently Docker image uses Twisted 16.1.1, so it doesn't affect dockerized Splash (yet).

ablepharus:
https://twistedmatrix.com/trac/ticket/8692 has been fixed july  [1](https://github.com/twisted/twisted/pull/833) and should be available in twisted-17.9.

kmike:
Awesome, thanks for bringing this up @ablepharus! I wan under impression issue is "wontfix" and was not watching it.

PR #539, opened 2016-11-30
update twisted to 16.6.0

kmike:
See #497.


kmike:
It looks like the reason is https://twistedmatrix.com/trac/ticket/8692

PR #540, opened 2016-12-01
Fix link to document.querySelector() docs

redapple:
Alternative link: https://dom.spec.whatwg.org/#parentnode


redapple:
Another one: https://www.w3.org/TR/selectors-api/#findelements

kmike:
I like MDN docs, thanks @redapple!

Issue #91, opened 2014-04-17
Splash returns 504 Timeout error when server returns a response with incorrect or unsupported Content-Type header


kmike:

zyxue:
I had a similar problem, and fixed but not loading that script at all using the [allowed_domains](https://splash.readthedocs.org/en/latest/api.html#render-html) option. I think the filters option may have similar effect.

theotheo:
I think I faced the same issue. @zyxue What script don't you loading with allowed_domains option?

zyxue:
You need to look into the html returned and find out which script is relevant to your problem, and which isn't. I don't think there is a general solution for everybody. It depends on the project you're working on.

kmike:
It is still an issue in Qt5.

starrify:
### Information on this issue

The root cause of this issue is believed to be that the `SplashQWebPage` class does not handle it correctly when encounters content it cannot render. Two common cases that can trigger this issue:

1. When `QWebPage` receives a reply that has an unsupported `Content-Type` header.
2. When `QWebPage` receives a reply that has the header `Content-Disposition: attachment`.

A real world browser would popup a dialog and prompt the user to save the content in such scenarios.


### Suggested solutions

The `SplashQWebPage` class shall enable its property [`forwardUnsupportedContent`](http://doc.qt.io/qt-5.5/qwebpage.html#forwardUnsupportedContent-prop) and register handlers for signal [`unsupportedContent`](http://doc.qt.io/qt-5.5/qwebpage.html#unsupportedContent). It's also suggested to add another parameter to specify whether Splash shall keep the headers only, or download the full content which might be large.

### Other notes

Some project that I involve in requires this issue to be fixed, or the results might be incorrect in specific conditions. Thus I would like to spend some time trying to fix it. However it's not of great urgency and I'm not planning to do it right now.  
If you're going to work on the fix, please leave a message in case we do the same thing twice. :) (I'll be leaving messages if I start, too)

wpxgit:
Hi,

because i need that feature i've decided to make a job post with the target to contrubute to this project. If anybody is interested - just attend.
https://www.upwork.com/jobs/~017b5635b07e5eb885

Thanks

mohamed--abdel-maksoud:
Hi,

I'm working on a fix to this issue (sponsored by @wpxgit 's SearchVIU ). I would like to take your input before creating the PR. I'm at the point where rendering unsupported format raises the following error:
`
[...]

PR #560, opened 2017-01-21
handling unsupported content as per issues/91

mohamed--abdel-maksoud:
ref: https://github.com/scrapinghub/splash/issues/91


mohamed--abdel-maksoud:
Hello, I'd appreciate some help with an issue here:
on unsupported_content_handler (line 129 in browser_tab.py), I try to read the whole response and write it into a file. reply.readAll() doesn't seem to return the whole file, so the download is always corrupt ... any idea what I do wrong?

Thanks!

splash/browser_tab.py
```diff
@@ -95,6 +100,53 @@ def _init_webpage(self, verbosity, network_manager, splash_proxy_factory,
         self.web_page.setNetworkAccessManager(network_manager)
         self.web_page.splash_proxy_factory = splash_proxy_factory
         self.web_page.render_options = render_options
+        
+        
+        def unsupported_content_handler(reply):
```
kmike: could you please make it a method? You can pass different arguments using e.g. functools.partial

kmike:
reply.readAll() likely doesn't work because it can be called only once, and qt may also need to call it to read the response. There is a workaround - keep track of chunks which are received and build a body from there, see e.g. https://github.com/scrapinghub/splash/pull/494/ implementation.

splash/browser_tab.py
```diff
@@ -129,10 +139,15 @@ def unsupported_content_handler(self, reply):
             
             fname = os.path.join(self.download_directory, fname)
             
-            with open(fname, 'wb+') as f: f.write(content)
-            
-            self.web_view.setHtml("<html><head /><body><h1>file was downloaded from %s to %s</h1></body>"%(url, fname))
+            # NOTE trying to retrieve the already downloaded chunks, cannot find the request in _response_bodies
+            #req_id = self.http_client.network_manager._get_request_id(reply.request())
+            #bodies = self.http_client.network_manager._response_bodies[req_id]
+            #log.msg('request id: %d, %d chunks' %(req_id, len(bodies)))
```
mohamed--abdel-maksoud: I tried to retrieve the already downloaded chunks, but cannot find them in response_bodies

mohamed--abdel-maksoud:
Thanks for the tips! I still cannot download unsupported files in an optimal way, please see the comment at browser_tab.py:142

splash/browser_tab.py
kmike: is response_body_enable set to True? See http://splash.readthedocs.io/en/stable/scripting-ref.html#splash-response-body-enabled and http://splash.readthedocs.io/en/stable/api.html#arg-response-body

mohamed--abdel-maksoud: Thanks for the tip!
I enabled the property in defaults.py then wrote the request's response body to a file as below, it still writes a part of the file (just as it did with reply.readAll). I scanned the code to see what might be wrong, didn't find any hints.
`
# defaults.py
#
RESPONSE_BODY_ENABLED = True

# browser_tab.py
#
            req_id = self.http_client.network_manager._get_request_id(reply.request())
            bodies = self.http_client.network_manager._response_bodies[req_id]
            log.msg('request id: %d, %d chunks %s' %(req_id, len(bodies), type(bodies[0]) ))
            
            with open(fname, 'w+b') as f: 
                f.write(bytearray(bodies))
`

wpxgit:
Hi,

i've thought about it. Is it possible that that is the standard behavior of qt5. If i think about my browser he tries to get the file, if it realise that its a file for download it opens download manager and begins to download that file.
I've read a little bit in qt5 discussions and it seems that the solution for an unsupported file is to download download it with a separate QNetworkAccessManager. 

And it makes a little sense - the browser needs the first lines of the file to recognize if its an html file or an image, gzipped or else...
?

mohamed--abdel-maksoud:
Hi @wpxgit ,
Indeed, great insight. I just tried the code on pdfs of several sizes and the unsupported type handler is called after downloading ~1.2 KB of the file, regardless of the whole file size.
Can we then conclude that issuing an http_get on the unsupported file (i.e. the last commit) carries very little overhead? would this approach be acceptable, @kmike ?

wpxgit:
any chance to get this pull request merged?

Issue #559, opened 2017-01-20
Current version of Docker working for proxies? 


wflanagan:
I'm not sure if I'm doing something wrong.. but I cannot get splash to work through either Crawlera Proxies, or TrustedProxies (our current proxy provider), both commercial providers (yes, I'm aware that Crawlera is YOUR proxy service). 

Splash is rejecting the request, saying an authentication error.  I have literally spent all day working on this.. and have triple verified that the authentication parameters are correct. These are used in production, running thousands upon thousands of requests per day. 

I have tried using the .ini method of initialization of proxies, as well as passing the URL. I am using the render.xxx method. All destinations tested, without proxies, pass through Splash without problems.  

I'm using an off the shelf Docker image... using docker compose.  I have a Gist that has my config in it (neutralized for username password and IP addresses). 

Again, I have confirmed that the username/password are correct. 

I'd love to use this, but without proxies, it's a no go.  Help? 

https://gist.github.com/wflanagan/2141dd9605948a435c81881694c4de7b

P.s. I can't seem to add a "question" label. I would if I could.

kmike:
Sorry, I'm not sure how to reproduce it with Crawlera, and I don't have a TrustedProxies account. The following works for me:

```py
splash_url = 'http://<MY_SPLASH_SERVER>:8050/render.html'
resp = requests.post(splash_url, json={
    'url': 'http://httpbin.org/ip', 
    'proxy': 'http://<CRAWLERA_API_KEY>:@proxy.crawlera.com:8010'}
)
print(resp.text)  # the IP is changed when proxy is enabled
```

wflanagan:
Hi,

This is using the render.html endpoint? I was talking about this all in the
context of the render.html. endpoint.

I will try to figure out scripting today.

kmike:
Hey @wflanagan,

Yes, I used render.html and passed `'http://<CRAWLERA_API_KEY>:@proxy.crawlera.com:8010'` in the 'proxy' argument. I used a JSON POST request for render.html endpoint: body is a json-encoded `{'url': '...', 'proxy': '...'}` dict, and Content-Type header is set to `application/json`. 

The script I've pasted does that; it is not a Splash script, but a Python script which uses [requests](http://docs.python-requests.org/en/master/) library to send an HTTP request. If you're using Ruby there must be a way to do the same. 

[...]

PR #561, opened 2017-01-24
Fixed proxy auth

kmike:
This should fix https://github.com/scrapinghub/splash/issues/559. https://github.com/scrapinghub/splash/pull/348 fixed proxy auth for proxies set with `request:set_proxy`, this PR fixes it for proxies set using `proxy` argument as well.


kmike:
Tests in this PR don't expose the bug, they pass without the fix. What fails is auth in case of HTTPS proxying, but I haven't added a HTTPS proxy to mockserver, so tests which should fail are skipped. 

I checked the changes on a real https proxy (Crawlera); it was not working before the fix, but started to work after it.

PR #568, opened 2017-02-17
Adds validator for downloading file while installing pyQt5

vedantrathore:


vedantrathore:
@kmike please check this :)

dockerfiles/splash/provision.sh
```diff
@@ -102,7 +102,10 @@ install_pyqt5 () {
     chmod a+rw /downloads && \
     curl -L -o /downloads/sip.tar.gz http://sourceforge.net/projects/pyqt/files/sip/sip-${SPLASH_SIP_VERSION}/sip-${SPLASH_SIP_VERSION}.tar.gz && \
     curl -L -o /downloads/pyqt5.tar.gz http://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-${SPLASH_PYQT_VERSION}/PyQt-gpl-${SPLASH_PYQT_VERSION}.tar.gz && \
-    # TODO: check downloads
+    if [ -f '/downloads/sip.tar.gz' ] && [-f '/downloads/pyqt5.tar.gz' ]; then
+        echo "Error downloading"
+        exit 0
+    fi
```
kmike: I think the implementation has a couple of issues:

1. exit 0 is a "successful" exit code, while here an error happens;
2. commands are joined with `&&` in order to make the whole function fail if one of commands fail. But you're not joininig `if` with further commands. It means that e.g. if activate_venv fails then `mkdir -p /builds` is still executed, and the whole script is not finished.

vedantrathore: @kmike Okay, sorry about the delay, I'll fix those now.

immerrr: There's also a problem with the second condition: `-f` should be detached from `[`, otherwise it tries to invoke a `[-f` command and will most likely fail.

Also, right now the code checks if both files exist (downloaded successfully) and tries to raise an error if that's the case.

Youngpans:
How to install python on Windows7 pls

PR #571, opened 2017-03-03
update psutil and pillow in Docker image

kmike:


kmike:
Pillow is updated in https://github.com/scrapinghub/splash/pull/861.

PR #588, opened 2017-03-31
Submit UI form in new tab

gabrii:
I think this way is better, otherwise the current script gets lost after submit.


gabrii:
Sorry, I forgot to update my fork. Will fix pull request later today.

kmike:
I'm not sure I follow - for me the script is preserved in the editor after submitting. Is it not the case for you? Are there JS error logs in the console?

PR #589, opened 2017-03-31
Added code snippet to explain how to use splash behind proxy

rava-dosa:


splash/examples/proxyEg.lua
```diff
@@ -0,0 +1,19 @@
+function main(splash)
```
lucywang000: I think "use-proxy.lua" would be a better name (the convention in the `examples/` folder is to use dash-separated file name, not camel case).

splash/examples/proxyEg.lua
```diff
@@ -0,0 +1,19 @@
...
+  splash:on_request(function(request)
+    request:set_proxy{
+        host = "10.3.100.207",
+        port = 8080,
+        username = splash.args.username,
+        password = splash.args.password,
+    }
+end)
+  local url = splash.args.url
+  assert(splash:go(url))
```
lucywang000: It doesn't make sense to create a local variable and only use it once. Simply `assert(splash:go(splash.args.url))` would do.

PR #596, opened 2017-04-13
Update manual install instructions

zscholl:
It took me an unreasonable amount of time to figure out how to do a manual splash install. 

I ended up finding [this](https://hub.docker.com/r/scrapinghub/splash/builds/bog9kxd7dpkwkfjjdkyhxd/) build log to figure it out. 

Thought this may benefit others who are looking to manually install splash. This should be moved into a script in the future.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -77,14 +77,17 @@ Ubuntu 14.04 (manual way)
         $ sudo ./provision.sh install_pyqt5
        
         $ sudo ./provision.sh install_python_deps
-       
-        $ sudo ./provision.sh remove_builddeps
-       
-        $ sudo ./provision.sh remove_extra
-       
+        
         Change back to the parent directory of splash, i.e. `cd ~`
        
         $ sudo pip3 install splash/
+        
+3. **OPTIONAL** Remove packages unecessary for Splash:
+        
+        $ sudo ./provision.sh remove_builddeps
+
+        $ sudo ./provision.sh remove_extra
+
 
 To run the server execute the following command::
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -81,12 +81,6 @@ Ubuntu 14.04 (manual way)
         Change back to the parent directory of splash, i.e. `cd ~`
        
         $ sudo pip3 install splash/
-        
-3. **OPTIONAL** Remove packages unnecessary for Splash:
-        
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -56,17 +56,35 @@ OS X + Docker
...
+        
+        $ sudo ./provision.sh install_extra_fonts
+       
+        $ sudo ./provision.sh install_pyqt5
+       
+        $ sudo ./provision.sh install_python_deps
+       
+        $ sudo ./provision.sh remove_builddeps
+       
+        $ sudo ./provision.sh remove_extra
```
kmike: I think remove steps shouldn't be in these docs because they may remove something user installed.

zscholl: Yeah good point, I'll add an "optional" header before those steps.

zscholl: Fixed

kmike: These steps are really aggressive - they are to trim Docker image size. For example, remove_extra removes all man files (documentation) from the system. I think they shouldn't be used on user machines, this is dangerous.

zscholl:
How's that @kmike?

docs/install.rst
```diff
@@ -56,17 +56,32 @@ OS X + Docker
...
+        
+        $ sudo ./provision.sh install_builddeps
+        
+        $ sudo ./provision.sh install_deps
+        
+        $ sudo ./provision.sh install_extra_fonts
+       
+        $ sudo ./provision.sh install_pyqt5
+       
+        $ sudo ./provision.sh install_python_deps
```
kmike: https://github.com/scrapinghub/splash/blob/master/Dockerfile also installs Flash; is there a reason `install_flash` is not in a list?

These commands can be combined in a single command; do you prefer separate provision.sh calls? I don't mind either way.

zscholl: Oops, `install_flash` was omitted by mistake. 

I haven't tested it, but from looking at the script it looks like it will properly operate via:
`sudo ./provision.sh prepare_install install_msfonts install_builddeps install_deps etc..`

That is certainly more elegant, is that what you mean?

kmike: yep!

zscholl:
Alrighty, cool beans. Think that should be good to go!

kmike:
Thanks @zscholl!

PR #601, opened 2017-04-23
typo in index.rst

xioustic:


kmike:
A good catch, thanks @xioustic!

Issue #578, opened 2017-03-13
element:mouse_click() doesn't work when element is not in the page's visible area


stummjr:
I tried executing this:

```splash:select("li.next > a"):mouse_click(10,10)```

But it didn't work because the element wasn't visible in the viewport. This [method's documentation](http://splash.readthedocs.io/en/stable/scripting-element-object.html#element-mouse-click) doesn't mention this constraint, though.

Is there any chance of changing the default behavior to scrolling down to the element when `mouse_click()` is called on it?

kmike:
Fixed by #633.

PR #633, opened 2017-06-30
Element improvements

kmike:
TODO:

* [x] splash.scroll_position
* [x] element:png() and element:jpeg() should work for out-of-window elements;
* [x] element:mouse_click() should work for out-of-window elements (#578);
* [x] element:mouse_hover() should work for out-of-window elements
* [x] updated Element docs
* [x] don't use `.node` everywhere


kmike:
Tests are failing because of #635, but they pass for me locally (with a cached PyQT version), so I'm merging this pull request.

PR #679, opened 2017-09-26
fix doc `sudo provision.sh` => `sudo ./provision.sh`

codenoid:
with `sudo provision.sh` (not executed) => `sudo: provision.sh: command not found`


kmike:
Thanks @codenoid!

PR #713, opened 2018-01-05
Docs : missing /tmp/script.qs file in manual install

yalhyane:
I received the error '/tmp/script.qs file does not exist' when installing Splash manually and running ./provision.sh install_official_qt command


kmike:
Thanks @yalhyane!

PR #725, opened 2018-02-08
Added --xvbf_screen_size parameter.

deademo:


kmike:
Code looks good to me 👍 
When is setting xvfb screen size useful?

mksh:
It is really useful for retrieving information from websites, which try to implement fingerprinting measures in order to track scrapers.

Matching screen size with Javascript is one of the most common techniques to fingerprint user browsers today, and having an ability to change screen size could help one to avoid being tracked by website admins (or any thirdparty service which provides user tracking features), in case if the aforementioned measures take place.

kmike:
Thanks @deademo for the fix and @mksh for the explanation!

PR #749, opened 2018-03-14
status is actually status_code

macarthy:
update to the docs

### DIFF
--- a/docs/scripting-response-object.rst
+++ b/docs/scripting-response-object.rst
@@ -21,7 +21,7 @@ This field is read-only.
 
 .. _splash-response-status:
 
-response.status
+response.status_code
 ---------------
 
 HTTP status code of the response.


kmike:
Hm, I don't think this is correct; there are tests which show that attribute name is "status", not "status_code": https://github.com/scrapinghub/splash/blob/3cba485e603998e6325277d1298e98ace1378f09/splash/tests/test_execute_callbacks.py#L498.

macarthy:
is I try 

`response = requests.get('http://localhost:8050/render.html', params={'url': url, 'wait': 2, 'images': 0})
	return response`

Response has a status_code but not status 

status_code  is in the tests here : 


`

def assertStatusCode(self, response, code):
--
89 | msg = (response.status_code, truncated(response.text, 1000),
90 | response.url)
91 | self.assertEqual(response.status_code, code, msg)

`
and here 

https://github.com/scrapinghub/splash/blob/49f403e723e49d5019dfb2d78d91e75e38ff5cb1/splash/tests/stress.py#L62

kmike:
The documentation you refer to is for a Response object provided by Splash in a Lua script; Response object returned by `requests.get` is privided by requests Python library (we use `requests` only for tests and benchmarks, it is not used in core Splash itself). These are completely different objects, not even the same programming language :)

PR #808, opened 2018-09-02
Add callback for requests locked by lock_navigation

vfcosta:
The `on_navigation_locked` callback proposed here is useful to track requests even after calling `splash:lock_navigation()`.

### DIFF
--- a/docs/scripting-ref.rst
+++ b/docs/scripting-ref.rst
@@ -2783,3 +2783,49 @@ Example: select all ``<img />`` elements and get their ``src`` attributes
 
         return treat.as_array(srcs)
     end
+
+
+.. _splash-on-navigation-locked:
+
+splash:on_navigation_locked
+---------------------------
+
+Register a function to be called before a request is discarded when navigation is locked.
+
+**Signature:** ``splash:on_navigation_locked(callback)``
+
+**Parameters:**
+
+* callback - Lua function to call before a request is discarded.
+
+**Returns:** nil.
+
+**Async:** no.
+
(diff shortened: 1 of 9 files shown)


docs/scripting-ref.rst
```diff
@@ -2783,3 +2783,49 @@ Example: select all ``<img />`` elements and get their ``src`` attributes
 
         return treat.as_array(srcs)
     end
+
+
+.. _splash-on-navigation-locked:
+
+splash:on_navigation_locked
+-----------------
```
kmike: I think it may raise "underline is too short" error

docs/scripting-ref.rst
```diff
@@ -2783,3 +2783,49 @@ Example: select all ``<img />`` elements and get their ``src`` attributes
...
+Example 1 - log all URLs discarded using :ref:`splash-request-url` attribute:
+
+.. literalinclude:: ../splash/examples/log-locked-requests.lua
+   :language: lua
+
+
+.. _splash-on-navigation-locked-reset:
+
+splash:on_navigation_locked_reset
+-----------------------
```
kmike: it seems underline is too short

kmike:
Looks awesome, thanks @vfcosta! Code is good, docs and tests are in place 👍 

Would you mind fixing small issues with docs before a merge>

1) underlines should match text length;
2) it'd be great to have these new methods mentioned in https://splash.readthedocs.io/en/stable/scripting-overview.html

vfcosta:
@kmike done! Thanks for your feedback

kmike:
It was quick :) Thanks @vfcosta!

PR #811, opened 2018-09-21
Provide request body in request info.

### LINKED ISSUE #688: Request body not provided in request.info.
Currently, Splash does not provide the `postData` field in https://splash.readthedocs.io/en/stable/scripting-request-object.html#request-info.

Would a pull request to implement this be accepted?

ArturGaspar:
Fixes #688 

Besides the issues mentioned in the added documentation (non-standard HAR, request body not available in certain Lua functions), when this functionality is enabled it breaks the HAR viewer, which always expects postData with type "application/x-www-form-urlencoded" to have the "params" field.


ArturGaspar:
Oops I broke some tests, will reopen when I fix it.

docs/scripting-ref.rst
```diff
@@ -149,6 +149,24 @@ Enable or disable browser plugins (e.g. Flash).
...
+**Signature:** ``splash.request_body_enabled = true/false``
+
+By default Splash doesn't keep bodies of each request in memory. It means that
+request content is not available in :ref:`splash-request-info` and in HAR_
+exports. To make request content available to a Lua script set
+``splash.request_body_enabled = true``.
+
+Note that request data in :ref:`splash-request-info` is not available in the
+callback :ref:`splash-on-response-headers` or in the request of the response
+returned by :ref:`splash-http-get` and :ref:`splash-http-post`.
```
kmike: Where is it available then? You mentioned HAR, is it available in on_request callbacks?

ArturGaspar: Yes, and also in on_response (in response.request.info).

I was going to add a test for that now, but I hit what I believe to be a bug, which I reported as #814

ArturGaspar: Now it also has a test for on_response.

splash/network_manager.py
```diff
@@ -123,6 +123,38 @@ def createRequest(self, operation, request, outgoingData=None):
...
+                # In a sequential QIODevice, size() returns the value of
+                # bytesAvailable(), which is only the size of the data in the
+                # QIODevice buffer and not the total size of the output. Until
+                # a read attempt, the buffer is empty and size() returns zero.
+                #
+                # In requests generated by Qt WebKit, outgoing_data is a
+                # WebCore::FormDataIODevice object, which is sequential. Its
+                # getFormDataSize() method cannot be accessed through PyQt5,
+                # but Qt WebKit puts its value in the Content-Length header
+                # (see WebCore::QNetworkReplyHandler::getIODevice).
```
kmike: great observation 👍

splash/network_manager.py
```diff
@@ -123,6 +123,38 @@ def createRequest(self, operation, request, outgoingData=None):
             self.log(traceback.format_exc(), min_level=1, format_msg=False)
             return super(ProxiedQNetworkAccessManager, self).createRequest(operation, request, outgoingData)
 
+    def _get_request_body(self, request, outgoing_data):
+        if outgoing_data is not None:
```
kmike: It is often easier to read functions when special cases are handled first:

```py
if outgoing_data is None:
    return None
# ... the rest
```
this also provides more horizontal space (less nesting).

splash/network_manager.py
```diff
@@ -123,6 +123,38 @@ def createRequest(self, operation, request, outgoingData=None):
...
+            size = None
+            if outgoing_data.isSequential():
+                # In a sequential QIODevice, size() returns the value of
+                # bytesAvailable(), which is only the size of the data in the
+                # QIODevice buffer and not the total size of the output. Until
```
kmike: Do you know if tests trigger this behavior (data with a size larger than a buffer)? It seems new tests only use very small request bodies.

kmike:
Looks great, thanks @ArturGaspar!

PR #816, opened 2018-09-29
Update provision.sh

Lucioric2000:
Updated provision.sh to be able to install in Ubuntu headless systems.
When trying to install Splash in the system (not in a docker container because the docker containers begun being unresponsive after some time of scraping using Scrapy), I found some dependencies missing that stalled the installation on a Ubuntu 16-04 DigitalOcean Droplet.

### DIFF
--- a/dockerfiles/splash/provision.sh
+++ b/dockerfiles/splash/provision.sh
@@ -180,9 +180,7 @@ install_pyqt5 () {
         -e QtWebKit \
         -e QtWebKitWidgets \
         -e QtSvg \
-        -e QtPrintSupport
-#        -e QtPrintSupport && \
-    echo beforemake
+        -e QtPrintSupport && \
     make -j ${SPLASH_BUILD_PARALLEL_JOBS} && \
     make install && \
     popd  && \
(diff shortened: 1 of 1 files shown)


dockerfiles/splash/provision.sh
```diff
@@ -151,6 +151,7 @@ install_pyqt5 () {
     curl -L -o /downloads/pyqt5.tar.gz https://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-${SPLASH_PYQT_VERSION}/PyQt5_gpl-${SPLASH_PYQT_VERSION}.tar.gz && \
 #    curl -L -o /downloads/sip.tar.gz https://www.riverbankcomputing.com/static/Downloads/sip/sip-${SPLASH_SIP_VERSION}.tar.gz && \
 #    curl -L -o /downloads/pyqt5.tar.gz https://www.riverbankcomputing.com/static/Downloads/PyQt5/PyQt5_gpl-${SPLASH_PYQT_VERSION}.tar.gz && \
+    apt-get install -y --no-install-recommends libqt5svg5-dev libqt5webkit5-dev sip-dev && \
```
kmike: I wonder why you needed it - sip is installed later in this script, libqt5.. packages shouldn't be needed, as they're wrong versions (we install official qt, from qt websites). `install_pyqt5` command also looks like a wrong place to install libqt5svg5-dev and libqt5webkit5-dev. 

Could you share an error log?

Issue #817, opened 2018-09-30
interoperability with har2warc


wsdookadr:
Hi, I'm noticing some interoperability problems that I've [described here](https://github.com/webrecorder/har2warc/issues/1). In short, if I convert the .har obtained from splash, to a .warc and import it in webrecorder, the styling information is gone.

Please have a look.

Thanks,
Stefan

wsdookadr:
isolated the problem to #821 so I'm closing this issue

Issue #519, opened 2016-10-27
Cache doesn't save full HAR, visible on second request on single page with 200 status code


gabrii:
How to reproduce:
- Install splash==2.2.1
- Run "python -m splash.server"
- Visit http://localhost:8050/info?wait=0.5&images=1&expand=1&timeout=60.0&url=https%3A%2F%2Fraw.githubusercontent.com%2Fscrapinghub%2Fsplash%2Fmaster%2FLICENSE&lua_source=
- Note the HAR entry.
- Refresh page.
- Note there's no HAR entry.

Any setting or patch to disable the cache?

kmike:
Hey @gabrii,

There is a PR to allow disabling cache (see https://github.com/scrapinghub/splash/issues/203 and https://github.com/scrapinghub/splash/pull/339), but it has some problems, so it is not merged. I don't have plans to pick it up soon; contributions ar welcome!

gabrii:
Thanks for referencing the issues, I ended up patching a private fork with #339 for the moment.

nramirezuy:
@gabrii Your fork isn't working for me, are you still using it?

Gallaecio:
Closing in favor of https://github.com/scrapinghub/splash/issues/203

PR #821, opened 2018-10-07
[MRG+1] Added option to disable the webkit in-memory and network caches

wsdookadr:
Hi, I've been reading through the existing open issues.
This pull request is intended to address #203 , #519 , scrapy-plugins/scrapy-splash#168 , #817 , webrecorder/har2warc#1

Please review

Thanks,
Stefan

Note: To try this out, you can use this docker image:

    docker pull wsdookadr/splash:v1
    docker run -it -p 8050:8050  --name splash_v1 wsdookadr/splash:v1 --disable-browser-caches

And then load a web page and generate a .har file from it like so

    curl "http://localhost:8050/render.har?url=https://www.digitalocean.com/community/tutorials/how-to-install-java-with-apt-on-debian-9&timeout=13&wait=4&response_body=1" > /tmp/q.har

If you run this multiple times, the amount of data fetched by curl should be about the same (4mb for that url). Before this pull-request, the first time, it would produce around 4mb, and the subsequent ones about ~200-300k (the .har files would be missing resources that are cached by webkit in its page cache, the first time the page was loaded).

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -343,6 +343,19 @@ Please refer to `this great answer from kmike on reddit.
 
 .. _why-lua:
 
+Why are CSS styling and images missing from the .har archive?
+-------------------------------------------------------------
+
+Webkit has an in-memory cache (also `called page-cache <https://webkit.org/blog/427/webkit-page-cache-i-the-basics/>`)
+and a `network cache <http://doc.qt.io/qt-5/qnetworkrequest.html#CacheLoadControl-enum>`. 
+
+If you tell splash to load two pages that share some common resources,
+the second page's .har file will not contain the shared resources because
+they were cached through the page cache.
+
+If you want the .har file to contain all the resources for that page,
+run splasy with the command-line option ``--disable-browser-caches``.
+
 Why does Splash use Lua for scripting, not Python or JavaScript?
 ----------------------------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -341,7 +341,7 @@ Why was Splash created in the first place?
 Please refer to `this great answer from kmike on reddit.
 <https://www.reddit.com/r/Python/comments/2xp5mr/handling_javascript_in_scrapy_with_splash/cp2vgd6>`__
 
-.. _why-lua:
+.. _why-css-images:
 
 Why are CSS styling and images missing from the .har archive?
 -------------------------------------------------------------
@@ -354,7 +354,9 @@ the second page's .har file will not contain the shared resources because
 they were cached through the page cache.
 
 If you want the .har file to contain all the resources for that page,
-run splasy with the command-line option ``--disable-browser-caches``.
+run splash with the command-line option ``--disable-browser-caches``.
+
+.. _why-lua:
 
 Why does Splash use Lua for scripting, not Python or JavaScript?
 ----------------------------------------------------------------
--- a/splash/server.py
+++ b/splash/server.py
@@ -355,7 +355,7 @@ def _set_global_render_settings(js_disable_cross_domain_access, private_mode, di
     if disable_browser_caches is True:
         # disables the page cache (also known as the in-memory webkit cache)
         # see https://webkit.org/blog/427/webkit-page-cache-i-the-basics/
-        settings.setObjectCacheCapacities(0,0,0)
+        settings.setObjectCacheCapacities(0, 0, 0)
         settings.setMaximumPagesInCache(0)
 
 
--- a/splash/server.py
+++ b/splash/server.py
@@ -51,6 +51,8 @@ def parse_opts(jupyter=False, argv=sys.argv):
         help="disable Xvfb auto start")
     op.add_option("--disable-lua-sandbox", action="store_true", default=False,
         help="disable Lua sandbox")
+    op.add_option("--disable-browser-caches", action="store_true", default=False,
+        help="disables in-memory and network caches used by webkit")
     op.add_option("--lua-package-path", default="",
         help="semicolon-separated places to add to Lua package.path. "
              "Each place can have a ? in it that's replaced with the module name.")
@@ -178,6 +180,7 @@ def splash_server(portnum, ip, slots, network_manager_factory, max_timeout,
                   lua_sandbox_allowed_modules=(),
                   strict_lua_runner=False,
                   argument_cache_max_entries=None,
+                  disable_browser_caches=False,
                   verbosity=None):
     from twisted.internet import reactor
     from twisted.web.server import Site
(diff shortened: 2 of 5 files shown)


docs/faq.rst
```diff
@@ -343,6 +343,19 @@ Please refer to `this great answer from kmike on reddit.
...
+
+Webkit has an in-memory cache (also `called page-cache <https://webkit.org/blog/427/webkit-page-cache-i-the-basics/>`)
+and a `network cache <http://doc.qt.io/qt-5/qnetworkrequest.html#CacheLoadControl-enum>`. 
+
+If you tell splash to load two pages that share some common resources,
+the second page's .har file will not contain the shared resources because
+they were cached through the page cache.
+
+If you want the .har file to contain all the resources for that page,
+run splasy with the command-line option ``--disable-browser-caches``.
```
cathalgarvey: Minor typo; `splasy` vs. `splash`

splash/server.py
```diff
@@ -345,6 +352,12 @@ def _set_global_render_settings(js_disable_cross_domain_access, private_mode):
     settings.setAttribute(QWebSettings.PrivateBrowsingEnabled, private_mode)
     settings.setAttribute(QWebSettings.LocalStorageEnabled, not private_mode)
 
+    if disable_browser_caches is True:
+        # disables the page cache (also known as the in-memory webkit cache)
+        # see https://webkit.org/blog/427/webkit-page-cache-i-the-basics/
+        settings.setObjectCacheCapacities(0,0,0)
```
cathalgarvey: Minor pep8 gripe; spacing between arguments.

docs/faq.rst
```diff
@@ -343,6 +343,19 @@ Please refer to `this great answer from kmike on reddit.
 
 .. _why-lua:
 
+Why are CSS styling and images missing from the .har archive?
```
kmike: the whole section should be moved above `_why-lua:` label, because it is a label for the next section.

wsdookadr: yup, moved the section above `_why-lua:` and added its own label

kmike:
Hey @wsdookadr! The PR looks good, but could you please add a test case? Check e.g. [here](https://github.com/scrapinghub/splash/blob/a3f7e757b3bdbc9d3cb6f79c5bfc316a8a05ae69/splash/tests/test_request_filters.py#L149) how a custom startup option is tested. I guess you'd need to define some resource in https://github.com/scrapinghub/splash/blob/a3f7e757b3bdbc9d3cb6f79c5bfc316a8a05ae69/splash/tests/mockserver.py#L975 with caching behavior, then request it 

1) twice with a Splash started with default options - resource should be in har only for one of them,
2) twice with a Splash started with `--disable-browser-caches` - resource should be in har in both cases

wsdookadr:
@kmike I've pushed another commit with tests for this. However, it seems like the problem is not being captured by the test, in other words both default options and the new `--disable-browser-caches` options will bring all the entries with content in the HAR.
The new test can be run using the following command:

    pytest -s --verbosity=6 ./splash/tests/test_request_filters.py -k test_disable_browser_caches

splash/tests/mockserver.py
```diff
@@ -901,6 +901,22 @@ def render_GET(self, request):
             return base64.decodebytes(b'R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=')
 
 
+class SubresourcesWithCaching(Subresources):
+    """ Embedded css and image """
+
+    @use_chunked_encoding
+    def render_GET(self, request):
+        return ("""<html><head>
+                <link rel="stylesheet" href="subresources-with-caching/style.css" />
```
kmike: I think the problem could be that related resources don't set any specific cache headers, so the browser doesn't cache them. Could you try setting [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Caching_static_assets) header for these static resources?

wsdookadr: @kmike you are correct, I've pushed a new commit that sets that Cache-Control header and now it seems that at least for the image.gif resource, the entire scenario is reproducible.

splash/tests/test_request_filters.py
```diff
@@ -155,3 +170,47 @@ def test_file_scheme_can_be_enabled(self):
...
+            entry22 = self.findHAREntry(r.json(),r"^.*subresources-with-caching/style.css$")
+            entry23 = self.findHAREntry(r.json(),r"^.*subresources-with-caching/img.gif$")
+
+            return [[entry11,entry12,entry13],[entry21,entry22,entry23]]
+
+    # run just this test using
+    # pytest -s --verbosity=6 ./splash/tests/test_request_filters.py -k test_disable_browser_caches
+    def test_disable_browser_caches(self):
+        # entries if run with caches enabled
+        # in this case we should have  [[content,content,content],[content,content,no content]]
```
kmike: I wonder why isn't it working for css files. Maybe there are some other response headers needed to make it work, or maybe browser uses ETag or something like that, so there is no "full" response. It doesn't look good that opur code sets cache headers, but then checks that the resource is not cached; test will break if resource start to becached.

I think there are two ways to fix that:

1) Investigate what's going on and make sure css is cached as well;
2) remove a check for css file, check only html and gif.

wsdookadr: I agree with you on that and I went with the 2nd approach, I have removed the css checks. I've pushed a new commit for this.
For this pull request I'm ok with having it this way.

splash/tests/mockserver.py
```diff
@@ -901,6 +901,51 @@ def render_GET(self, request):
...
+    """
+
+    def getChild(self, name, request):
+        if name == b"style.css":
+            return self.StyleSheet()
+        if name == b"img.gif":
+            return self.Image()
+        return self
+
+    class StyleSheet(Resource):
```
kmike: Can it be removed now, as well as `<link rel=stylesheet>` tag in html?

splash/tests/mockserver.py
```diff
@@ -901,6 +901,39 @@ def render_GET(self, request):
             return base64.decodebytes(b'R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=')
 
 
+class SubresourcesWithCaching(Resource):
+    """
+        Embedded css and image.
```
kmike: docstring is no longer correct, as it embeds only image

splash/tests/mockserver.py
```diff
@@ -901,6 +901,39 @@ def render_GET(self, request):
...
+        @use_chunked_encoding
+        def render_GET(self, request):
+            request.setHeader(b"Content-Type", b"image/gif")
+            request.setHeader(b"Cache-Control", b"public, max-age=999999, s-maxage=999999")
+            return base64.decodebytes(b'R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=')
+
+    @use_chunked_encoding
+    def render_GET(self, request):
+        return ("""<html><head>
+            </head>
```
kmike: I think you can remove `<head>`, to make an example a bit shorter.

kmike:
Thanks @wsdookadr! 🎉

Mideen:
Hi team,
 When can we expect the next stable version with these changes? 
@kmike

PR #823, opened 2018-10-09
Fix Lua attribute types in _ExposedRequest.

### LINKED ISSUE #814: Incorrect string comparison with response.request.method in splash:on_response callback
In an `on_response` callback, comparison of the `method` attribute of the request is incorrect.

The following script demonstrates that the comparison `response.request.method == "GET"` is false, while it should be true. In the result, `num` should be the number of responses to GET requests, but it is zero. `method_attr_type` is "userdata".

    function main(splash, args)
        local num = 0
        local method_attr_type = nil
        splash:on_response(function(response)
            method_attr_type = type(response.request.method)
            if response.request.method == "GET" then
                num = num + 1
            end
        end)
        assert(splash:go("https://splash.readthedocs.io/"))
        assert(splash:wait(0.5))
        return {
            method_attr_type = method_attr_type,
            num = num
        }
    end

The following change fixes it while not breaking any existing tests, and with it `method_attr_type` in the above script is "string":

    diff --git a/splash/qtrender_lua.py b/splash/qtrender_lua.py
    index 1aea624..a542ad8 100644
    --- a/splash/qtrender_lua.py
    +++ b/splash/qtrender_lua.py
    @@ -2032,8 +2032,8 @@ class _ExposedRequest(BaseExposedObject):
     
         def __init__(self, lua, exceptions, url, method, headers, info):
             super(_ExposedRequest, self).__init__(lua, exceptions)
    -        self.url = url
    -        self.method = method
    +        self.url = self.lua.python2lua(url)
    +        self.method = self.lua.python2lua(method)
             # TODO: make info and headers attributes lazy
             self.headers = self.lua.python2lua(headers, encoding='latin1')
             self.info = self.lua.python2lua(info)

ArturGaspar:
Fixes #814.

### DIFF
--- a/splash/qtrender_lua.py
+++ b/splash/qtrender_lua.py
@@ -2032,8 +2032,8 @@ class _ExposedRequest(BaseExposedObject):
 
     def __init__(self, lua, exceptions, url, method, headers, info):
         super(_ExposedRequest, self).__init__(lua, exceptions)
-        self.url = url
-        self.method = method
+        self.url = self.lua.python2lua(url)
+        self.method = self.lua.python2lua(method)
         # TODO: make info and headers attributes lazy
         self.headers = self.lua.python2lua(headers, encoding='latin1')
         self.info = self.lua.python2lua(info)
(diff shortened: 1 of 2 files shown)


kmike:
Thanks @ArturGaspar!

PR #828, opened 2018-10-17
Added: URL to the code where the list of available functions in the sandbox is defined

starrify:


kmike:
Makes sense, thanks @starrify!

PR #867, opened 2019-02-02
[wip] initial QtWebEngine support

kmike:
This is very experimental, but I've finally managed to get QtWebEngine working in Splash Docker image - both engines are available for import now.

* [x] make QtWebEngine available in Docker
* [x] log Chromium version at startup
* [x] make render.html with default options work in Chromium
* [x] make basic render.png and render.jpeg work in Chromium
* [x] implement viewport support for Chromium (maybe with limitations)
* [x] refactor code, to assume Webkit less often
* [x] make it possible to disallow chromium, as a startup option
* [x] reduce Docker image size


lucywang000:
Great! Would this means splash could use the same rendering engine as headless chrome, on top of existing features like lua scripting support?

kmike:
@lucywang000 something like that, yes - though this PR only implements basics. 

Headless Chrome probably still is going to be more up to date than Qt's WebEngine - "We do update to the latest Chromium version in use before a Qt release", but Qt releases are not that frequent, and Splash is currently behind Qt (this is solvable) - Splash uses 5.9.2, while latest is 5.12.

pawelmhm:
what about /execute @kmike  , are there any chances for /execute to work in the long term? (you mention only render.jpg, render.png)

kmike:
@pawelmhm yes, I expect most of /execute work without changes to user scripts if we move forward with this.

splash/qtutils.py
```diff
@@ -8,12 +8,13 @@
...
 
 from PyQt5.QtCore import (QAbstractEventDispatcher, QDateTime, QObject,
-                          QUrl, QVariant, QEvent, Qt, QByteArray)
+                          QUrl, QVariant, QEvent, Qt, QByteArray, QSize)
 from PyQt5.QtGui import QKeyEvent
 from PyQt5.QtWidgets import QApplication
 from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkReply, QNetworkProxy
 from PyQt5.QtWebKit import QWebSettings
 from PyQt5.QtWebKitWidgets import QWebFrame
+from PyQt5.QtWebEngineWidgets import QWebEngineProfile
```
pawelmhm: one weird thing that happens to me is that splash crashes on this error ```python bin/splash 
  File "bin/splash", line 3, in <module>
    from splash.server import main
  File "/home/pawel/scrapinghub/splash/splash/server.py", line 11, in <module>
    from splash.qtutils import init_qt_app
  File "/home/pawel/scrapinghub/splash/splash/qtutils.py", line 17, in <module>
    from PyQt5.QtWebEngineWidgets import QWebEngineProfile
ImportError: /home/pawel/Qt5.9.1/5.9.1/gcc_64/lib/libQt5WebEngineCore.so.5: symbol _ZSt24__throw_out_of_range_fmtPKcz version Qt_5 not defined in file libQt5Qml.so.5 with link time reference
``` with above traceback, but when I move this import up to line 9, before import of QtCore it works fine. Not sure if this is something about my local dev environment (I use ubuntu 18.04 with all libraries installed manually), or something else, it works fine on master though

Comment (kmike) [reply splash/qtutils.py]: Does it happen in Docker, or are you installing everything without it?

Comment (kmike) [reply splash/qtutils.py]: That's weird. 

Installing everything was tricky, I got quite a lot of errors like that until making it work, but haven't seen this particular issue. PyQT bundles its own QT, it can cause some issues, I'm not sure.

Comment (pawelmhm) [reply splash/qtutils.py]: ah sorry my fault @kmike  I see now even on master it crashes later with: "Cannot mix incompatible Qt library (version 0x50905) with this library (version 0x50901)"

so I probably installed something wrong, need to fix that

File: splash/qtrender_image.py [hunk 2]
```diff
@@ -68,20 +77,20 @@ def is_jpeg(self):
     def is_png(self):
         return self.image_format == 'PNG'
 
-    def qimage_to_pil_image(self, qimage):
+    def _qimage_to_pil_image(self, qimage: QImage) -> Image:
```
lucywang000: The travis tests are failed with
```
E   AssertionError: 400 != 200 : (400, 
'{"info": {"message": "Lua error: AttributeError(\\"\'QtImageRenderer\' object has no attribute 
\'qimage_to_pil_image\'\\",)", "type": "LUA_ERROR"}, "description": "Error happened while executing Lua script", 
"error": 400, "type": "ScriptError"}',
 'http://localhost:44451/execute?url=http%3A%2F%2Flocalhost%3A49991%2Fvery-long-green-page&viewport=50x1024&render_all=1&engine=webkit&lua_source=main+%3D+require%28%22emulation%22%29.render_png&wait=0.01')
```

https://travis-ci.org/scrapinghub/splash/builds/496767326

splash/render_options.py
```diff
@@ -353,6 +353,12 @@ def get_forbidden_content_types(self):
     def get_html5_media(self):
         return self._get_bool("html5_media", defaults.HTML5_MEDIA_ENABLED)
 
+    def get_engine(self):
+        engine = self.get("engine", default="webkit", type=str)
+        if engine not in {"webkit", "chromium"}:
```
lucywang000: I think it's helpful to add "chrome" as an alias to "chromium"

splash/qtrender_image.py
lucywang000: see #889

splash/render_options.py
kmike: hey @lucywang000! I'd prefer not to add "chrome", to avoid confusion; also, technically, engine is not chrome.

lucywang000:
Glad to see this merged!  @kmike Is it possible to add developer tool support when using splash's chromium engine? https://doc.qt.io/qt-5/qtwebengine-debugging.html

background: i have spend quite long time this week (8h+) to debug a weird behavior of using splash to render a website (with custom lua scripts) and it's really painful experience - it's like debugging a python program with only "print" statement ... It would be much easier if we can live debug it.

kmike:
@lucywang000 it should be possible to have a live window with a debugger for webkit (see https://splash.readthedocs.io/en/stable/kernel.html#live-webkit-window). 

I guess to make it work for chromium we need to allow passing a startup option somehow? Pull requests are welcome, though chromium currently doesn't work very well with splash at the moment - a lot of features are missing, screenshots are not in a good state.

lucywang000:
> it should be possible to have a live window with a debugger for webkit (see https://splash.readthedocs.io/en/stable/kernel.html#live-webkit-window).

Oh that's a great to know that we can live debug the webkit window! I'll try it out.

> I guess to make it work for chromium we need to allow passing a startup option somehow? Pull requests are welcome, though chromium currently doesn't work very well with splash at the moment - a lot of features are missing, screenshots are not in a good state.

Thanks for the information, I'll see what I can do.

PR #877, opened 2019-02-19
Update faq.rst for FAQ section about rendering (add examples and reformat)

PyExplorer:


docs/faq.rst
```diff
@@ -258,8 +258,10 @@ https://github.com/nabilm/ansible-splash.
 
 Website is not rendered correctly
 ---------------------------------
+In this case we have two types of issues: wrong rendering and Splash crashing.
```
kmike: What do you think about splitting this in two sections - "Website is not rendered correctly" and "Splash is crashing"? currently both crashing and wrong rendering is under "wrong rendering"

docs/faq.rst
```diff
@@ -300,6 +292,43 @@ Common reasons:
   A quick (though not precise) way to check it is to try opening a page
   in Safari.
 
+**Splash crashes.**
+
+Common reasons:
+
+* Qt or WebKit bugs which cause Splash to hang or crash. Often the whole
+  website works, but some specific .js (or other) file causes problems.
+  In this case you can try this steps:
```
kmike: this -> these

docs/faq.rst
```diff
@@ -300,6 +292,43 @@ Common reasons:
...
+  * If Splash instance failed and stopped - go to the next steps. For example the last request was to the url like ``https://.../static/tagtag.min.js`` with JS,  and we can suspect that probably it caused the issue.
+  * Try to filter out this script with using :ref:`splash-on-request` with this template (or use use :ref:`request filters` to filter it out) :
+  
+    .. code-block:: lua
+
+     function main(splash, args)
+     splash:on_request(function(request)
+       if request.url:find('tagtag') ~= nil then
+             request:abort()
+         end
```
kmike: it seems indentation is a bit off

kmike: it'd be also good to use consistent indentation - here 4 spaces are used, while in return statement 2 spaces are used

kmike:
Thanks @PyExplorer! It looks great overall, much more readable than before.

PyExplorer:
Thanks for comments! Will fix them.

docs/faq.rst
```diff
@@ -300,6 +292,45 @@ Common reasons:
...
+                request:abort()
+            end
+         end)
+         assert(splash:go(args.url))
+         assert(splash:wait(0.5))
+         return {
+            html = splash:html(),
+            png = splash:png(),
+            har = splash:har(),
+         }
```
kmike: Could you please check indentation again? There are places in this function where 3 spaces are used, and places where 4 spaces are used. I think in this example it should be 4 spaces everywhere.

kmike:
Thanks @PyExplorer!

Issue #58, opened 2013-11-12
Pages that take a lot of CPU to process don't respect timeout


kmike:
Big XML sitemap files can take a lot of CPU time to render; timeout doesn't work for them. It actually happened that such 4MB sitemap (https://gust.com/sitemap.xml)  took 100% CPU for 30 minutes, and splash became unresponsive. We must fix this somehow.

PR #902, opened 2019-05-24
[WIP] Implement a maximum response size option

Gallaecio:
Based on #58.


splash/network_manager.py
```diff
@@ -26,10 +26,77 @@
...
+                value = value.decode('latin1')
+                value = int(value)
+            except (UnicodeDecodeError, ValueError):
+                raise _InvalidContentLength(value)
+            if value < 0:
+                raise _InvalidContentLength(value)
+            return value
+
+
+def _size_warrants_abort(sizes_and_sources, render_options, log, reply):
```
lucywang000: this function looks over-complicated, can we improve it somehow?

splash/defaults.py
```diff
@@ -6,6 +6,9 @@
 
 MAX_TIMEOUT = 90.0
 
+RESPONSE_SIZE_LIMIT = None
+MAX_RESPONSE_SIZE_LIMIT = None
```
lucywang000: This two looks very alike so a bit confusing, please add a comment about what each they stands for? 

BTW why do we need two variables for this limit?

splash/network_manager.py
```diff
@@ -413,6 +502,16 @@ def _on_reply_headers(self):
         self.log("Headers received for {url}", reply, min_level=3)
 
     def _on_reply_download_progress(self, received, total):
+        reply = self.sender()
+        request = reply.request()
```
lucywang000: `request` is not used

splash/network_manager.py
```diff
@@ -86,6 +153,7 @@ class ProxiedQNetworkAccessManager(QNetworkAccessManager):
     * Tracks information about requests/responses and stores it in HAR format,
       including request and response content.
     * Allows to set per-request timeouts.
+    * Handles per-request response size limits.
```
lucywang000: I don't find how to set the "per-request response size limit" in this PR.

splash/defaults.py
Gallaecio: If I recall right:

- `RESPONSE_SIZE_LIMIT` is the limit to use as default if the Splash request does not specify one.

- `MAX_RESPONSE_SIZE_LIMIT` is the maximum value that a Splash request may specify. If a higher value is specified by a Splash request, `MAX_RESPONSE_SIZE_LIMIT` is used as limit instead.

It is indeed confusing, I should indeed add a comment.

splash/network_manager.py
Gallaecio: Maybe I can extract some of the inner logic into separate functions, such as the parsing of `max_size`.

splash/network_manager.py
Gallaecio: By adding `max_size` to the render options of a request.

If you mean that the documentation should explain this, indeed.

Issue #511, opened 2016-10-10
scrapy jupyter webkit not showing


Dauth:
i am having the same issue as mentioned previously by @genius001  #361 and #376 whereby webkit is not showing up (installed using the manual way). Is there anway i can do to make this work?

 I have also tried the docker way but an error message displayed was shown below

No protocol specified
QXcbConnection: Could not connect to display unix:0

acowpy:
I solved this by running

`xhost +local:docker`

to give docker the relevant permissions

mcmino:
@acowpy this solution worked for me too! Thanks :D

Gallaecio:
It may be worth covering in the documentation.

Gallaecio:
Fixed by #903

PR #903, opened 2019-05-24
Add xhost configuration

mcmino:
Allow docker to connect to the X server, as described here https://github.com/scrapinghub/splash/issues/511#issuecomment-379152062


kmike:
Awesome, thanks @mcmino for the fix, and @acowpy for the example!

PR #921, opened 2019-07-19
Ubuntu 18.04

kmike:


lucywang000:
Why 3.6 instead of use the latest 3.7?

kmike:
@lucywang000 no specific reason, I just wanted to do it in steps, and Python 3.6 is default in Ubuntu 18.04.

kmike:
I've checked that it uses a similar amount of memory, and has a similar Docker image size, and that Flash still works, but haven't checked speed changes. Anyways, merging it.

PR #928, opened 2019-08-01
upgrade qtwebkit to QtWebkit-5.212.3-alpha3

### LINKED ISSUE #470: Splash looses fragment part of url (# hash sign part) after redirect
Today I stumbled on one bug that results from somewhat unusual behavior when redirecting to urls containing "#" hash sign. 

I have an url http://groceries.asda.com/asda-webstore/landing/home.shtml#search/ibuprofen/1/relevance_desc , when you request this url without any cookies site responds with redirect to: https://groceries.asda.com/asda-webstore/landing/home.shtml (same url but over https, note that you need to view traffic with mitm proxy as dev tools seem to hide redirects to first url). After redirect my desktop browser (Chrome/51.0.2704.84) keeps hash part of url. Splash seems to discard hash part after redirect. This means that site will not be rendered properly because hash part is missing.

I added tests on branch, see here: https://github.com/scrapinghub/splash/commit/cdae7c490210b83a41031decae6d8312566bee57 to reproduce you need to simply:

```
# start mockserver
> python3 splash/tests/mockserver.py
```

create following lua

``` lua
function main(splash)
    assert(splash:go(splash.args.url))
    return splash:url()
end
```

visit browser url: http://localhost:8998/redirect-hash#something you will see that hash is in url in input box.This confirms that test server is running and that your browser keeps url with hash even after redirect. 

Now check splash:

http://localhost:8050/execute?lua_source=function+main%28splash%29%0A++++assert%28splash%3Ago%28splash.args.url%29%29%0A++++return+splash%3Aurl%28%29%0Aend&url=http%3A%2F%2Flocalhost%3A8998%2Fredirect-hash%23something-bad

will return http://localhost:8998/redirect-hash⏎ instaed of http://localhost:8998/redirect-hash#something-bad.

I found some workaround for this (luckily I can avoid redirects by using https url and sending some extra cookie) but I'm opening issue for future reference in case someone else stumbles on this. I see in  QWebPage we have extension method and there is extensive comment there about redirects [here](https://github.com/scrapinghub/splash/blob/master/splash/qwebpage.py#L102), I'm not sure if it matters or how it matters here, but maybe that's some path for investigation.

kmike:
Currently this builds qtwebkit as a part of Docker build. It may be too heavy for CI, let's see.

Fixes #470.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,8 @@ RUN /tmp/download-qt-installer.sh /tmp/qt-installer.run
 # =====================
 
 FROM byrnedo/alpine-curl as qtwebkit-downloader
-COPY dockerfiles/splash/download-qtwebkit.sh /tmp/download-qtwebkit.sh
-RUN /tmp/download-qtwebkit.sh /tmp/qtwebkit.tar.gz
+COPY dockerfiles/splash/download-qtwebkit-source.sh /tmp/download-qtwebkit-source.sh
+RUN /tmp/download-qtwebkit-source.sh /tmp/qtwebkit.tar.xz
 
 # =====================
 
@@ -42,7 +42,6 @@ RUN /tmp/install-qtwebkit-deps.sh
 FROM qtbase as qtbuilder
 ENV DEBIAN_FRONTEND noninteractive
 
-COPY --from=qtwebkit-downloader /tmp/qtwebkit.tar.gz /tmp/
 COPY --from=qt-downloader /tmp/qt-installer.run /tmp/
 
 COPY dockerfiles/splash/qt-installer-noninteractive.qs /tmp/script.qs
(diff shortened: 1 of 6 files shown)


lucywang000:
I tried to build it (the webkit inspector in splash-jupyter image doesn't work properly so I think it may worth trying this new version of webkit), and it seems to take forever to build the webkit from source, with my laptop's cpu load rising out of charts :P

kmike:
@lucywang000 I recall inspector working properly in past, so I wonder if that's Splash problem, not webkit's. Have you got live window working? 

Building takes forever, right (a hour or so on a dual xeon, 2x8 cores workstation); we need to either upgrade to Qt 5.13 (to be able to use binaries), or build this image separately, once, and then just pull from it. 

Upgrading to Qt 5.13 is the way to go (we need to do it anyways), but having a separate Docker image which we copy artifacts from can work in a short term, and has its own benefits.

kmike:
(after talking with @imduffy15: probably better not to create docker images for webkit building - just build binaries locally and upload somewhere instead)

lucywang000:
@kmike Yeah I can see the live window, but none of elements / network / console tabs work. +1 to upload prebuilt binaries  somewhere and use it when build the image!

PR #929, opened 2019-08-01
Increase viewport defaults

ivanprado:


splash/defaults.py
```diff
@@ -13,9 +13,9 @@
 VIEWPORT_SIZE = '1024x768'
 
 # Window size limitations.
-VIEWPORT_MAX_WIDTH = 20000
-VIEWPORT_MAX_HEIGTH = 20000
-VIEWPORT_MAX_AREA = 4000*4000
+VIEWPORT_MAX_WIDTH = 40000
+VIEWPORT_MAX_HEIGTH = 40000
+VIEWPORT_MAX_AREA = 7200*7200  # ~52 Millions of pixels
```
kmike: It could be more clear if this is written as 1280 * 40000 - "support max height @ 1280 width"

ivanprado: Much better! I didn't notice where you had you gotten the magic number of 52M :-)

splash/defaults.py
```diff
@@ -13,9 +13,9 @@
 VIEWPORT_SIZE = '1024x768'
 
 # Window size limitations.
-VIEWPORT_MAX_WIDTH = 20000
-VIEWPORT_MAX_HEIGTH = 20000
-VIEWPORT_MAX_AREA = 4000*4000
+VIEWPORT_MAX_WIDTH = 40000
+VIEWPORT_MAX_HEIGTH = 40000
+VIEWPORT_MAX_AREA = 1280*VIEWPORT_MAX_HEIGTH  # support max height @ 1280 width
```
kmike: a nitpick: could you please add spaces around `*`, as per pep8?

ivanprado: Sure! I didn't put them because I thought it was the standard for the project.

ivanprado: Done

kmike:
Thanks @ivanprado!

PR #936, opened 2019-08-14
Fix splash-jupyter

kmike:
* [x] building of Docker image
* [x] permission errors
* [x] outdated dependencies
* [x] splash-jupyter starts
* [ ] splash-jupyter works


kmike:
It still doesn't work reliably, but it is better than before => merging.

PR #944, opened 2019-09-05
Bump version to 3.4

lucywang000:
The motivation is I'm doing some comparison to locate a splash bug so I need to switch back and forth between 3.3.1 and 3.4b1, but for 3.4b1 the web ui still shows 3.3.1 which is confusing :)


CHANGES.rst
```diff
@@ -1,6 +1,20 @@
 Changes
 =======
 
+3.4b1 (2019-08-17)
```
kmike:
```suggestion
3.4 (TBA)
```
I'd avoid release notes for 3.4b1, and go straight into 3.4; 3.4b1 is not supposed to be a public release anyways.

CHANGES.rst
```diff
@@ -1,6 +1,20 @@
...
+------------------
+
+* Suport using :ref:`chromium <arg-engine>` instead of webkit as the
+  rendering engine. It is in pre-alpha stage so not suggested for
+  using in production projects.
+* Qt is updaeted to 5.13, and QtWebkit is updated to
+  QtWebkit-5.212.3-alpha3.
+* Twisted is updated to 19.7.0.
+* Use ubuntu:18.04 as the base docker image.
+* HTTP2 can be disabled with
```
kmike: the important thing is that Splash got http2 support, which it haven't had before

kmike:
Hey @lucywang000! Sorry, I was unclear in a comment. 3.4 release is not ready, I'd like a couple more fixes before release:

1) either wait for alpha4 webkit release (https://github.com/qtwebkit/qtwebkit/issues/889#issuecomment-521962909), or disable http2 by default. Currently http2 is enabled by default, but it doesn't work properly on many websites.
2) Less critical: it'd be great to have a way to disable logging of Lua scripts, and probably other arguments, because log size becomes too large. This can be done later though.

What I meant is that it is better to have complete 3.4 changelog in 3.4 section, without a separate section for 3.4b1, which is not an official release.

kmike:
Thanks @lucywang000! I've made 3.4 release, waiting for docker images to build.

PR #961, opened 2019-10-27
Update Windows multiline command using caret

jmartens:
Windows uses caret to split command over multiple lines


kmike:
Thanks @jmartens!

PR #970, opened 2019-11-29
Make HTTP2 opt-in in the default engine

Gallaecio:


splash/engines/chromium/webpage.py
```diff
@@ -4,6 +4,8 @@
 
 
 class ChromiumWebPage(QWebEnginePage):
+    http2_enabled = True
```
Gallaecio: I’m not sure if this is needed. The Webkit counterpart has a few default values specified like this.

splash/defaults.py
```diff
@@ -104,4 +104,5 @@
 MEDIA_SOURCE_ENABLED = True
 
 # HTTP2
-HTTP2_ENABLED = True
\ No newline at end of file
+HTTP2_ENABLED = False
```
kmike: What do you think about renaming it to WEBKIT_HTTP2_ENABLED?

splash/tests/test_render.py
```diff
@@ -368,15 +368,17 @@ def assertResponse200Get(self, code):
...
+        })
         self.assertStatusCode(r, 200)
         self.assertIn("http2", r.text)
 
     @https_only
-    def test_disable_http2_support(self):
+    def test_http2_disabled_by_default(self):
         r = self.request({
             'url': self.ts.mockserver.https_url("http-version"),
-            'http2': '0',
```
kmike: I haven't checked it, but why isn't explicit disabling working?

Gallaecio: It does work, but since we had no test explicitly enabling, I did not though of having one explicitly disabling. I’ll add a test relying on the default but keep a test explicitly disabling.

splash/engines/chromium/webpage.py
kmike: I think it is not needed here. In webkit counterpart it is used in these places:

* https://github.com/scrapinghub/splash/blob/86b0f934f17d4118ba4fd7c359e6d322dfaf7f74/splash/engines/webkit/browser_tab.py#L216
* https://github.com/scrapinghub/splash/blob/86b0f934f17d4118ba4fd7c359e6d322dfaf7f74/splash/network_manager.py#L328

kmike:
Thanks @Gallaecio!

PR #976, opened 2019-12-09
Configure bumpversion

Gallaecio:
Running `bumpversion --dry-run --verbose patch` with `bumpversion` installed after these changes seems to describe the right list of changes to apply.


kmike:
Thanks for a careful PR!

PR #977, opened 2019-12-09
Configure Travis CI to deploy tagged versions to PyPI

Gallaecio:
@kmike You should be able to edit these changes and include the deployment username and token before merging.


Gallaecio:
Adding valid credentials is not as trivial as I hoped due to https://travis-ci.community/t/travis-encrypt-data-too-large-for-pypi-tokens-with-older-repos/5792

PR #997, opened 2020-02-13
Update utils.py

ater49:
Make HAR of Splash compatible with HAR1.2 specification. https://github.com/ahmadnassri/har-spec/blob/master/versions/1.2.md


ater49:
Not relevant

PR #998, opened 2020-02-18
Better viewport limits error message

ivanprado:


lopuhin:
@kmike could you please review this as well, as I'm not sure about the process for merging splash PRs - e.g. I don't see the tests running for this PR.

kmike:
The change looks good, thanks @ivanprado! A good catch about the tests; it seems the build started, but was killed due to lack of activity on one of the steps. Restarted it now. I'm not sure why it is not showing here. Build: https://travis-ci.org/scrapinghub/splash/builds/651993499

kmike:
Actually, the build for other PR was also stalled: https://travis-ci.org/scrapinghub/splash/builds/649925646. Maybe CI is broken (which is a problem, because it means there won't be a new Docker image).

kmike:
Thanks @ivanprado!

kmike:
Hey @ivanprado! Good news: bulid is fixed, thanks to @imduffy15, @lucywang000 and @Gallaecio. Bad news: tests are failing after this changes, because they were checking error message (see https://travis-ci.org/github/scrapinghub/splash/builds/675004312). Do you think you can take a look? At seems at least one of the failures is related to the fix here.

ivanprado:
Great! @kmike sure, I'm going to investigate.

ivanprado:
Working for a fix in https://github.com/scrapinghub/splash/pull/1021

PR #999, opened 2020-02-18
Update CHANGES.rst

hrbrmstr:
3.4.1's year was 2019 and I changed it to 2020


kmike:
Thanks @hrbrmstr! This is already fixed in master, but makes sense to fix it in a branch as well.

PR #1008, opened 2020-02-27
show logging in jupyter

lucywang000:
Seeing the errors and other logs is helpful when debugging lua scripts.


kmike:
I haven't checked it - where are these logs printed - is it console log, or right in the notebook? I wonder why it was disabled in a first place.

lucywang000:
@kmike it is printed in the command line console.

kmike:
Thanks @lucywang000 and @Gallaecio !

PR #1010, opened 2020-03-03
Workaround for #1003

### LINKED ISSUE #1003: Fix CI
At the moment, Travis CI seems to get stuck during Qt installation: https://travis-ci.org/scrapinghub/splash/builds/651993499

It may have something to do with https://www.qt.io/blog/qt-offering-changes-2020 (they now require a Qt account to download Qt)

imduffy15:
By disabling the QT installers internet access login details are no longer required.

To achieve this I have had to split the build process up slightly as shown by the changes to the travis file.

Fixes #1003

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,8 @@ ENV DEBIAN_FRONTEND noninteractive
 
 COPY --from=qt-downloader /tmp/qt-installer.run /tmp/
 
+ARG DISPLAY
+
 COPY dockerfiles/splash/qt-installer-noninteractive.qs /tmp/script.qs
 COPY dockerfiles/splash/run-qt-installer.sh /tmp/run-qt-installer.sh
 RUN /tmp/run-qt-installer.sh /tmp/qt-installer.run /tmp/script.qs
--- a/dockerfiles/splash/run-qt-installer.sh
+++ b/dockerfiles/splash/run-qt-installer.sh
@@ -2,8 +2,14 @@
 # XXX: if qt version is changed, Dockerfile should be updated,
 # as well as qt-installer-noninteractive.qs script.
 
+if [ -z ${DISPLAY+x} ]; then
+	command="xvfb-run $1"
+else
+	command="$1"
+fi
+
 chmod +x "$1" && \
-http_proxy="http://localhost:8080" https_proxy="http://localhost:8080" xvfb-run "$1" --script "$2" \
+http_proxy="http://localhost:8080" https_proxy="http://localhost:8080" $command --script "$2" \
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -9,7 +9,10 @@ branches:
         - /^\d\.\d+$/
 
 before_install:
-    - docker build -t splash .
+    - docker build -t deps -f DepsDockerfile .
+    - docker build -t base -f SplashDockerfile .
```
Gallaecio: Why is this image build twice?

imduffy15: Mistake, will resolve. Thank you.

kmike:
From a technical side, I think this is going to break https://hub.docker.com/ builds (well, at least not fix them).

Gallaecio:
I wonder if we can disable network from within the Dockerfile, just for the Qt installer, some other way (e.g. https://stackoverflow.com/q/50907073/939364)

Gallaecio:
An option would be to manually build an image up to the Qt-being-installed point, create a new Docker repository for that image, and push tags of that image manually from local builds, and have the automated Splash build based on tags of that image.

Although it seems rather complicated, decreasing the level of automation we have for images.

imduffy15:
> I wonder if we can disable network from within the Dockerfile, just for the Qt installer, some other way (e.g. https://stackoverflow.com/q/50907073/939364)

Leave this idea with me, I might come up with something ;)

lucywang000:
A possible way to disable network access when installing qt, is to launch the installer through [proxychains](https://github.com/haad/proxychains) with a non-exist proxy. In the slackbot project's [integration tests](https://github.com/lins05/slackbot/blob/master/scripts/slackbot-test-ctl), it need  to verify the bot could successfully reconnect to the slack server after a network disconnect, which is achieved by using this trick.

imduffy15:
Interesting idea @lucywang000 will try that locally and change this PR if it works.

imduffy15:
@lucywang000 fantastic suggestion and I'm happy to report that locally it looks to have worked nicely. Thankfully, just setting the environment variables to specify a non-existent proxy was enough.

To test this, I ran X11 on my mac with it configured to accept network connections. I set environment variable DISPLAY within the container to `host.docker.internal:0` and dropped `xvfb-run`. The GUI opened up on my mac during docker build and went on unattended.

On removing all this, and just leaving the http_proxy and https_proxy `docker build .` ran as expected.

@kmike @Gallaecio I hope this is a resolution that works for you both? :)

kmike:
Thanks @imduffy15, @Gallaecio and @lucywang000!

PR #1017, opened 2020-04-03
Fix a typo in faq

ahivert:
timeous -> timeout

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -196,7 +196,7 @@ If you want to get started quickly take a look at Aquarium_
 (which is a Splash setup without many of the pitfalls) or use
 a hosted solution like `ScrapingHub's <http://scrapinghub.com/splash/>`__.
 
-Don't forget to use resource timeout in your client code (see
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -196,7 +196,7 @@ If you want to get started quickly take a look at Aquarium_
 (which is a Splash setup without many of the pitfalls) or use
 a hosted solution like `ScrapingHub's <http://scrapinghub.com/splash/>`__.
 
-Don't forget to use resource timeous in your client code (see
+Don't forget to use resource timeout in your client code (see
```
Gallaecio, line 199: Given the referenced section and the distance between T and S in a QWERTY keyboard, I think the right fix may be:

```suggestion
Don't forget to use resource timeouts in your client code (see
```

:slightly_smiling_face:

ahivert: :heavy_check_mark:

kmike:
Thanks @ahivert!

PR #1021, opened 2020-04-15
Fix test_viewport_size_validation test

ivanprado:


ivanprado:
This seems to fix the bug :clap: https://travis-ci.org/github/scrapinghub/splash/builds/675223461

kmike:
Awesome, thanks @ivanprado!

PR #1037, opened 2020-05-28
Container for building

whalebot-helmsman:
In this PR:
- ability to set Qt version using docker build arguments
- special container is added to build https://github.com/qtwebkit/qtwebkit from source and create a compatible packages. E.g.
```
docker build --target qtwebkitbuilder-base . -t qtwebkit-builder
git clone git@github.com:qtwebkit/qtwebkit.git ../qtwebkit
docker run --rm -it -v `pwd`/../qtwebkit:/qtwebkit qtwebkit-builder
 # inside container
> cd /qtwebkit
> mkdir build 
> cd build
> cmake -G Ninja -DPORT=Qt -DCMAKE_BUILD_TYPE=Release .. 
> ninja -j 8
> ninja install
> /tmp/create-package.sh install_manifest.txt '' 7z
# /qtwebkit/buid/build.7z is package compatible with qtwebkit/qtwebkit's releases
```
- docker build argument to set amount of build workers in every build script is added
- docker build argument to set URL of webkit binary build is added
- qt version is updated to 5.14.1 
- qtwebkit version is updated to 5.212.0-alpha-4


dockerfiles/splash/install-python-splash-deps.sh
```diff
@@ -5,7 +5,7 @@ install_python_deps () {
     # Install python-level dependencies.
     ${_PYTHON} -m pip install -U pip setuptools six && \
     ${_PYTHON} -m pip install \
-        qt5reactor==0.5 \
+        qt${QT_MAJOR_VERSION}reactor==0.5 \
```
kmike, line 8: this is a specific pypi package, I'm not sure they're going to change the package name

dockerfiles/splash/download-pyqt5.sh
```diff
@@ -2,10 +2,9 @@
 # XXX: these URLs needs to be replaced with sourceforge in future,
 # because riverbank tend to remove old releases.
 SIP="https://www.riverbankcomputing.com/static/Downloads/sip/4.19.19/sip-4.19.19.tar.gz"
-PYQT5="https://www.riverbankcomputing.com/static/Downloads/PyQt5/5.13.1/PyQt5_gpl-5.13.1.tar.gz"
-WEBENGINE="https://www.riverbankcomputing.com/static/Downloads/PyQtWebEngine/5.13.1/PyQtWebEngine_gpl-5.13.1.tar.gz"
+PYQT="https://www.riverbankcomputing.com/static/Downloads/PyQt${QT_MAJOR_VERSION}/${QT_FULL_VERSION}/PyQt${QT_MAJOR_VERSION}_gpl-${QT_FULL_VERSION}.tar.gz"
+WEBENGINE="https://www.riverbankcomputing.com/static/Downloads/PyQtWebEngine/${QT_FULL_VERSION}/PyQtWebEngine_gpl-${QT_FULL_VERSION}.tar.gz"
```
kmike, line 6: I'm not sure QT and PyQT versions match; it seems PyQt 5.13.x supports Qt 5.13.x, but .x is updated separately. Also, https://github.com/scrapinghub/splash/pull/1038 shows that you can use a non-matching PyQT version.

The problem here is that it is not possible to update to PyQT 5.14 just by changing version number - package may be not available on the same URL, and they also changed the installation method (current scripts don't work).

dockerfiles/splash/qt-installer-noninteractive.qs
```diff
@@ -24,11 +24,19 @@ Controller.prototype.IntroductionPageCallback = function() {
 
 Controller.prototype.TargetDirectoryPageCallback = function()
 {
-    gui.currentPageWidget().TargetDirectoryLineEdit.setText("/opt/qt-5.13");
+    var short_version = installer.environmentVariable("QT_SHORT_VERSION");
```
kmike, line 27: nice!

dockerfiles/splash/download-pyqt5.sh
sibiryakov: what if we will be building it ourselves?

kmike: We are building PyQT and PyQtWebkit ourselves, using source packages; this PR adds an option to build qtwebkit ourselves.

whalebot-helmsman: Riverbank is providing source tars in pypi now https://files.pythonhosted.org/packages/47/9f/60e630711fd1dd14ef3bd95c86c733c86b8c0853749c7a03691f681f13fd/PyQtWebEngine-5.14.0.tar.gz, https://files.pythonhosted.org/packages/4d/81/b9a66a28fb9a7bbeb60e266f06ebc4703e7e42b99e3609bf1b58ddd232b9/PyQt5-5.14.2.tar.gz and https://files.pythonhosted.org/packages/bc/dd/aa60ae73bc213466fd158e3d8b91d55ac00a248b14a98919bfe0e12c5d74/sip-5.2.0.tar.gz

README.rst
```diff
@@ -41,3 +41,32 @@ Commercial support is also available by `Scrapinghub`_.
 .. _create a question issue: https://github.com/scrapinghub/splash/issues/new?labels=question
 .. _Scrapinghub: https://scrapinghub.com
 .. _scrapy-splash library: https://github.com/scrapy-plugins/scrapy-splash
+
+Building qtwebkit from sources
```
kmike, line 45: Could you please move it either to https://github.com/scrapinghub/splash/blob/master/docs/install.rst, or to a separate README file inside dockerfiles/splash?

Dockerfile
```diff
@@ -1,4 +1,22 @@
+ARG QT_MAJOR_VERSION=5
+ARG QT_MINOR_VERSION=14
+ARG QT_PATCH_VERSION=1
+# docker build has arguments but didn't has variables
```
kmike, line 4:
```suggestion
# docker build has arguments but doesn't have variables
```

dockerfiles/splash/install-pyqt5.sh
```diff
@@ -1,15 +1,14 @@
...
-SPLASH_BUILD_PARALLEL_JOBS=8
 
 mkdir -p /tmp/builds/sip && \
 mkdir -p /tmp/builds/pyqt5 && \
 pushd /tmp/builds && \
 # sip
 tar xzf "$1" --keep-newer-files -C sip --strip-components 1 && \
 pushd sip && \
-${_PYTHON} configure.py --sip-module PyQt5.sip && \
+${_PYTHON} configure.py --sip-module PyQt${QT_MAJOR_VERSION}.sip && \
```
kmike, line 11: As the file is called install-pyqt5, I'm not sure module name should be generated. We don't know how if PyQt5 is going to change its name or not, when Qt6 is landed.

```suggestion
${_PYTHON} configure.py --sip-module PyQt5.sip && \
```

dockerfiles/splash/install-pyqt5.sh
```diff
@@ -35,7 +34,7 @@ ${_PYTHON} configure.py -c -j ${SPLASH_BUILD_PARALLEL_JOBS} \
 make -j ${SPLASH_BUILD_PARALLEL_JOBS} && \
 make install && \
 popd  && \
-${_PYTHON} -c "import PyQt5.QtCore; print(PyQt5.QtCore.__file__)"
+${_PYTHON} -c "import PyQt${QT_MAJOR_VERSION}.QtCore; print(PyQt${QT_MAJOR_VERSION}.QtCore.__file__)"
```
kmike, line 37:
```suggestion
${_PYTHON} -c "import PyQt5.QtCore; print(PyQt5.QtCore.__file__)"
```

Dockerfile
```diff
@@ -82,16 +122,27 @@ FROM qtbase as splash-base
...
+#RUN ls -l /opt/qt-${QT_SHORT_VERSION}/${QT_FULL_VERSION}/gcc_64/lib
+ENV PATH="/opt/qt-${QT_SHORT_VERSION}/${QT_FULL_VERSION}/gcc_64/bin:${PATH}"
+ENV LD_LIBRARY_PATH="/opt/qt-${QT_SHORT_VERSION}/${QT_FULL_VERSION}/gcc_64/lib:$LD_LIBRARY_PATH"
 
 # =====================
 
 FROM splash-base as qt5-builder
 
+ARG QT_MAJOR_VERSION
+ENV QT_MAJOR_VERSION=${QT_MAJOR_VERSION}
```
kmike, line 141: I haven't checked it, but it seems these args are not needed if QT_MAJOR_VERSION is not used to generate PyQt5 package name

kmike:
Thanks a ton @whalebot-helmsman and @imduffy15!

Gallaecio:
Working on the release notes, I’ve noticed a version mismatch between Qt (`5.14.1`) and PyQt (`5.13.2`), is that OK?

imduffy15:
@Gallaecio fantastic question, sadly no 5.14.x exists for PyQT, guessing a URL gives 404 https://www.riverbankcomputing.com/static/Downloads/PyQt5/5.14.1/PyQt5-5.14.1.tar.gz and the download page lists nothing https://riverbankcomputing.com/software/pyqt/download5 .

We are using this mix of versions internally for a high traffic system and haven’t seen any issues yet. 

@kmike may be able to detail potential issues?

imduffy15:
One looks to exist on pypi https://files.pythonhosted.org/packages/4d/81/b9a66a28fb9a7bbeb60e266f06ebc4703e7e42b99e3609bf1b58ddd232b9/PyQt5-5.14.2.tar.gz

kmike:
I tried using 5.14.2 some weeks ago; the problem is that it changed its installation method, so the current script is no longer working; I spent a couple hours, but was unable to make it work. We would need to figure out how to upgrade, but given that all tests pass with PyQt 5.13, I think there is no hard in releasing a version without fixing it.

PR #1041, opened 2020-06-08
Upgrade sip, pyqt5, pyqtwebengine

imduffy15:


dockerfiles/splash/download-pyqt5.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env sh
 # XXX: these URLs needs to be replaced with sourceforge in future,
```
kmike, line 2:
```suggestion
# XXX: riverbankcomputing URLs need to be replaced with files.pythonhosted.org in future,
```

imduffy15: Your attention to detail is amazing :)

kmike:
Awesome, thanks @imduffy15!

PR #1059, opened 2020-08-05
Allow overriding target test files in ‘docker run’ calls

Gallaecio:
After this change, when running the tests locally, you can for example append `tests/test_proxy.py` to the `docker run` command to run only the tests defined in that file.


kmike:
Thanks @Gallaecio!

PR #1067, opened 2020-09-04
Switch to a mirrored copy of SIP

imduffy15:
~The URL for SIP now 404s and is nowhere to be found. There is a new version of SIP which should resolve the issue.~

new version of SIP didn't resolve the issue, mirroring old sip and adjusting the URL.

### DIFF
--- a/dockerfiles/splash/download-pyqt5.sh
+++ b/dockerfiles/splash/download-pyqt5.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 # XXX: riverbankcomputing URLs need to be replaced with files.pythonhosted.org in future,
 # because riverbank tend to remove old releases.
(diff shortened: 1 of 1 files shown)


imduffy15:
@kmike Should we be mirroring these dependencies somewhere to avoid issues like this in the future?

kmike:
hey @imduffy15! Yeah, definitely having a mirror somewhere would be better, to prevent such issues. 

The build is failing for this PR, maybe because of the version upgrade - should we try to recover the old sip version instead of upgrading, or do you think a failure is caused by something else?

imduffy15:
Yeah failure is caused due to upgrade sadly. No idea where to start with fixing that. 

Will mirror the old version and update the URL.

kmike:
Thanks @imduffy15!

PR #1141, opened 2021-10-13
Add Github Actions

Laerte:
I pinned the `setuptools` version to `57.5.0`, because `58.0.0` have a [breaking change](https://github.com/pypa/setuptools/blob/main/CHANGES.rst#breaking-changes):
- #2086: Removed support for 2to3 during builds. Projects should port to a unified codebase or pin to an older version of Setuptools using PEP 518 build-requires.

That make `funcparserlib` dependency not work because it use `use_2to3`. 

Changed the `CIRCLE_CI` env to `CI` that is the variable that [Github Actions](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) use.

Used a [dirty hack](https://github.com/actions/runner/issues/241#issuecomment-924327172) in order to make TTY work. Explanation [here](https://explainshell.com/explain?cmd=script+--return+--quiet+--command+%22bash+%7B0%7D%22)

For some unknown reason i could not make it Codecov work in my tests.

I also have the action to publish image to docker repository but i not add to this PR (But i can if is needed).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,5 +23,6 @@ jobs:
           tags: splash-tests
       -
         name: Run tests
-        run: | 
+        shell: 'script --return --quiet --command "bash {0}"'
+        run: |
           docker run -it splash-tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,5 @@ jobs:
           tags: splash-tests
       -
         name: Run tests
-        shell: 'script --return --quiet --command "bash {0}"'
         run: | 
           docker run -it splash-tests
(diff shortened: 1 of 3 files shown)


Gallaecio:
I think the codecov issue is because [`.git` does not get included into the image](https://github.com/scrapinghub/splash/blob/master/.dockerignore#L20). Maybe it can be mounted as a Docker volume?

Laerte:
@kmike @pawelmhm Can we merge it?

Laerte:
@Gallaecio Any update on this?

wRAR (approved):

wRAR:
Thank you!

